Issue #101, opened 2018-05-24
Add a cascade generator 


dubzzz:
Provide an easier way to build an arbitrary whose parameters depends on another one. For instance we might use it to build a 2d array:

```
fc.cascade(
  fc.tuple(fc.nat(100),fc.nat(100)),
  ([x, y] ) = fc.array(fc.nat(),x*y,x*y).map(a => split2d(a , x, y) )
) ;
```

dubzzz:
Closed duplicate of #103

PR #103, opened 2018-05-28
chain()

hath995:
I started to try out fast-check on interview style algorithm problems but quickly hit a wall generating any sort of complicated situation where some values depended on randomly generated values. Adding flatMap can fix it though. Not sure if it is really correctly implemented in respect to shrinking but copied and tweaked the .map tests and it seems to work alright. 

Although the name flatMap is definitely correct, other libraries named it `then` which might be reasonable as well. Maybe add it as an alias?


dubzzz:
I opened a ticket for that `fc.cascade`, I had an implementation but the shrinking sucked. I will have a look to your PR asap. Thanks

hath995:
Despite what travis is saying the unit tests pass, however, typescript is upset because of the anonymous classes I think. Not sure how to resolve that.

dubzzz:
I have something to have a viable shrink, just trying to find the snippets I did.
Even if it is a `flatMap` I would maybe prefer `then`.  But I am opened to discussion for this choice

dubzzz:
Few days ago I created an issue regarding this need - https://github.com/dubzzz/fast-check/issues/101

As a good shrinker is difficult to have with a `then` or `flatMap` I was thinking of adding it as another arbitrary. Adding it as another arbitrary was a way to discourage usage of `flatMap` and favor approches with better shrinkers.

dubzzz:
Hi @hath995 ,

Thanks a lot for the Pull Request.
Concerning the reasons why it failed, I leave several comments on some of your commits.

**Concerning the feature itself:**
There is a real need for such feature. I started implemented one before the version 1.0.0 but got stuck on the shrinking part which is not trivial at all. I created a ticket #101 few days ago to add such feature in the framework. As the shrinking part was not trivial I though about adding it as another arbitrary but as time goes by I think adding it as a method of `Arbitrary` is a nice choice.

Nonetheless I would prefer something else than `flatMap`. Indeed the fact that shrinker is not as powerful as `filter` or `map` make it misleading, I prefer another name for that. What was the framework you talked about that had a `then` method for that? I think this might be a good naming.

**Your implementation iteself:**
Your implementation only shrink on the result of `fmapper`. Thus if you have something like:
`fc.tuple(fc.nat(100)).flatMap(num => fc.array(fc.nat(), num, num))`

You will not be able to shrink on the first level arbitrary and your shrunk array will have the size leading to the first failure and not a smaller one. But let's keep it that way for the moment.
It already a good implementation.

Once again, thanks a lot for your contribution,
Nicolas

---

My past implementation was the following:

```js
    then<U>(transformer: (t: T) => Arbitrary<U>): Arbitrary<U> {
        class RecursiveArbitrary<T, U> extends Arbitrary<U> {
            constructor(readonly rootArb: Arbitrary<T>, readonly transformer: (t: T) => Arbitrary<U>) {
                super();
            }
            generate(mrng: MutableRandomGenerator): Shrinkable<U> {
                const stepMrng = mrng.clone();
                const st = this.rootArb.generate(stepMrng);
                const su = this.transformer(st.value).generate(stepMrng);
                const shrinkAll: () => Stream<Shrinkable<U>> = () => {
                    return st.shrink().map((v: Shrinkable<T>) => {
                        return this.transformer(v.value)
                                .generate(mrng.clone());
                    });
                };
                return new Shrinkable(su.value, shrinkAll);
            }
        }
        return new RecursiveArbitrary(this, transformer);
    }
```

hath995:
The other library I used was test-check, https://github.com/leebyron/testcheck-js, http://leebyron.com/testcheck-js/api#valuegeneratorthen

dubzzz:
I am double checking the code to understand why travis failed.

Basically I run my test (of src) using fast-check itself (from lib - to have executor outside of the coverage). So there might be naming conflicts.

dubzzz:
Here is the diff you should apply to make your code run.

patch

Just a quick explanation concerning this import nightmare - I must clearly do something for that in order to help next contributors.

The tests in `test/unit` are composed of pure unit-tests and also property based tests. They check the code under `src`.

Those property based tests are run using fast-check itself (the one from `lib`). The choice to use the fast-check from `lib` as a property based testing framework and not the one from `src` was just to prevent counting it inside the coverage. Thus the coverage only count the code really tested.

*It can be challenged ;)*

dubzzz:
Your code runs well after the modifications above.

Here is the patch corresponding to the diff above:

[0001-fix.zip](https://github.com/dubzzz/fast-check/files/2046239/0001-fix.zip)

hath995:
Thanks for the help! That patch worked! That would have taken a while to figure out.

hath995:
I'll try to address the tests in next couple days and the documentation examples. How do you test shrinking in general? What sort of metric are you using to say something is a good or bad shrinker? Thanks

dubzzz:
Testing shrinking in general is quite difficult. The shrinker as to do a trade-offs between convergence speed and coverage. So might not be able to converge as well as you want depending on the shape of the counterexamples.

The way I test it today is the following:
- In what I called *unit* section: I check that my shrinkers build a finite tree of shrinks. For some shrinkers like integers, I check that the values decrease as we go in the shrinking tree. To cover all those cases I introduced an helper `genericHelper.isValidArbitrary`. It is used in most of the tests of arbitraries except the ones I forgot. https://github.com/dubzzz/fast-check/blob/master/test/unit/check/arbitrary/IntegerArbitrary.spec.ts#L151
- In the *e2e* section: I test for the convergence of the arbitraries. https://github.com/dubzzz/fast-check/blob/master/test/e2e/arbitraries/ArrayArbitrary.spec.ts

For the scope of this PR, I think you can just adapt as stated in https://github.com/dubzzz/fast-check/pull/103#discussion-diff-191501789R186

Then in another PR we can:
- add tests for convergence in `e2e` section
- check bias
- ...

*I can help you on all those points, as soon as it gets merge*

dubzzz:
If you want to understand better the problem solved by bias, you can have a look to the tests of ArrayArbitrary in e2e. There is twice the same test, one with bias, the other without.

The idea: higher priority for small entries in first runs while keeping large entries.

dubzzz:
For another PR - Ideas of end-to-end for withBias:

Examples:

Should bias on source arbitrary, 
It should be able to find a failure when using: `fc.nat().then(c => fc.constant(c))`
With predicate: `v => v > 10`

Basically it will pass most of the time if unbiased parameter is set to true.

Should bias on destination arbitrary, 
It should be able to find a failure when using: `fc.nat().then(c => fc.nat())`
With predicate: `v => v > 10`

Should bias on both levels together, 
It should be able to find a failure when using: `fc.nat().then(c => fc.tuple(fc.constant(c),fc.nat()))`
With predicate: `v => v[0] > 10 || v[1] > 10`

dubzzz:
In reality we could have 4 tests for withBias:
- bias nothing
- bias source only
- bias destination only
- bias both

All could be based on the arbitrary of the last example I provided. And we might consider something below 100 as biased and above as non biased

paldepind:
> Although the name flatMap is definitely correct, other libraries named it then which might be reasonable as well. Maybe add it as an alias?

The method name `then` is essentially hijacked by the [Promise/A+ specification](https://promisesaplus.com/). I'm not if it would be a good idea to add a `then` method to something that is not a promise. The JS [Fantasy Land](https://github.com/fantasyland/fantasy-land/) on the other hand suggests using the name `chain` for `flatMap`.

dubzzz:
@paldepind I have not thought of `chain` before and it is indeed a very good naming. We should definitely go for it. 

@hath995 I think the PR is OK. We can merge and rename to `chain` right after. I quickly checked the diffs yesterday and it seemed OK for me.

Are you OK to merge?

Before integrating it in the next minor, I will just rename it and add some more checks for biased and shrink.

It should come very soon :)

dubzzz:
Merged it ;)

@hath995 thanks a lot for the contribution

I just added a CONTRIBUTORS.md file, fill free to add your name in.
https://github.com/dubzzz/fast-check/blob/master/CONTRIBUTORS.md

dubzzz:
I should release a minor version of fast-check with your feature in it this week-end

dubzzz:
The feature has just been released with version 1.2.0 of fast-check

paldepind:
The docs says:

> kind of flatMap-operation called chain

While the text in this PR and the documentation claims that is it an actual `flatMap`. The type of `chain` also looks identical to `flatMap`. Isn't `chain` exactly like `flatMap`/`>>=`/`bind`?

dubzzz:
Indeed the original implementation was very close from a bind implementation.

It basically took the generated value, create a new arbitrary from it and forgot about the first arbitrary (the one of the value) when it was time to shrink.

I revamped a bit the implementation so that it does reuse the first arbitrary on shrink. The current chain operation is neither a bind (does a bit more) nor a flatMap (implementing map with the current flatMap would not be as powerful as the current map).

PR #133, opened 2018-06-28
Wrap assert and property into a single method

### LINKED ISSUE #109: Wrap assert and property into a single method
As suggested in #108, fast-check should come with a method doing `fc.assert(fc.property(..),..)` in one command. Same for asyncProperty.

Some frameworks already implemented such feature: jsverify for instance. 

assertProp and assertAsyncProp?

dubzzz:
Fixes #109


paldepind:
Why is the code generated from strings? Couldn't the same thing be accomplished with overloads?

dubzzz:
I agree with the fact that code generated by strings is far from good. Unfortunately I found no suitable alternative so far. I would be really interested if you have an hint for that.

Basically the first time I encountered this problem was with `property`. My aim: provide the user with a fully typed API.

```typescript
property(arb1: Arbitrary<T1>, predicate: ((v1: T1) => boolean));
property(arb1: Arbitrary<T1>, arb2: Arbitrary<T2>, predicate: ((v1: T1, v2: T2) => boolean));
// ... and more...
```

With this API, the compiler should warn the user if it comes with invalid types:
```typescript
fc.property(fc.string(), (v: number) => true); // does not work
```

I found no elegant way to do this in TypeScript. I wanted to use some variadic-types but they are not yet part of TypeScript.

dubzzz:
I should wait for TypeScript 3.0 official release in order to benefit from `Extracting and spreading parameter lists with tuples`. It will highly simplify the code but might require a major version upgrade of fast-check.

dubzzz:
Closing the PR, will do another one later

PR #140, opened 2018-07-01
Make seed and path copy pasteable

paldepind:
This PR changes the error message on a failed test slightly. It looks like this:

```
Property failed after 1 tests
{ seed: 1530466923032, path: "1:1:0:0:0:0:0:0:1:1:3:8:8:9:11:16:20:19:19:19:19:19:22:23" }
Counterexample: [[32,32738,34]]
Shrunk 0 time(s)
```

The major change is that the `seed` and the `path` is now printed so that they can be easily copy-pasted into one's code as an argument to `assert`. Additionally, the counterexample is prefixed with the text "Counterexample" to make it easier to understand and distinguish.

### DIFF
--- a/src/check/runner/utils/utils.ts
+++ b/src/check/runner/utils/utils.ts
@@ -31,9 +31,11 @@ function throwIfFailed<Ts>(out: RunDetails<Ts>) {
       );
     }
     throw new Error(
-      `Property failed after ${out.numRuns} tests (seed: ${out.seed}, path: ${out.counterexamplePath}): ${pretty(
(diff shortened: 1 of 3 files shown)


dubzzz:
Very good idea 💡 

Seems like prettier is complying: you should run `npm run format:fix`

paldepind:
Ah. I was only running `npm test`. Travis is happy now :smile:

dubzzz:
Merged, thanks for the contribution ;)
Please add yourself in the contributors file

PR #182, opened 2018-08-23
Move rollup on esm build

dubzzz:


dubzzz:
PR #181 will have to be reverted as soon as this test goes green

Issue #89, opened 2018-05-11
Introduce a placeholder for min/max parameters


dubzzz:
Building an array of length 3 or more is currently not ideal, solutions are the following:
- `fc.array(fc.nat()).filter(data => data.length >= 3)` - not optimized, will reject too many arrays
- `fc.array(fc.nat(), 3, 100)` - but why 100? Why do the user has to specify this value?
- `fc.tuple(fc.nat(), fc.nat(), fc.nat(), fc.array(fc.nat())).map(([a,b,c,data]) => [a,b,c,...data])` - really to complicated

The idea is to introduced a placeholder whenever possible to be able to do:
- `fc.array(fc.nat(), 3, fc._)` - min = 3, max = auto
- `fc.array(fc.nat(), fc._, 3)` - min = auto, max = 3
- `fc.array(fc.nat(), fc._, fc._)` - min = auto, max = auto

dubzzz:
The feature is more or less ready. But I prefer to delay it for the v1.2.0.

SupernaviX:
Would an options hash be an acceptable API? I think `fc.array(fc.nat(), { min: 3 })` is a pretty idiomatic way to solve this.

dubzzz:
@SupernaviX 
Indeed I think I will move to this kind of signature - with `min` and `max` as attributes of an object.
The placeholder way - `fc.array(fc.nat(), 3, fc._)` - seems less idiomatic and does not adapt to all situations. The only benefit I see is that it seems lighter for cases where you need to specify both min and max.

I think I will update the branch I started to implement this feature to the `{min,max}` way.

In any cases, I need some kind of: *I want a min not a max, or the contrary*. I prefer not forcing the user into specifying a value for min or max - the framework should do the magical part itself ;)

SupernaviX:
Definitely agreed. With this kind of library in particular, it's useful for consumers to only describe constraints they need, and fall back to sensible defaults everywhere else.

dubzzz:
@SupernaviX Even if the `{min, max}` notation is more idiomatic ES6+, I think I will stay to the placeholder approach for the moment (at least for 1.x versions).

Indeed it better fits with the current API. The `{min, max}` would have introduced another notation. 

Moreover, knowing that I will not remove any of the existing signatures for 1.x (want to respect semver as much as possible), I am not sure users wanting only a max will be using `fc.array(arb, {max: maxSize})` as they can simply call `fc.array(arb, maxSize)` instead.

[...]

PR #204, opened 2018-10-04
Add performance measurement example

dubzzz:


perf/index.js
```diff
@@ -0,0 +1,47 @@
+const Benchmark = require('benchmark');
+const fc = require('../lib/fast-check');
+
+const benchSettings = function(label, arb) {
```
trollepierre: what does `arb`mean? Maybe you could use the complete word to make code more simple to understand

perf/index.js
```diff
@@ -0,0 +1,47 @@
...
+        .run({ async: false });
+};
+
+benchSettings('boolean', fc.boolean());
+benchSettings('integer', fc.integer());
+benchSettings('integer (min, max)', fc.integer(Number.MIN_SAFE_INTEGER, Number.MAX_SAFE_INTEGER));
+benchSettings('float', fc.float());
+benchSettings('double', fc.double());
+benchSettings('array', fc.array(fc.integer()));
+benchSettings('array (min size)', fc.array(fc.integer(), 5, 10));
```
trollepierre: 5 and 10 look like magic number here

perf/index.js
```diff
@@ -0,0 +1,47 @@
...
+
+(new Benchmark.Suite())
+    .add('filter', () => fc.assert(fc.property(fc.nat(3).filter(i => i != 0), () => { return true; })))
+    .add('pre', () => fc.assert(fc.property(fc.nat(3), (i) => { fc.pre(i != 0); return true; })))
+    .on('cycle', (event) => console.log(String(event.target)))
+    .on('complete', function() { console.log(`Fastest is ${this.filter('fastest').map('name')}\n`); })
+    .run({ async: false });
+
+(new Benchmark.Suite())
+    .add('check', () => fc.assert(fc.property(fc.nat(3), () => { return true; })))
```
trollepierre: don't you want to use external lib like ramda? The last function looks like https://ramdajs.com/docs/#T

perf/index.js
```diff
@@ -0,0 +1,47 @@
+const Benchmark = require('benchmark');
```
trollepierre: I didn't get exactly what Benchmark will bring to fastcheck but why doesn't this code  need to be tested? ;-)

perf/index.js
dubzzz: `arb` stands for `arbitrary`, the whole documentation uses this denomination.

perf/index.js
dubzzz: This code in neither a production code nor a test code, it is used to run performance measurements on the library. Those 'hardcoded' values have been used to take into account cases with an array having a minimal size.

For the moment, specifying a minimal size forces the user to specify a maximal one. This problem should be fixed in a later iteration thanks to #89.

perf/index.js
dubzzz: I don't want to rely on an external library to do a simple `() => true` otherwise I would also import one to replace `3` by `magicLib.Three`...

perf/index.js
dubzzz: Benchmark is used to check and measure performances gaps. It has to be run and checked manually. It should be considered as a way to detect slow sections of the code.

PR #208, opened 2018-10-06
Bump pure-rand for better performances

dubzzz:


dubzzz:
pure-rand@1.4.1 replaces the usages of array destructuring by direct uses of `[idx]`. The performance improvement is important due to es3 transpilation that was adding lots of boilerplate to do the destructuration. Nonetheless it seems that even without transpilation, destructuring was more costly than direct accesses.

PR #226, opened 2018-10-24
Fix LoremIpsum import for esm 

hath995:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "fast-check",
-  "version": "1.7.0",
+  "version": "1.7.1",
```
dubzzz: Please do not bump manually, I will on my side with a specific commit.

src/check/arbitrary/LoremArbitrary.ts
```diff
@@ -1,18 +1,25 @@
 // tslint:disable-next-line:no-var-requires
```
dubzzz: I think we can remove this line now

hath995:
I think if we added a headless browser test to make sure it can run in the browser.

https://docs.travis-ci.com/user/chrome
https://developers.google.com/web/updates/2017/04/headless-chrome

I may give this a shot this weekend.

dubzzz:
I will publish the new version tomorrow morning (Paris time)

PR #233, opened 2018-11-03
Too many calls to [cloneMethod] during runs

### LINKED ISSUE #232: Too many calls to [cloneMethod] during runs
When running properties having cloneable (with [cloneMethod]) arbitraries, `fc.assert` automatically clones the value before using it for the first time.

It causes an extra clone which is not necessary.

Not exactly a bug as everything still works fine but it has to be fixed before releasing the feature.

dubzzz:
Fixes #232


dubzzz:
Merging the code as the red test does not seem to be related to the change.
Issue #234 has been opened to treat that flaky e2e test

PR #235, opened 2018-11-03
Better handling of commands

### LINKED ISSUE #230: Better handling of commands
Following #229, we can remove some hacks that have been used in order to speed-up shrinking and output of `commands`.

Basically `modelRun` was calling a `errorDetected` method onto the `Iterable` in order to snapshot the state (ran once or not?).

dubzzz:
Fixes #230


dubzzz:
**WARNING**

The combination of this PR and #229 introduced a major regression in the `fc.commands`. The shrinking of such commands no longer works as expected.

It basically stops to work after the first level of shrink.

PR #261, opened 2018-12-15
Fix: Unable to shrink mapped command

dubzzz:


dubzzz:
Thanks to this commit it is now possible to shrink mapped commands like:

```typescript
fc.nat().map(v => new MyCommand(v))
```

Before that commit, the shrinker would not have tried to shrink on the natural number generator - `fc.nat`.

PR #262, opened 2018-12-15
Fix: Too many shrinks for commands

dubzzz:


dubzzz:
Flaky failure, fixed by #260

PR #263, opened 2018-12-15
Default seed must use Math.random in addition of time

### LINKED ISSUE #257: Default seed must use Math.random in addition of time
Currently using time as the default seed, means that tests executed in parallel might be using the same seed. In order to prevent such behavior, the seed will be the combination of a source of random added to the time

dubzzz:
Fixes #257


dubzzz:
Flaky failure, fixed by #260

PR #292, opened 2019-01-25
Provide a flag to stop the test as soon as it fails

### LINKED ISSUE #225: Runners with a given seed and path can run more times than claimed
Despite the runner claiming that it runs once and shrinks 0 times, it can actually run several times.

For example:

```js
const fc = require('fast-check');

fc.assert(fc.property(fc.nat(), fc.nat(), (x, y) => {
    console.log('start');
    return x < y + 1000000;
}), { seed: 1539977860329, path: "6:1:0:0:0:0:0:0:2:3:7:10:12:13:15:17:19:16:16" });
```

Prints:

```
start
start
start
start
start
start
start
start
start
start
start
start
start
start
start
start
start
start
start
start

…

Error: Property failed after 1 tests
{ seed: 1539977860329, path: "6:1:0:0:0:0:0:0:2:3:7:10:12:13:15:17:19:16:16" }
Counterexample: [1000000,0]
Shrunk 0 time(s)
Got error: Property failed by returning false
```

Despite claiming that the property fails after 1 test and 0 shrinks, `start` is printed several times.

This makes debugging very difficult, because the user cannot easily log the failing run or drop into an interactive debugger on the failing run. This is exasperated by the last run possibly not being the counterexample.

dubzzz:
Fixes #225


dubzzz:
Codeclimate is complaining about the code of `readEndOnFailure` which is similar to `readUnbiased`.

```ts
private static readEndOnFailure = <T>(p?: Parameters<T>): boolean => p != null && p.endOnFailure === true;
private static readUnbiased = <T>(p?: Parameters<T>): boolean => p != null && p.unbiased === true;
```

A fix will be provided in a later PR

PR #320, opened 2019-02-25
Stringify have to handle boxed values and Set/Map

### LINKED ISSUE #318: Better error reporting for Set
When failing on an instance of Set, fast-check prints `[object Set]`.
Better reporting is required for built-in ;)

### LINKED ISSUE #319: Better error reporting for boxed values

dubzzz:
Fixes #318
Fixes #319


dubzzz:
Maybe add some checks for cyclic objects, so that it does not loop infinitely (today it will crash if cyclic)

PR #328, opened 2019-03-06
docs: Use assert()

gunar:
- Replaces `chai` for `assert`, in order to make it more readable for people without prior knowledge of the library.
- Also, replaces `.filter().length` with `.includes()`.

### DIFF
--- a/documentation/HandsOnPropertyBased.md
+++ b/documentation/HandsOnPropertyBased.md
@@ -113,9 +113,9 @@ test('should contain the same items', () => {
   fc.assert(
     fc.property(fc.array(fc.integer()), data => {
       const sorted = sort(data);
-      expect(sorted.length).toEqual(data.length);
+      assert(sorted.length === data.length);
       for (const item of data) {
-        expect(count(sorted, item)).toEqual(count(data, item));
+        assert(sorted.includes(item));
       }
     })
   );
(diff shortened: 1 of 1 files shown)


gunar:
If you want to go one step further I'd suggest suggesting [power-assert](https://www.npmjs.com/package/power-assert) for pretty diffs.

documentation/HandsOnPropertyBased.md
```diff
@@ -109,13 +109,12 @@ import * as fc from 'fast-check';
...
 test('should contain the same items', () => {
-  const count = (tab, element) => tab.filter(v => v === element).length;
   fc.assert(
     fc.property(fc.array(fc.integer()), data => {
       const sorted = sort(data);
-      expect(sorted.length).toEqual(data.length);
+      assert(sorted.length === data.length);
       for (const item of data) {
-        expect(count(sorted, item)).toEqual(count(data, item));
+        assert(sorted.includes(item));
```
dubzzz: The check is not totally equivalent. Before your change having `data = [1,2,2]` and `sorted = [1,1,2]` would have been wrong. Following your change it passes.

gunar: Shucks! You've clearly thought longer about this than I have. Thanks.

gunar:
Yeah that's what makes [power-assert](https://www.npmjs.com/package/power-assert) so powerful. It allows you to write your assertions/checks with plain JavaScript, while also giving you pretty useful diffs.

PR #343, opened 2019-04-09
Documentation site with docsify

spion:
Hi!

I tried to generate a cool looking documentation page - you can preview it here: https://spion.github.io/fast-check/

It uses api-extractor, a custom fork of docsify with improved full text search and a custom fork of Microsoft's API Documenter that generates fewer markdown files. It combines the documentation directory with the generated docs into a single site with two sections (Guides and API)

Let me know what you think?


spion:
Note that any package level items from the API can be hidden by marking them with "@internal". I may be able to add a feature to api-documenter to additionally not index certain items...

package.json
```diff
@@ -50,11 +52,14 @@
...
   },
   "devDependencies": {
+    "@microsoft/api-extractor": "^7.0.39",
     "@types/jest": "^24.0.11",
     "@types/node": "^11.11.3",
     "benchmark": "^2.1.4",
     "browserify": "^16.2.3",
     "coveralls": "^3.0.3",
+    "docsify-tools": "^1.0.8",
+    "gh-pages": "^2.0.1",
```
dubzzz: No need for the gh-pages package for the moment. I already handle the deployment of documentation in https://github.com/dubzzz/fast-check/blob/master/.travis.yml#L63:L76

package.json
```diff
@@ -50,11 +52,14 @@
     "pure-rand": "^1.6.2"
   },
   "devDependencies": {
```
dubzzz: Please commit the package-lock.json instead of the yarn.lock. I prefer to only handle one locking system for the moment.

package.json
```diff
@@ -31,7 +31,9 @@
     "format": "prettier --write \"**/*.{js,ts}\"",
     "format:check": "prettier --list-different \"**/*.{js,ts}\"",
     "lint": "tslint --project tsconfig.json --fix",
-    "lint:check": "tslint --project tsconfig.json"
+    "lint:check": "tslint --project tsconfig.json",
+    "docs:2": "mkdir -p docs/1-Guides && mkdir -p docs/2-API && cp documentation/*.md docs/1-Guides && api-extractor run --local && generate-ts-docs markdown -i docs/2-API -o docs/2-API && cp README.md docs/ && docsify-auto-sidebar",
```
dubzzz: Rename: replace the script docs of today by yours (and remove typedoc from the package.json). Your generated documentation looks clearly better ;)

package.json
```diff
@@ -31,7 +31,9 @@
...
+    "docs:2:publish":"gh-pages -d docs"
```
dubzzz: Remove: already handled by my travis script - We can maybe add it later (another review and commit)

package.json
dubzzz: As the new script relies on the *.d.ts files, you should edit the https://github.com/dubzzz/fast-check/blob/master/.travis.yml#L63:L76 script and add `npm run build:publish-types` before the existing `npm run docs` (and remove `mkdir` from the travis step as you already do that in the script itself).

spion:
Cool, I am glad you like it! I have a lot more work to do for sure, and I will address your feedback.

Just thought I'd mention some of the other things I haven't yet finished:

One thing I was wondering about is "Edit on github". For it to work at least for the guides, I'll need to tell the docsify-init script that the target directory for "edit on github" is not the same as the build directory where docsify stuff end up. I'll fix that.

Another thing that seemed to pop up are the links to the markdown files, they are absolute so they lead to the github site. But they can be made to work exactly the same on github and on the docs site if they are

1. relative (e.g. links to "documentation/File.md")
2. the end structure in docs will need to have the same subdirectories.

Currently the automatic sidebar generator sorts by the prefix (e.g. the folders 1-Guides 2-API become "Guides" and "API" in the sidebar) but maybe I can make it easier to customize to make it work better with this relative linking structure.

spion:
Ok I believe I managed to address all fedeback, updated the docs script and the travis publish script, the npm lockfile (removed yarn one), got rid of gh-pages package and script;

In addition, I moved the guides to a "1-Guides" subdirectory and updated all the links so they work on github, and in the documentation; in addition the absolute README links are transformed for the documentation version only so they continue to work on npm and github.

spion:
By the way, the search results are pretty cool :grinning: 
image

dubzzz:
Great job ;)
Thanks a lot for this contribution

PR #347, opened 2019-04-29
Towards cleaner arbitrary test suite

dubzzz:
Today arbitraries are checked using the helper function `genericHelper.isValidArbitrary`.

But this helper comes with very poor traces and proved difficult to troubleshoot when adding new arbitraries to fast-check.


dubzzz:
Will not be merged as-is. Switching to such helper utility would highly increase the time it takes to run the tests.

Nonetheless many paths are opened to make such utility viable in the future:
- reduce number of shrinks in the shrink tree (maybe not feasible) - we want to ensure we can converge to a minimal counterexample most of the time
- flag those tests as integartion tests and run both: units, integration and e2e tests suites

PR #353, opened 2019-05-05
Polished logo

treydavis:
I thought the logo could use a little polish. The font is Mozilla's Fira Sans.
logo
The file size is a little bigger by about 10kb but it looks nicer with better anti-aliasing. I can also include the svg files.


dubzzz:
Let's go for it 😍🤩 It looks clearly better 

Please include the svg file too. And add your name into CONTRIBUTORS.md

treydavis:
👍 Thanks! I included an svg of the icon by itself and the whole logo. I put all the images in a `logo` dir and updated the README to point at the new path.

dubzzz:
Commit merged ;)

We should maybe add a white border around the text (_or change the text color_) so that it can be visible when the png is displayed on a black background

PR #370, opened 2019-05-23
Add minimum and maximum validation to integer and nat

### LINKED ISSUE #293: integer and nat stuck in infinite loop when max < min
Such as
`fc.sample(fc.nat(-1))`
`fc.sample(fc.integer(0, -1))`

I recommend throwing an exception if max < min. `integer` could theoretically flip max and min if max < min, but this could lead to unexpected behavior, generating values greater than the given min or larger than the given max.

diegopedro94:
Fixes #293 

@dubzzz I noticed that BigInt and Float arbitraries have the same issue. If this solution is valid for you, I can take care of those cases in this PR as well.

### DIFF
--- a/src/check/arbitrary/IntegerArbitrary.ts
+++ b/src/check/arbitrary/IntegerArbitrary.ts
@@ -59,6 +59,7 @@ function integer(max: number): ArbitraryWithShrink<number>;
  */
 function integer(min: number, max: number): ArbitraryWithShrink<number>;
 function integer(a?: number, b?: number): ArbitraryWithShrink<number> {
+  if (a && b && a > b) throw new Error('fc.integer maximum value should be equal or greater than the minimum one');
   return b === undefined ? new IntegerArbitrary(undefined, a) : new IntegerArbitrary(a, b);
 }
 
@@ -79,6 +80,7 @@ function nat(): ArbitraryWithShrink<number>;
  */
 function nat(max: number): ArbitraryWithShrink<number>;
 function nat(a?: number): ArbitraryWithShrink<number> {
+  if (a && a < 0) throw new Error('fc.nat value should be greater than or equal to 0');
   return new IntegerArbitrary(0, a);
 }
 
--- a/src/check/arbitrary/IntegerArbitrary.ts
+++ b/src/check/arbitrary/IntegerArbitrary.ts
@@ -59,7 +59,8 @@ function integer(max: number): ArbitraryWithShrink<number>;
  */
 function integer(min: number, max: number): ArbitraryWithShrink<number>;
 function integer(a?: number, b?: number): ArbitraryWithShrink<number> {
-  if (a && b && a > b) throw new Error('fc.integer maximum value should be equal or greater than the minimum one');
+  if (a !== undefined && b !== undefined && a > b)
+    throw new Error('fc.integer maximum value should be equal or greater than the minimum one');
   return b === undefined ? new IntegerArbitrary(undefined, a) : new IntegerArbitrary(a, b);
 }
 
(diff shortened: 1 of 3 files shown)


src/check/arbitrary/IntegerArbitrary.ts
```diff
@@ -59,6 +59,7 @@ function integer(max: number): ArbitraryWithShrink<number>;
  */
 function integer(min: number, max: number): ArbitraryWithShrink<number>;
 function integer(a?: number, b?: number): ArbitraryWithShrink<number> {
+  if (a && b && a > b) throw new Error('fc.integer maximum value should be equal or greater than the minimum one');
```
dubzzz: I'd recommend to check the following instead:
`a !== undefined && b !== undefined && a > b`

Current code has an issue if one of `a` or `b` is 0.
I found that with a property based test (see below).


Eg.: `a = 1` and `b = 0`

src/check/arbitrary/IntegerArbitrary.ts
```diff
@@ -79,6 +80,7 @@ function nat(): ArbitraryWithShrink<number>;
  */
 function nat(max: number): ArbitraryWithShrink<number>;
 function nat(a?: number): ArbitraryWithShrink<number> {
+  if (a && a < 0) throw new Error('fc.nat value should be greater than or equal to 0');
```
dubzzz: Same, no bug here but check for undefined instead: `a !== undefined && a < 0`

test/unit/check/arbitrary/IntegerArbitrary.spec.ts
```diff
@@ -181,6 +184,9 @@ describe('IntegerArbitrary', () => {
     });
   });
   describe('nat', () => {
+    it('Should throw when the number is less than 0', () => {
```
dubzzz: Suggestion, a property based test:

```js
    it('Should throw when the number is less than 0', () =>
      fc.assert(
        fc.property(fc.integer(-0x80000000, -1), n => {
          expect(() => nat(n)).toThrowError();
        })
      ));
```

`-0x80000000` or `Number.MIN_SAFE_INTEGER`

test/unit/check/arbitrary/IntegerArbitrary.spec.ts
```diff
@@ -146,6 +146,9 @@ describe('IntegerArbitrary', () => {
           return -log2(-min) <= g && g <= log2(max);
         })
       ));
+    it('Should throw when minimum number is greater than maximum one', () => {
```
dubzzz: Suggestion, a property based test:

```js
    it('Should throw when minimum number is greater than maximum one', () =>
      fc.assert(
        fc.property(fc.integer(), fc.integer(), (a, b) => {
          fc.pre(a !== b);
          if (a < b) expect(() => integer(b, a)).toThrowError();
          else expect(() => integer(a, b)).toThrowError();
        })
      ));
```

This property detected the issue reported for: a = 1 and b = 0

src/check/arbitrary/IntegerArbitrary.ts
diegopedro94: Hi! In the new commit, I fixed the issue, but although I checked the code several times, I couldn't get why the property test was failing with `if (a && b && a > b) `but not with `if (a !== undefined && b !== undefined && a > b)`

Any ideas why that happened?

dubzzz: Hi! The reason of the bug is that when `b = 0` the condition inside the `if` statement becomes falsy: `if (0) { /* never executed */ }`. So when we have `a = 1` and `b = 0` the statement is equivalent to `if (1 && 0 && 1 > 0)` which evaluates to `false` because of the `&& 0 &&`.

dubzzz:
Thanks a lot for the contribution. Please let me know if you plan to do the same fix for float, double and bigint.

PR #377, opened 2019-06-13
Mutually recursive arbitrary builder

### LINKED ISSUE #374: Add easy mutually recursive arbitrary builder
Building tree-like structures in fast-check need to build more complex arbitraries relying on `chain` calls.
Ideally, it should be something easy to use.

Related tweet https://twitter.com/stefanobaghino/status/1131937110133493760?s=20

dubzzz:
Fixes #374

### DIFF
--- a/src/check/arbitrary/LetRecArbitrary.ts
+++ b/src/check/arbitrary/LetRecArbitrary.ts
@@ -39,8 +39,8 @@ export class LazyArbitrary extends Arbitrary<any> {
 }
 
 /** @hidden */
-function isLazyArbitrary(arb: Arbitrary<any>): arb is LazyArbitrary {
-  return arb.hasOwnProperty('underlying');
+function isLazyArbitrary(arb: Arbitrary<any> | undefined): arb is LazyArbitrary {
+  return arb !== undefined && arb.hasOwnProperty('underlying');
 }
 
 /**
@@ -66,20 +66,16 @@ export function letrec<T>(
     return lazyArbs[key]!;
   };
   const strictArbs = builder(tie as any);
-  for (const key in lazyArbs) {
-    if (!lazyArbs.hasOwnProperty(key)) {
+  for (const key in strictArbs) {
+    if (!strictArbs.hasOwnProperty(key)) {
       // Prevents accidental iteration over properties inherited from an object’s prototype
       continue;
     }
-    const arb = strictArbs[key];
-    const lazyArb = lazyArbs[key]!;
-    if (!arb) {
-      throw new Error(`Missing arbitrary for ${JSON.stringify(key)}`);
-    }
-    if (!isLazyArbitrary(lazyArb)) {
-      throw new Error(`Invalid lazy arbitrary for ${JSON.stringify(key)}`);
-    }
-    lazyArb.underlying = arb;
+
+    const lazyAtKey = lazyArbs[key];
+    const lazyArb = isLazyArbitrary(lazyAtKey) ? lazyAtKey : new LazyArbitrary(key);
+    lazyArb.underlying = strictArbs[key];
+    lazyArbs[key] = lazyArb;
   }
   return strictArbs;
 }
--- a/src/check/arbitrary/LetRecArbitrary.ts
+++ b/src/check/arbitrary/LetRecArbitrary.ts
@@ -2,7 +2,7 @@ import { Random } from '../../random/generator/Random';
 import { Arbitrary } from './definition/Arbitrary';
 import { Shrinkable } from './definition/Shrinkable';
 
-class LazyArbitrary extends Arbitrary<any> {
+export class LazyArbitrary extends Arbitrary<any> {
   private static readonly MaxBiasLevels = 5;
   private numBiasLevels = 0;
   underlying: Arbitrary<any> | null = null;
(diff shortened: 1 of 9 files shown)


dubzzz:
Tasks to be tackle before merging this PR:

- [ ] Various tests to be added to cover the feature
- [ ] Keep bias support
- [ ] Investigate how too easily avoid `Maximum call stack size exceeded` errors
- [ ] Investigate alternative implementation based on `chain` if possible

 
I made it work with arbitraries defined as:

```js
const { node, leaf } = fc.letrec(tie => ({
  node: fc.tuple(
    fc.frequency({ arbitrary: tie('leaf'), weight: 5 }, { arbitrary: tie('node'), weight: 1 }),
    fc.frequency({ arbitrary: tie('leaf'), weight: 5 }, { arbitrary: tie('node'), weight: 1 })
  ),
  leaf: fc.nat()
}));
```

Or even:

```js
const { node, leaf } = fc.letrec(tie => ({
  node: fc.tuple(
    fc.oneof(tie('leaf'), tie('node')),
    fc.oneof(tie('leaf'), tie('node'))
  ),
  leaf: fc.nat()
}));
```

But it failed (sometimes) with definitions like:

```js
const { node, leaf } = fc.letrec(tie => ({
  node: fc.tuple(
    fc.oneof(tie('leaf'), tie('node')),
    fc.oneof(tie('leaf'), tie('node')),
    fc.oneof(tie('leaf'), tie('node'))
  ),
  leaf: fc.nat()
}));
```

With `Maximum call stack size exceeded` errors. Ideally the framework should have prefered `leaf` over `node` (artificially done in the first snippet).

dubzzz:
I also tried with another signature for `letrec`. The advantage is that I get better typings for `tie` but I lose the types for arbitraries relying on one of the values coming from `tie`. (see A and B)

```ts
declare const Int: Arbitrary<number>;
declare const Str: Arbitrary<string>;
declare function MyArr<T>(a: Arbitrary<T>): Arbitrary<T[]>;

declare function letrecA<T>(
  def: { [K in keyof T]: (d: { [KK in keyof T]: Arbitrary<T[KK]> }) => Arbitrary<T[K]> }
): { [K in keyof T]: Arbitrary<T[K]> };

const outA = letrecA({
  a: () => Int,
  b: ({ a }) => MyArr(a),
  c: () => Str,
  d: ({ c }) => c
});

declare function letrecB<T>(
  def: { [K in keyof T]: (tie: <KK extends keyof T>(k: KK) => Arbitrary<T[KK]>) => Arbitrary<T[K]> }
): { [K in keyof T]: Arbitrary<T[K]> };

const outB = letrecB({
  a: () => Int,
  b: tie => MyArr(tie('a')),
  c: () => Str,
  d: tie => tie('c')
});

declare function letrecC<T>(
  builder: (tie: (key: string) => Arbitrary<any>) => { [K in keyof T]: Arbitrary<T[K]> }
): { [K in keyof T]: Arbitrary<T[K]> };

const outC = letrecC(tie => ({
  a: Int,
  b: MyArr(tie('a')),
  c: Str,
  d: tie('c')
}));
```

dubzzz:
Here are various tries to implement `letrec` with no built-ins provided by fast-check:

❌ Not compiling
```js
// Not compiling because node is used before being initialized
const leaf = fc.nat();
const node = fc.tuple(fc.oneof(node, leaf), fc.oneof(node, leaf));
```

❌ Not running
```js
// Unable to instantiate node because it instantiates another one which does the same..
const leaf = () => fc.nat();
const node = () => fc.tuple(fc.oneof(node(), leaf()), fc.oneof(node(), leaf()));
```

✔️ Working
```js
// (+) Explicit terminal case
// (+) Ability to use depth to parametrize the sub arbitraries
// (+) No LazyArbitrary
// (-) Has to pass the n parameter to sub arbitraries
const lazy = (f, def) => {
  let previous = {};
  return cur => {
    const n = (cur || 0) + 1;
    if (cur >= 5) return def;
    if (!previous.hasOwnProperty(n)) previous[n] = f(n);
    return previous[n];
  };
};
const leaf = () => fc.nat();
const node = lazy(n => fc.tuple(fc.oneof(node(n), leaf()), fc.oneof(node(n), leaf())), leaf());
```

src/check/arbitrary/LetRecArbitrary.ts
```diff
@@ -0,0 +1,69 @@
+import { Random } from '../../random/generator/Random';
+import { Arbitrary } from './definition/Arbitrary';
+import { Shrinkable } from './definition/Shrinkable';
+
+export class LazyArbitrary extends Arbitrary<any> {
```
dubzzz: Missing jsdoc

src/check/arbitrary/LetRecArbitrary.ts
```diff
@@ -0,0 +1,69 @@
...
+
+    --this.numBiasLevels;
+    if (this.numBiasLevels !== numBefore) {
+      throw new Error(`Mismatched bias level for ${JSON.stringify(this.name)}`);
+    }
+    return biasedArb;
+  }
+}
+
+function isLazyArbitrary(arb: Arbitrary<any>): arb is LazyArbitrary {
```
dubzzz: Missing jsdoc

src/check/arbitrary/LetRecArbitrary.ts
```diff
@@ -0,0 +1,69 @@
...
+  return arb.hasOwnProperty('underlying');
+}
+
+export function letrec<T>(
```
dubzzz: Missing jsdoc

PR #388, opened 2019-07-03
Fix type inferrence bug in modelRun

hamiltop:
Example of bug:
```
interface Model {
  foo: string;
  bar: number;
}

type Real = number;

class Command implements fc.Command<Model, Real> {
  check(m: Model) {
    return true;
  }
  run(m: Model, r: Real) {}
  toString() {
    return "";
  }
}

it("should fail typecheck", () => {
  return fc.assert(
    fc.property(fc.commands([fc.constant(new Command())]), cmds => {
      const c = 5;
      return fc.modelRun(
        () => ({ model: { foo: "what about bar?" }, real: c }),
        cmds
      );
    })
  );
});
```
This should ideally not type-check, because `{foo: "what about bar?"}` does not meet the requirement s for the `Model` interface. The generic type parameters for `modelRun` are inferred to be `<{foo: string}, number>` and conveniently `Command<Model, number>` structurally matches `Command<{foo: string", number>` which is why the typescript misses this.

In this PR, I introduce another type parameter for `modelRun` that explicitly extends the other type parameter. This forces the setup function to structurally match the type parameter in the commands rather than the other way around.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -6,3 +6,4 @@ People who commit code to the project are encouraged to add their names here. Pl
 - [Nicolas Dubien](https://github.com/dubzzz)
 - [Gjorgji Kjosev](https://github.com/spion)
 - [Trey Davis](https://github.com/treydavis)
+- [Peter Hamilton](https://github.com/hamiltop)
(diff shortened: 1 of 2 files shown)


dubzzz:
Please add yourself into https://github.com/dubzzz/fast-check/blob/master/CONTRIBUTORS.md

dubzzz:
Commit merged, will be integrated into the coming soon release - `1.16.0`

PR #393, opened 2019-07-09
Support all the range of valid ipV4

dubzzz:


dubzzz:
According to a recent thread on Twitter, it seems that fast-check already support all the range of IPv6 addresses: https://twitter.com/ndubien/status/1161190040212516869?s=19

dubzzz:
Maybe an update missing for `withExtendedIp`

dubzzz:
IPv6 can still be treated by adding 0 or mixing upper case and lower case within the same IP address

It will be treated as a FAR

src/check/arbitrary/IpArbitrary.ts
```diff
@@ -16,6 +17,38 @@ function ipV4(): Arbitrary<string> {
...
+ */
+function ipV4Extended(): Arbitrary<string> {
+  const natRepr = (maxValue: number) =>
+    tuple(constantFrom('dec', 'oct', 'hex', 'HEX'), nat(maxValue)).map(([style, v]) => {
+      switch (style) {
+        case 'oct':
+          return `0${Number(v).toString(8)}`;
+        case 'hex':
+          return `0x${Number(v).toString(16)}`;
+        case 'HEX':
```
dubzzz: Not needed anymore: the uppercase can be handled by the feature #426

PR #411, opened 2019-08-18
Add UUID arbitrary

### LINKED ISSUE #404: UUID arbitrary
## 🚀 Feature Request

UUID arbitrary - https://en.wikipedia.org/wiki/Universally_unique_identifier

## Motivation

UUID are more and more frequent in the industry. They might be a built-in of fast-check.

## Example

Here are two possible implementations for uuid, but both of them suffer from issues:

```js
const uuidArb = fc.tuple(
  fc.hexaString(8, 8),
  fc.hexaString(4, 4),
  fc.hexaString(4, 4),
  fc.hexaString(4, 4),
  fc.hexaString(12, 12),
).map(([a, b, c, d, e]) => `${a}-${b}-${c}-${d}-${e}`)
// Drawback: slow shrink - will shrink one digit at a time
```

```js
const upperBound = 16n ** 32n - 1n;
const uuidArb = fc.bigUint(upperBound).map(bint => {
  const st = bint.toString(16);
  const s = "0".repeat(32 - st.length) + st;
  return `${s.substring(0, 8)}-${s.substring(8, 12)}-${s.substring(
    12,
    16
  )}-${s.substring(16, 20)}-${s.substring(20, 32)}`;
});
// Drawback: depends on BigInt
```

dubzzz:
## Why is this PR for?

Fixes #404 

## In a nutshell

✔️ New feature
❌ Fix an issue
❌ Documentation improvement
❌ Other: *please explain*

## Potential impacts

None.


dubzzz:
The implementation of `uuidExtended` could have been:

```js
export function uuidExtended() {
  return tuple(uuid(), nat(0xffffffff)).map(([g, uppers]) => {
    let newUuid = '';
    for (let idx = 0, mask = 1; idx !== 36; ++idx, mask <<= 1) {
      if (g[idx] === '-') {
        newUuid += g[idx];
        ++idx;
      }

      if (uppers & mask) newUuid += g[idx].toUpperCase();
      else newUuid += g[idx];
    }
    return newUuid;
  });
}
```

But it would have meant that we may (often) encounter multiple times the same uuid in the shrink path.

dubzzz:
Remove uuidExtended as it can easily be produced by using the feature that will be introduced by #426

PR #418, opened 2019-08-20
Infinite loop when path is one level deep and run succeeds

### LINKED ISSUE #415: Infinite loop when trying to replay failure
## 🐛 Bug Report

I'm trying to reproduce a failure with a specific `seed` and `path`, but the unit test just enters an infinite loop. The most intriguing thing is that jest doesn't timeout :thinking: 

Apparently, if I change the path to `2:0`, it doesn't enter an infinite loop. However, a wrong path shouldn't prevent jest from timing out.

## To Reproduce

```typescript
import fc from 'fast-check';

describe('suite', () => {
    it('unit test', () => {
        const property = fc.property(fc.nat(), num => {
            console.log(num);
            console.log('******');
        });

        fc.assert(property, { seed: -1386863032, path: '2', endOnFailure: true });
    });
});
```

## Expected behavior

It should only run the property `numRuns` times and it shouldn't prevent jest from timing out.

## Your environment

| fast-check            |     1.16.0       |
| node                  |     10.14.1       |
| TypeScript*           |      3.4.5      |

*Only for TypeScript's users

dubzzz:
## Why is this PR for?

Fixes #415

## In a nutshell

❌ New feature
✔️ Fix an issue
❌ Documentation improvement
❌ Other: *please explain*

(✔️: yes, ❌: no)

## Potential impacts

None.

### DIFF
--- a/src/check/runner/Runner.ts
+++ b/src/check/runner/Runner.ts
@@ -100,7 +100,7 @@ function check<Ts>(rawProperty: IProperty<Ts>, params?: Parameters<Ts>) {
   const property = decorateProperty(rawProperty, qParams);
   const generator = toss(property, qParams.seed, qParams.randomType, qParams.examples);
 
-  const maxInitialIterations = qParams.path.length === 0 ? qParams.numRuns : -1;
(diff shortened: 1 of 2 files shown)


dubzzz:
The regression has been introduced in the version 1.3.x

Issue #423, opened 2019-09-05
skipAllAfterTimeLimit throws an error instead of skipping tests


ZeRego:
## 🐛 Bug Report

skipAllAfterTimeLimit throws an error instead of skipping the remaining tests like the name suggests

## To Reproduce

Steps to reproduce:

```describe('skipAllAfterTimeLimit', () => {
    it(‘should skip tests’, async () => {
        jest.useRealTimers();
        const property = fc.asyncProperty(
            fc.constant(''),
            async () => {
                await new Promise(resolve => setTimeout(resolve, 100));
                expect(true).toBe(true);
            }
        );
        await fc.assert(property, {
            numRuns: 10,
            skipAllAfterTimeLimit: 1,

        });
    });
});
```

## Expected behavior

It should skip all remaining tests instead of throwing error. 
We could have a parameter, similar to maxSkipsPerRun, to specify the max amount of tests it can skip.

## Your environment

| Packages / Softwares  | Version(s) |
| --------------------- | ---------- |
| fast-check            |       1.16.2     |
| node                  |     10.14.1       |
| TypeScript          |     3.4.5       |

dubzzz:
@ZeRego Thanks a lot for the report. I indeed reproduced the issue immediately. I fixed _part_ of the issue and added the tests to avoid this issue to come back in the future.

---

I initially added the flag `skipAllAfterTimeLimit` for very long model based tests applied onto UI. The idea was just to avoid running into _infinite_ shrinks in my CI (maybe hours of shrink when I launched it on really complex sets of commands). After this delay, instead of running the tests, fast-check does as if you shrunk but marked the test with precondition failure (as `fc.pre` does).

Basically the feature adds a precondition on top of all the properties:
```js
fc.pre(notExpired());
// here the property code
```

---

The need you have seem to be a bit different: you want to stop the run where it is and report its curent status. I definitely have to add such feature.

dubzzz:
I'll release the fix into 1.15.x and 1.16.x as soon as possible, so that the feature does what it says. As told above I'll need to add another feature that just stops where the runner is and reports its current status (no matter if it still have to generate test cases to fulfill `numRuns` constraint).

dubzzz:
I double-checked the code this morning, the feature you asked should be quite easy to add.

[...]

PR #428, opened 2019-09-09
✨ Implement interruptAfterTimeLimit

dubzzz:
## Why is this PR for?

Related to #423

This feature will make fast-check users able to interrupt the execution of a test while it stills have job to do. Assert will report the status as if the run was supposed to stop there. If not test failed the status will be success (even if the assert has not reached the expected number of runs).

**This is a draft for a potential futur feature, not sure if the feature will reach the master**

## In a nutshell

✔️ New feature
❌ Fix an issue
❌ Documentation improvement
❌ Other: *please explain*

(✔️: yes, ❌: no)

## Potential impacts

None.


dubzzz:
**What behaviour do we want in the following scenario? Are there some cases where we would like to consider it as a failure (`skipAllAfterTimeLimit` already partially covers part of the problem).**

The user asked for 100 runs by setting the value of `numRuns` but only 20 got executed (with no failure detected) before the interrupt signal.
With the current implementation, the run will be marked as successful.

src/fast-check-default.ts
```diff
@@ -97,6 +98,7 @@ export {
   assert,
   // pre conditions
   pre,
+  interrupt,
```
dubzzz: Not sure of the need for such feature. Not sure for the naming.

ZeRego:
When using `markInterruptAsFailure: true` can the error message contain all seeds and paths used ? Providing just the seed and path from the last run won't be enough because the problem might be in one of the previous runs.
Otherwise the developers won't have a way to reproduce the issue.

dubzzz:
What's the best option? `markInterruptAsFailure` or `markInterruptAsSuccess`? 🤔

I think once we answer this question we can merge the PR and move on to add seed, path and others in the report 😁

src/fast-check-default.ts
dubzzz: Reverting it as it might be added later in another PR

dubzzz:
@ZeRego The flag is now released in version 1.19.0 of fast-check ;)

99Edorde:
> ## Why is this PR for?
> 
> 
> 
> Related to #423
> 
> 
> 
> This feature will make fast-check users able to interrupt the execution of a test while it stills have job to do. Assert will report the status as if the run was supposed to stop there. If not test failed the status will be success (even if the assert has not reached the expected number of runs).
> 
> 
> 
> **This is a draft for a potential futur feature, not sure if the feature will reach the master**
> 
> 
> 
> ## In a nutshell
> 
> 
> 
> ✔️ New feature
> 
> ❌ Fix an issue
> 
> ❌ Documentation improvement
> 
> ❌ Other: *please explain*
> 
> 
> 
> (✔️: yes, ❌: no)
> 
> 
> 
> ## Potential impacts
> 
> 
> 
> None.
> 
>

PR #436, opened 2019-09-26
any -> unknown in exported signatures for object/anything arbitraries

### LINKED ISSUE #435: safer typings for anything/object
## 🚀 Feature Request

Currently `anything()`/`object()` arbitraries are typed as `Arbitrary<any>` (e.g. https://github.com/dubzzz/fast-check/blob/master/src/check/arbitrary/ObjectArbitrary.ts#L211).
This allows to inadvertently pass their generated values to any function, making definition of the property itself more prone to errors.

Switching to `Arbitrary<unknown>` would make this safer

## Motivation

Please outline the motivation for the proposal.

## Example

this is valid code with current typings:
```ts
declare function sum(a: number, b: number): number;

fc.integer().chain(a => fc.object().map(b => sum(a, b)));
```

giogonzo:
## Why is this PR for?

closes #435 

I only changed the exported definitions signatures in `object` arbitrary to use `unknown` instead of `any`. I believe there might be other instances where this change could be done (I didn't check yet), this could be a first step

## In a nutshell

❌ New feature
✔️ Fix an issue
❌ Documentation improvement
❌ Other: *please explain*

(✔️: yes, ❌: no)

## Potential impacts

Typings impact

This might be breaking for some users, but for the better 😎

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -4,6 +4,7 @@ People who commit code to the project are encouraged to add their names here. Pl
 - [Aaron Elligsen](https://github.com/hath995)
 - [Diego Pedro](https://github/diegopedro94)
 - [Nicolas Dubien](https://github.com/dubzzz)
+- [Giovanni Gonzaga](https://github.com/giogonzo)
 - [Gjorgji Kjosev](https://github.com/spion)
 - [Thibaut Boulard](https://github.com/volrk)
 - [Trey Davis](https://github.com/treydavis)
--- a/src/check/arbitrary/ObjectArbitrary.ts
+++ b/src/check/arbitrary/ObjectArbitrary.ts
@@ -17,7 +17,7 @@ import { tuple } from './TupleArbitrary';
 export class ObjectConstraints {
   constructor(
     readonly key: Arbitrary<string>,
-    readonly values: Arbitrary<any>[],
+    readonly values: Arbitrary<unknown>[],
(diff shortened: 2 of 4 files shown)


dubzzz:
Great job :) Maybe add yourself into the Contributors (I need to change this process but have taken the time yet) 

PR looks good to me.

I'll just try to confirm it will not break the unit tests of jest, js-yaml and query-string (both use fc.object or one of its variations). If it does, it will not be a real issue, it's more to fix them in advance.

giogonzo:
> I'll just try to confirm it will not break the unit tests of jest, js-yaml and query-string

😮I didn't know jest was using fast-check 🚀 

Let me know if you need any help

giogonzo:
@dubzzz actually, both js-yaml and query-string are written in JS (tests included) and jest is using some TS, but apparently not using `ts-jest` (i.e. they don't typecheck when running tests, so they will never fail because of this change).
In any case, I run jest tests with this branch of fast-check, and I'm only getting a few failures that are unrelated to this change (which to be clear, is only affecting users who type-check their tests)

dubzzz:
Let's merge it. Thanks a lot for your checks on other repos, change looks safe. 

Sorry for the delay, I'm just starting a new dev position :)

dubzzz:
Sorry for the delay, I'll publish a new minor tonight so that you can enjoy your fix 👍

PR #439, opened 2019-10-06
Provide global parameters definition

### LINKED ISSUE #402: Provide global parameters definition
## 🚀 Feature Request

Provide a way to define parameters on `fc` rather than defining them in each `fc.check` or `fc.assert`.

## Motivation

I'm using the same parameter `{ verbose: true }` for each call of `fc.check` and I'd like to provide it only once for all my tests

## Example

```js
fc.setParameter({ verbose: true })

fc.assert(
  fc.property(
    fc.integer(), num => { throw new Error(num) }
  ),
  // { verbose: true } No need to define the parameter here
)
```
Right now to achieve this behavior, I'm doing something like this:

```js
  const fcCheck = fc.check
  const fcAssert = fc.assert

  function wrapCheck (property, param = { verbose: true }) {
    return fcCheck(property, param)
  }

  function wrapAssert (property, param = { verbose: true }) {
    fcAssert(property, param)
  }

  fc.check = wrapCheck
  fc.assert = wrapAssert
```

dubzzz:
## Why is this PR for?

Fixes #402

## In a nutshell

✔️ New feature
❌ Fix an issue
❌ Documentation improvement
❌ Other: *please explain*

(✔️: yes, ❌: no)

## Potential impacts

None.

Very very limited risk to change the behaviour of `fc.sample(arb, numRuns)` or `fc.statistics(arb, map, numRuns)`. Tests have been added to cover such problems.


dubzzz:
There are still missing features to add before merging this PR:
- document an easy way to set those global parameters on classical test runners

PR #445, opened 2019-10-09
🚧✨ Add `fc.scenario` arbitrary

### LINKED ISSUE #424: Generate series of values that are dependent on the previous value
## 💬 Question and Help

I have a reducer that maintains a set of numbers stored as an array.  I'm trying to generate a series of actions to test the reducer with.  (This is a simplified example of the actual problem)

```
[1, 2]
```
And there are the following actions:
```
{ type: "add", value: number }
{ type: "remove", value: number }
```
I want to generate a series of actions to dispatch, where the actions are dependent on the state _after_ dispatching the previous action.  For example, two remove actions for the same value would be invalid, because the first action would have already removed the value.

Valid Examples:
```
initial state: [1, 2]
// Can remove either of the start values
[{ type: "remove", value: 1 }, { type: "remove", value: 2 }]
// even though 3 is not in the initial state, after it's added it becomes a valid remove target
[{ type: "add", value: 3 }, { type: "remove", value: 3 }]
```

Invalid Examples:
```
initial state: [1, 2]
// Should not be able to remove a value not in state
[{ type: "remove", value: 3 }]
// Should not be able to remove a value twice
[{ type: "remove", value: 2 }, { type: "remove", value: 2 }]
```

I can generate an Arbitrary for valid actions based on a state, I just can't figure out how to create an array where each element is dependent on the value of the previous.

I've poked around Model based testing / Commands, but it seems like it has the same problem, where you must define all your commands up front.

So my question is:  Is this use case something the library currently supports?  If not, is it something that could be added?  I'd be happy to help out if so!

dubzzz:
## Why is this PR for?

Fixes #424

## In a nutshell

✔️ New feature
❌ Fix an issue
❌ Documentation improvement
❌ Other: *please explain*

(✔️: yes, ❌: no)

## Potential impacts

None.


dubzzz:
Deleting the branch for now, but here is the code, I initially drafted for it:

```ts
import { Arbitrary } from './definition/Arbitrary';
import { constant } from './ConstantArbitrary';
import { oneof } from './OneOfArbitrary';

export function scenario<TModel, TAction>(
  initialModel: TModel,
  modelUpdater: (model: TModel, action: TAction) => TModel,
  eligibleActions: (model: TModel) => Arbitrary<TAction>[],
  scenarioLength?: number
): Arbitrary<TAction[]> {
  const finalScenarioLength = typeof scenarioLength === 'number' ? scenarioLength : 10;

  const extendScenario = (current: {
    model: TModel;
    scenario: TAction[];
  }): Arbitrary<{ model: TModel; scenario: TAction[] }> => {
    // Extract eligible actions
    const actions = scenario.length < finalScenarioLength ? eligibleActions(current.model) : [];
    if (actions.length === 0) return constant(current);
    return oneof(...eligibleActions(current.model)).chain((a) =>
      extendScenario({ model: modelUpdater(current.model, a), scenario: [...current.scenario, a] })
    );
  };

  return constant({ model: initialModel, scenario: [] as TAction[] })
    .chain(extendScenario)
    .map(({ scenario }) => scenario);
}

/*const myScenarioArb = scenarioArb(
  [],
  (m, a) => {
    // simplified reducer
    switch (a.type) {
      case 'add':
        return [...m, a.value];
      case 'removeAll':
        return m.filter(v => v !== a.value);
      case 'clear':
        return [];
    }
  },
  m => {
    const eligible = [
      fc.record({ type: fc.constant('add'), value: fc.hexa() }),
      fc.record({ type: fc.constant('clear') })
    ];
    if (m.length > 0) eligible.push(fc.record({ type: fc.constant('removeAll'), value: fc.constantFrom(...m) }));
    return eligible;
  }
);
fc.sample(myScenarioArb, 1);*/
```

Issue #461, opened 2019-10-26
Export Property, AsyncProperty and IProperty types


tinydylan:
## 🚀 Feature Request

Export the Property, AsyncProperty and IProperty types, so calling code can reference these types.

## Motivation

I am working on a custom test framework, and wish to make use of fast-check properties. I need to be able to accept a Property or AsyncProperty as a parameter, and run it in a special way, in order to wire up the pass/fail callbacks correctly.

## Example

My testing framework uses "Steps" which are asynchronous functions with logs. Below is how I would need to wire it up to run AsyncProperties. Currently, I can't refer to the AsyncProperty type, as it is not exported.

```
import { TestLogs } from '@ephox/agar';
import * as fc from 'fast-check';

export interface Step<T, U> {
  runStep: (value: T, next: (value: U, logs: TestLogs) => void, die: (err: any, logs: TestLogs) => void, logs: TestLogs) => void;
}

export const foo = <T, P>(name: string, makeProp: (t: T) => fc.AsyncProperty<P>): Step<T, T> =>
  ({
    runStep: (value, next, die, logs) => {
      fc.assert(makeProp(value)).then(() => next(value, logs)).catch((r) => {
        die(r, logs);
      });
    }
  });
```

tinydylan:
What about something like this?

```
export interface IAsyncProperty {
}

class AsyncProperty implements IAsyncProperty {
}

```

tinydylan:
BTW - thanks for your help! I'm loving fast-check so far. My team is using it to replace our use of JSVerify in the TinyMCE codebase.

dubzzz:
You might be interested in https://github.com/dubzzz/jsverify-to-fast-check to help you migrate from jsverify

dubzzz:
@tinydylan Just opened a PR to expose interfaces for properties. Would it fit your needs? - see #463

tinydylan:
Yeah, that looks good to me.

dubzzz:
Sorry for the delay. Version 1.19.0 of fast-check, should fit your needs

Please re-open the issue in case it does not

PR #463, opened 2019-11-03
🏷️ Adapt and expose IRawProperty, IProperty and IAsyncProperty types

dubzzz:
## Why is this PR for?

Adapt and expose `IRawProperty`, `IProperty` and `IAsyncProperty` types.

Related to #461

## In a nutshell

✔️ New feature
✔️ Fix an issue
❌ Documentation improvement
❌ Other: *please explain*

(✔️: yes, ❌: no)

## Potential impacts

No impact expected *BUT* internal changes:
- `fc.assert` and `fc.check` have been adapted to rely on `IRawProperty`, `IProperty` and `IAsyncProperty` instead of `Property`* or `AsyncProperty`* to guess its signature BUT no impact is expected
- `IProperty<Ts>`* has been renamed and updated `IRawProperty<Ts, IsAsync>`
- Two new interfaces for properties `IProperty` and `IAsyncProperty` - simple syntaxic sugar over `IRawProperty`

\* `Property` internal implementation for `fc.property` - implements newly created `IProperty`
\* `AsyncProperty` internal implementation for `fc.asyncProperty` - implements newly created `IAsyncProperty`
\* `IProperty` was an internal interface

### DIFF
--- a/src/check/property/IRawProperty.ts
+++ b/src/check/property/IRawProperty.ts
@@ -9,7 +9,7 @@ import { PreconditionFailure } from '../precondition/PreconditionFailure';
  * - Arbitraries: how to generate the inputs for the algorithm
  * - Predicate: how to confirm the algorithm succeeded?
  */
-export interface IProperty<Ts, IsAsync extends boolean = boolean> {
+export interface IRawProperty<Ts, IsAsync extends boolean = boolean> {
   /**
    * Is the property asynchronous?
    *
--- a/src/check/property/SkipAfterProperty.ts
+++ b/src/check/property/SkipAfterProperty.ts
@@ -3,10 +3,10 @@ import { PreconditionFailure } from '../precondition/PreconditionFailure';
 import { IProperty } from './IProperty';
 
 /** @hidden */
-export class SkipAfterProperty<Ts> implements IProperty<Ts> {
+export class SkipAfterProperty<Ts, IsAsync extends boolean> implements IProperty<Ts, IsAsync> {
   private skipAfterTime: number;
   constructor(
-    readonly property: IProperty<Ts>,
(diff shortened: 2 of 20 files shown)


src/check/property/SkipAfterProperty.ts
```diff
@@ -17,7 +17,12 @@ export class SkipAfterProperty<Ts> implements IProperty<Ts> {
   generate = (mrng: Random, runId?: number) => this.property.generate(mrng, runId);
   run = (v: Ts) => {
     if (this.getTime() >= this.skipAfterTime) {
-      return new PreconditionFailure(this.interruptExecution);
+      const preconditionFailure = new PreconditionFailure(this.interruptExecution);
+      if (this.isAsync()) {
+        return Promise.resolve(preconditionFailure) as any; // IsAsync => Promise<PreconditionFailure | string | null>
+      } else {
+        return preconditionFailure as any; // !IsAsync => PreconditionFailure | string | null
+      }
```
dubzzz: I need to think about it. The best solution might just be to say that we return a `IRawProperty<Ts>` with no more details on the sync vs async behaviour. 🤔

src/check/property/IRawProperty.ts
```diff
@@ -9,13 +9,13 @@ import { PreconditionFailure } from '../precondition/PreconditionFailure';
...
  */
-export interface IProperty<Ts> {
+export interface IRawProperty<Ts, IsAsync extends boolean = boolean> {
   /**
    * Is the property asynchronous?
    *
    * true in case of asynchronous property, false otherwise
    */
-  isAsync(): boolean;
+  isAsync(): IsAsync;
```
dubzzz: Maybe: `isAsync(): this is IRawProperty<Ts, true>`

PR #489, opened 2019-12-04
Allow looser return types for AsyncProperty beforeEach and afterEach

### LINKED ISSUE #480: Looser return types for before/afterEach and non-async functions with asyncProperty
## 🚀 Feature Request

This is sort-of a two part feature request:

1. It would be nice if the return type of the `beforeEach` and `afterEach` hooks didn't need to return `void`, and instead could return `unknown`. This would be similar to what [Jest does](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/jest/index.d.ts#L322).
2. It would be nice if any of the functions of `asyncProperty` (the predicate, beforeEach hook and afterEach hook) didn't have to be async.

## Motivation

Both of these features would reduce the amount of code that needs to be written and improve the usability of the library.

For part 1, the main motivation is so that simple before/after hooks can be one line:

```ts
// This:
property(
  // ...
).afterEach(() => jest.clearAllMocks());

// Instead of this:
property(
  // ...
).afterEach(() => { // Needs braces as the return type must be void
  jest.clearAllMocks();
});
```

The motivation for part 2 is similar, but is only really applicable if part 1 is implemented too:

```ts
// This:
asyncProperty(
  // ...
).afterEach(() => jest.clearAllMocks());

// Instead of this:
asyncProperty(
  // ...
).afterEach(async () => {
  jest.clearAllMocks();
});
```

It's a small improvement, but it is nicer :slightly_smiling_face:.

Part 2 could also allow for a non-async predicate, while having async before/after each functions:

```ts
asyncProperty(
  fc.nat(),
  (x) => // Non-async predicate
).afterEach(async () => {
  // Some async code
});
```

There might not be many use cases for this though, but it would be easy to support.

sommd:
## Why is this PR for?

Closes #480

## In a nutshell

✔️ New feature
✔️ Fix an issue
❌ Documentation improvement
❌ Other: *please explain*

(✔️: yes, ❌: no)

## Potential impacts

Should be a non-breaking change, since the type is looser than it was before.

### DIFF
--- a/src/check/property/AsyncProperty.generic.ts
+++ b/src/check/property/AsyncProperty.generic.ts
@@ -9,17 +9,17 @@ import { IRawProperty, runIdToFrequency } from './IRawProperty';
  */
 export interface IAsyncProperty<Ts> extends IRawProperty<Ts, true> {}
 
+type HookFunction = (() => Promise<unknown>) | (() => void);
+
 /**
  * Asynchronous property, see {@link IAsyncProperty}
  *
  * Prefer using {@link asyncProperty} instead
  */
 export class AsyncProperty<Ts> implements IAsyncProperty<Ts> {
-  static dummyHook: () => Promise<void> = async () => {
-    return;
-  };
-  private beforeEachHook: () => Promise<void> = AsyncProperty.dummyHook;
-  private afterEachHook: () => Promise<void> = AsyncProperty.dummyHook;
+  static dummyHook: HookFunction = () => {};
+  private beforeEachHook: HookFunction = AsyncProperty.dummyHook;
+  private afterEachHook: HookFunction = AsyncProperty.dummyHook;
   constructor(readonly arb: Arbitrary<Ts>, readonly predicate: (t: Ts) => Promise<boolean | void>) {}
   isAsync = () => true as const;
   generate(mrng: Random, runId?: number): Shrinkable<Ts> {
@@ -45,7 +45,7 @@ export class AsyncProperty<Ts> implements IAsyncProperty<Ts> {
    * Define a function that should be called before all calls to the predicate
    * @param hookFunction Function to be called
    */
-  beforeEach(hookFunction: () => Promise<void>): AsyncProperty<Ts> {
+  beforeEach(hookFunction: HookFunction): AsyncProperty<Ts> {
     this.beforeEachHook = hookFunction;
     return this;
   }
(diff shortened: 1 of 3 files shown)


dubzzz:
@sommd Thanks a lot for the contribution. It seems that prettier complains about wrong format. You just have to run `npm run format` locally to fix it.

Thanks to your PR I realized that the current typings for synchronous `beforeEach` and `afterEach` are too permissive. While without your fix asynchronous functions were very restrictive on their side (too restrictive ^^). The following code compiles fine today:

```ts
type HookSync = () => void; // real typings for fc.property
function beforeEachSync(hookFunction: HookSync) {
}

beforeEachSync(() => { });
beforeEachSync(() => 123);
beforeEachSync(async () => { });
beforeEachSync(async () => 123);
```

Here is a suggestion for the `beforeEach` of `fc.property`:

```ts
type HookFunction = () => void;
class Property<Ts> {
  beforeEach<THookFunction extends HookFunction>(hookFunction: THookFunction):
      THookFunction extends () => Promise<any>
        ? 'beforeEach expects a synchronous function'
        : Property<Ts> {
    return this as unknown as Property<Ts>; // sadly I haven't found better so far
  }
}
```

sommd:
Sorry for the delay. Fixed the formatting and I found a slightly nicer solution to make non-async `before/afterEach` more restrictive with overloads.

```ts
  beforeEach(
    invalidHookFunction: () => Promise<unknown>
  ): 'beforeEach expects a synchronous function but was given a function returning a Promise';

  beforeEach(validHookFunction: HookFunction): Property<Ts>;

  beforeEach(hookFunction: HookFunction): unknown {
    this.beforeEachHook = hookFunction;
    return this;
  }
```

dubzzz:
@sommd Thanks a lot for the PR. Do you think it requires a major bump?

In my opinion,
- part of the PR is adding the ability to use non-async functions into beforeEach of asyncProperty,
- the other is _fixing_ a typing issue that was not preventing users from using beforeEach of property with async functions.

sommd:
Technically, this is an incompatible change since it could break existing code (only type checking, there's no behaviour change). Although any existing code that would break is likely incorrect.

I see this as more of a new feature rather than a bug fix. I.e. it will now catch a likely bug in existing code. Given that, I think a minor version bump would be best.

dubzzz:
Totally agree for a minor bump

Change is too much for a bug fix, not enough to be considered as a real breaking change

I'll try to publish this new minor very soon

dubzzz:
@sommd Release of next minor 1.21.0 including your dev is on its way. It' should be there very soon (waiting Travis to publish it)

Once again thanks a lot for the contribution

PR #503, opened 2019-12-18
Include note about using expect or assert

### LINKED ISSUE #502: Allow usage of `expect`
## 🚀 Feature Request

Multiple times, I've wanted to do things like the following:

```js
// neither == nor === does the correct thing here
// currently having to use lodash.isEqual
fc.property(fc.integer(), (x) => {
  expect(...).toMatchObject({ foo: 'bar' })
})
```
```js
// currently using &&
// not the end of the world in this case
fc.property(fc.integer(), (x) => {
  const result = ...
  expect(result.foo).toBe(...)
  expect(result.bar).toBe(...)
})
```
```js
// currently explicitly doing a try..catch
// and returning true/false
fc.property(fc.integer(), (x) => {
  expect(() => ...).toThrow()
})
```

I'm thinking maybe letting the property throw errors and handling those as `false`? (I believe `expect` throws an error if it fails)

## Motivation

See above

## Example

See above

brandon-leapyear:
## Why is this PR for?

Fixes #502

## In a nutshell

❌ New feature
❌ Fix an issue
✔️ Documentation improvement
❌ Other: *please explain*

## Potential impacts

N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,10 @@ describe('properties', () => {
   });
   // string a + b + c always contains b, whatever the values of a, b and c
   it('should always contain its substrings', () => {
-    fc.assert(fc.property(fc.string(), fc.string(), fc.string(), (a,b,c) => contains(a+b+c, b)));
+    fc.assert(fc.property(fc.string(), fc.string(), fc.string(), (a,b,c) => {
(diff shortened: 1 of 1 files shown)


dubzzz:
Thanks for the PR

PR #548, opened 2020-02-05
✨ Stringify should distinguish `{}` from `Object.create(null)`

### LINKED ISSUE #507: Stringify should distinguish `{}` from `Object.create(null)`
Knowing that `{}` and `Object.create(null)` are totally different objects, we need to be able to distinguish them in stringify. This will be a must have as soon as we add the ability to generate objects without any prototype

dubzzz:
## Why is this PR for?

Fixes #507

## In a nutshell

✔️ New feature
❌ Fix an issue
❌ Documentation improvement
❌ Other: *please explain*

(✔️: yes, ❌: no)

## Potential impacts

Might impact the value produced by `fc.stringify`, helper used internally to get stringified value of any object

### DIFF
--- a/src/utils/stringify.ts
+++ b/src/utils/stringify.ts
@@ -55,19 +55,31 @@ export function stringifyInternal<Ts>(value: Ts, previousValues: any[]): string 
     case '[object Number]':
       return typeof value === 'number' ? stringifyNumber(value) : `new Number(${stringifyNumber(Number(value))})`;
     case '[object Object]': {
-      try {
-        const defaultRepr: string = (value as any).toString();
-        if (defaultRepr !== '[object Object]') return defaultRepr;
-        return (
-          '{' +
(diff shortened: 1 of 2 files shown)


src/utils/stringify.ts
```diff
@@ -55,19 +55,31 @@ export function stringifyInternal<Ts>(value: Ts, previousValues: any[]): string
...
-        return (
-          '{' +
-          Object.keys(value)
-            .map(k => `${JSON.stringify(k)}:${stringifyInternal((value as any)[k], currentValues)}`)
-            .join(',') +
-          '}'
-        );
-      } catch (err) {
-        return '[object Object]';
+      if (typeof (value as any).toString === 'function' && (value as any).toString !== Object.prototype.toString) {
```
dubzzz: Accessing `toString` might throw if backed by a custom get/set

PR #555, opened 2020-02-18
✨ Support type guards while filtering

### LINKED ISSUE #545: Support type guards while filtering
## 🚀 Feature Request

Add an overloading to `filter`... 

## Motivation

...in order to refine the result type.

## Example

With the current definition:

```ts
export interface Arbitrary<T> {
  filter(predicate: (t: T) => boolean): Arbitrary<T>
  map<U>(mapper: (t: T) => U): Arbitrary<U>
}

interface Left<E> {
  type: 'Left'
  right: E
}
interface Right<A> {
  type: 'Right'
  right: A
}
type Either<E, A> = Left<E> | Right<A>

declare function isRight<E, A>(e: Either<E, A>): e is Right<A>

declare const arb: Arbitrary<Either<string, number>>

arb.filter(isRight).map(e => /* e is still Either<string, number> */ e)
```

Proposed change:

```ts
export interface Arbitrary<T> {
  filter<U extends T>(refinement: (t: T) => t is U): Arbitrary<U>
  filter(predicate: (t: T) => boolean): Arbitrary<T>
  map<U>(mapper: (t: T) => U): Arbitrary<U>
}

arb.filter(isRight).map(e => /* e is Right<number> now */ e)
```

dubzzz:
## Why is this PR for?

Support type guards while filtering
Fixes #545

## In a nutshell

✔️ New feature
❌ Fix an issue
❌ Documentation improvement
❌ Other: *please explain*

(✔️: yes, ❌: no)

## Potential impacts

None expecting. We need to confirm it will not break any existing typings.

### DIFF
--- a/src/check/arbitrary/OptionArbitrary.ts
+++ b/src/check/arbitrary/OptionArbitrary.ts
@@ -21,7 +21,7 @@ class OptionArbitrary<T, TNil> extends Arbitrary<T | TNil> {
     function* g(): IterableIterator<Shrinkable<T | TNil>> {
       yield new Shrinkable(nil);
     }
-    return new Shrinkable(s.value_, () =>
+    return new Shrinkable(s.value_ as T | TNil, () =>
       s
         .shrink()
         .map(v => OptionArbitrary.extendedShrinkable(v, nil))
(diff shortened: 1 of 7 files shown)


dubzzz:
I still need to see what's the best solution to handle https://github.com/dubzzz/fast-check/pull/555/files#diff-e874ec8f9333a86a7d38aaab01e2fa59R79-R81

As of today, it may produce invalid typings when calling `filter` on a shrinkable. Nonetheless, it should not be a huge issue.

dubzzz:
Green run in Travis - https://travis-ci.org/dubzzz/fast-check/builds/655993341

PR #600, opened 2020-06-02
🏷️ Better typings for RunDetails

dubzzz:
## Why is this PR for?

Better typings for RunDetails. Currently typings defined for RunDetails are too permisive. Even if most of our users will never have to rely on this type, for users relying on it, it requires a deep knowledge of its content.

For instance, there are 3 different causes that can mark a run as failed: property failed, too many skips or interrupted. And depending on the situation, we may or may not have a counterexample. By defining stricter typings we will provide our users with safer types but also prevent some potential bugs in our code. 

## In a nutshell

✔️ New feature
❌ Fix an issue
❌ Documentation improvement
❌ Other: *please explain*

(✔️: yes, ❌: no)

## Potential impacts

- Typings impact


src/check/runner/reporter/RunDetails.ts
```diff
@@ -6,7 +6,60 @@ import { ExecutionTree } from './ExecutionTree';
  *
  * A failing property can easily detected by checking the `failed` flag of this structure
  */
-export interface RunDetails<Ts> {
+export type RunDetails<Ts> =
+  | PropertyFailureRunDetails<Ts>
```
dubzzz, line 10: Be more explicit about the fact it corresponds to a failure in the name of the type

src/check/runner/reporter/RunExecution.ts
```diff
@@ -96,14 +96,14 @@ export class RunExecution<Ts> {
       // encountered a property failure
       return {
         failed: true,
-        interrupted: this.interrupted,
+        interrupted: false,
```
dubzzz, line 99: Those changes have to be double-checked. Ideally types should prevent us from such bugs 🤔

src/check/runner/reporter/RunDetails.ts
```diff
@@ -6,7 +6,60 @@ import { ExecutionTree } from './ExecutionTree';
...
+  | TooManySkipsRunDetails<Ts>
+  | InterruptedRunDetails<Ts>
+  | SuccessfulRunDetails<Ts>;
+
+/**
+ * Run reported as failed because
+ * the property failed
+ */
+export type PropertyFailureRunDetails<Ts> = RunDetailsWithDoc<Ts> & {
```
dubzzz, line 19: Re-export them at root

PR #609, opened 2020-06-08
💚 Generated typings where causing the doc generation to crash

dubzzz:
## Why is this PR for?

The errors were reported by @typescript-eslint/explicit-module-boundary-types as warnings. Unfortunately enabling them would ask to add several missing typings that may not be really needed.

The rule could be added only for the real source code by adding the following into `.eslintrc.cjs` file:
```
  overrides: [
    {
      files: ['src/**/*.ts'],
      rules: {
        '@typescript-eslint/explicit-module-boundary-types': ['error', { allowArgumentsExplicitlyTypedAsAny: true }],
      },
    },
  ],
```

## In a nutshell

❌ New feature
❌ Fix an issue
❌ Documentation improvement
✔️ Other: *fix build issue*

(✔️: yes, ❌: no)

## Potential impacts

None.

### DIFF
--- a/src/check/arbitrary/EmailArbitrary.ts
+++ b/src/check/arbitrary/EmailArbitrary.ts
@@ -3,13 +3,14 @@ import { buildLowerAlphaNumericArb } from './helpers/SpecificCharacterRange';
 import { domain } from './HostArbitrary';
 import { stringOf } from './StringArbitrary';
 import { tuple } from './TupleArbitrary';
+import { Arbitrary } from './definition/Arbitrary';
 
 /**
  * For email address
  *
  * According to RFC 5322 - https://www.ietf.org/rfc/rfc5322.txt
  */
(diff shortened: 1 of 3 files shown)


dubzzz:
For an unknown reason the generated `*.d.ts` file for `EmailArbitrary` is:
```ts
export declare function emailAddress(): import("./definition/Arbitrary").Arbitrary<string>;
```

While the one for uuid is:
```ts
import { Arbitrary } from './definition/Arbitrary';
export declare function uuid(): Arbitrary<string>;
export declare function uuidV(versionNumber: 1 | 2 | 3 | 4 | 5): Arbitrary<string>;
```

While both of them are not explicitely typing the exports 🤔
But uuid file is importing `import { Arbitrary } from './definition/Arbitrary';` for other functions declared in this file.

Issue #876, opened 2020-08-13
Global beforeEach/afterEach


brapifra:
## 🚀 Feature Request

Similarly to test runners like `jest`, it would be great if there was a way of setting global beforeEach/afterEach functions that get executed before/after a property is run.

## Motivation

When using libraries like react-testing-library, it is recommended to call its `cleanup` method between tests to prevent issues like memory leaks. While this is done automatically for popular test runners, I need to manually do it for fast-check tests.
https://testing-library.com/docs/react-testing-library/api#cleanup

## Example
Currently, I have to do this:
```typescript
fc.property(...).beforeEach(cleanup);
```
Ideally, I would want to do something like this:

```typescript
// In jest.setup.ts file
fc.configureGlobal({
    beforeEach: cleanup,
});
// Then I could skip the `beforeEach` from properties
fc.property(...);
```

dubzzz:
Indeed it would be a pretty cool and interesting feature to bring into fast-check. Actually even if for the moment `configureGlobal` is only used to store configuration settings that would have been passed towards `assert` or `check` I have already plans to extend it to more options like the one you suggested here (see https://github.com/dubzzz/fast-check/issues/452).

To go back on the suggested feature: it is clearly something that would be great to have built-in.

A suggestion would be to add it into the typings of [`GlobalParameters`](https://github.com/dubzzz/fast-check/blob/101eb97228b5e08bb0c0d5855e0fccbf0a6d4ccf/src/check/runner/configuration/GlobalParameters.ts#L6) with something like:

```ts
type GlobalParameters = Pick<Parameters...> & { otherOptionsNotUsedByAssert };
```

As for [`reporter`](https://github.com/dubzzz/fast-check/blob/101eb97228b5e08bb0c0d5855e0fccbf0a6d4ccf/src/check/runner/configuration/Parameters.ts#L116-L136) setting we would have to deal with two possible versions: one synchronous and the other one asynchronous. So it might require 4 parameters (2 for synchronous, 2 for asynchronous). [As for `reporter`](https://github.com/dubzzz/fast-check/blob/101eb97228b5e08bb0c0d5855e0fccbf0a6d4ccf/src/check/runner/configuration/Parameters.ts#L133-L134) I would say that if one of the asynchronous method is set asynchronous properties will be forbidden.

...

Anyway, let's go for such feature! Would you be interested into giving it a try?

I can clearly send you pointers for that if needed.

brapifra:
[...]

PR #900, opened 2020-08-18
Add global beforeEach/afterEach hooks

brapifra:
## In a nutshell
This PR adds 4 new global parameters: `beforeEach`, `afterEach`, `asyncBeforeEach`and`asyncAfterEach` that will be executed during the execution of any property.
It solves the issue https://github.com/dubzzz/fast-check/issues/876

✔️ New feature
❌ Fix an issue
❌ Documentation improvement
❌ Other: *please explain*

(✔️: yes, ❌: no)

## Potential impacts

None


test/unit/check/property/AsyncProperty.spec.ts
```diff
@@ -3,11 +3,14 @@ import { Shrinkable } from '../../../../src/check/arbitrary/definition/Shrinkabl
 import { asyncProperty } from '../../../../src/check/property/AsyncProperty';
 import { pre } from '../../../../src/check/precondition/Pre';
 import { PreconditionFailure } from '../../../../src/check/precondition/PreconditionFailure';
+import { configureGlobal, resetConfigureGlobal } from '../../../../src/check/runner/configuration/GlobalParameters';
 
 import * as stubArb from '../../stubs/arbitraries';
 import * as stubRng from '../../stubs/generators';
 
 describe('AsyncProperty', () => {
+  beforeEach(() => resetConfigureGlobal());
```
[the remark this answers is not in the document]
dubzzz: 🤔 Actually we may even mock `readConfigureGlobal` and don't use any of `configureGlobal` or `resetConfigureGlobal`. But you can keep it that way for this PR

test/unit/check/property/AsyncProperty.spec.ts
```diff
@@ -122,6 +125,33 @@ describe('AsyncProperty', () => {
...
+    configureGlobal({
+      asyncBeforeEach: () => (prob.beforeEachCalled = true),
+    });
+    const p = asyncProperty(stubArb.single(8), async (_arg: number) => {
+      const beforeEachCalled = prob.beforeEachCalled;
+      prob.beforeEachCalled = false;
+      return beforeEachCalled;
+    });
+    expect(await p.run(p.generate(stubRng.mutable.nocall()).value)).toBe(null);
+  });
```
dubzzz, line 139: Given (see other comments):

1. `beforeEach` and `asyncBeforeEach` should not be defined together in global settings (otherwise it will make the property fail to build)
2. For async properties the selected hook coming from global will be chosen as follow: take `asyncBeforeEach`, if not set take `beforeEach`
3. Users should be able to override them at anytime directly at call-site

My test suite would be different:
- Should take asyncBeforeEach coming from globals as the default for beforeEach if specified
- Should take beforeEach coming from globals as the default for beforeEach if specified
- Should throw if both beforeEach and asyncBeforeEach are defined in globals
- Should be able to override `beforeEach` globals locally

test/unit/check/property/Property.spec.ts
```diff
@@ -3,11 +3,14 @@ import { Shrinkable } from '../../../../src/check/arbitrary/definition/Shrinkabl
 import { property } from '../../../../src/check/property/Property';
 import { pre } from '../../../../src/check/precondition/Pre';
 import { PreconditionFailure } from '../../../../src/check/precondition/PreconditionFailure';
+import { configureGlobal, resetConfigureGlobal } from '../../../../src/check/runner/configuration/GlobalParameters';
 
 import * as stubArb from '../../stubs/arbitraries';
 import * as stubRng from '../../stubs/generators';
 
 describe('Property', () => {
+  beforeEach(() => resetConfigureGlobal());
```
dubzzz, line 12: Same use `afterEach` instead

test/unit/check/property/Property.spec.ts
```diff
@@ -121,6 +124,31 @@ describe('Property', () => {
     }).beforeEach(() => (prob.beforeEachCalled = true));
     expect(p.run(p.generate(stubRng.mutable.nocall()).value)).toBe(null);
   });
+  it('Should always execute the global beforeEach hook before the test', () => {
```
dubzzz, line 127: Given (see other comments):

1. `beforeEach` and `asyncBeforeEach` should not be defined together in global settings (otherwise it will make the property fail to build)
2. For async properties the selected hook coming from global will be chosen as follow: take `asyncBeforeEach`, if not set take `beforeEach`
3. Users should be able to override them at anytime directly at call-site

My test suite would be different:
- Should take beforeEach coming from globals as the default for beforeEach if specified
- Should throw if asyncBeforeEach is defined in globals
- Should throw if both beforeEach and asyncBeforeEach are defined in globals
- Should be able to override `beforeEach` globals locally

dubzzz (changes requested):
Great job @brapifra!

I put some comments in order to ensure we preserve consistency across the primitives offered by the API.

test/unit/check/runner/configuration/GlobalParameters.spec.ts
```diff
@@ -10,7 +10,13 @@ describe('GlobalParameters', () => {
...
 
   it('should be able to set, read and clear parameters globally', () => {
-    const myGlobalConfiguration = { numRuns: 123 };
+    const myGlobalConfiguration = {
+      numRuns: 123,
+      beforeEach: jest.fn(),
+      afterEach: jest.fn(),
+      asyncBeforeEach: jest.fn(),
+      asyncAfterEach: jest.fn(),
+    };
```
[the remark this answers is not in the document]
brapifra: Ooops, yeah. I was using while developing and I forgot to remove it. I'll revert the changes.

test/unit/check/property/AsyncProperty.spec.ts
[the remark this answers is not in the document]
brapifra: Yeah, we could mock it. Although I usually prefer to write integration-like tests. But it's really a personal preference so I don't mind changing it :smile: 
Regarding the `beforeEach`/`afterEach` comment. I'm pretty sure jest runs each test module/file in isolation. Therefore, I think it doesn't really matter to use one or the other in this case...

src/check/property/AsyncProperty.generic.ts
```diff
@@ -40,9 +41,18 @@ export interface IAsyncPropertyWithHooks<Ts> extends IAsyncProperty<Ts> {
...
+  private readonly globalAfterEachHook: HookFunction;
+  private beforeEachHook: HookFunction;
+  private afterEachHook: HookFunction;
+  constructor(readonly arb: Arbitrary<Ts>, readonly predicate: (t: Ts) => Promise<boolean | void>) {
+    const { asyncBeforeEach = AsyncProperty.dummyHook, asyncAfterEach = AsyncProperty.dummyHook } =
+      readConfigureGlobal() || {};
+    this.globalBeforeEachHook = asyncBeforeEach;
+    this.globalAfterEachHook = asyncAfterEach;
+    this.beforeEachHook = asyncBeforeEach;
+    this.afterEachHook = asyncAfterEach;
```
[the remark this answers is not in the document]
brapifra: Yeah, I initially implemented it like that. However, I soon realised that that kind of solution wouldn't fit my needs from the issue #876 
e.g. Let's say I have project with tons of property-based tests. Some of them are using beforeEach and/or afterEach. Now I want to add a call to a `cleanup` method globally via one of this new global parameters (e.g. afterEach). With your implementation, I would have to manually add that `cleanup` method to all these properties that were already implementing one/both hook(s).
Quite annoying IMO :sweat_smile:

src/check/runner/configuration/GlobalParameters.ts
```diff
@@ -4,8 +4,19 @@ import { Parameters } from './Parameters';
...
 
+type HookFunction = () => void;
+type AsyncHookFunction = (() => Promise<unknown>) | (() => void);
+
+type ExtraGlobalParameters = {
+  beforeEach?: HookFunction;
+  afterEach?: HookFunction;
+  asyncBeforeEach?: AsyncHookFunction;
+  asyncAfterEach?: AsyncHookFunction;
+};
```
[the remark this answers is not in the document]
brapifra: Not sure I understand 1 :thinking: 
What's the problem with being defined together?

src/check/property/AsyncProperty.generic.ts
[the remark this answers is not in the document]
dubzzz: Actually not really, my point is that if you define a value for `beforeEach` (the synchronous one) even asynchronous properties will take them into account.

src/check/runner/configuration/GlobalParameters.ts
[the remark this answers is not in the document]
dubzzz: In reality, the existence of both `property` (synchronous) and `asyncProperty` (asynchronous) is there for legacy reasons. Ideally all properties should be defined as asynchronous.

For instance, if you take a look into [this wrapper around fast-check designed for ava](https://github.com/dubzzz/ava-fast-check/), there is no more distinction between synchronous and asynchronous properties. Behind the scene everything is asynchronous as it only uses asyncProperty (but users don't know).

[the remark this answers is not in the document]
dubzzz: If you define the two global hooks together, there is probably an issue in your test suite. Or maybe hooks should not have been defined at a global level but at file or directory level.

My point is:
- If no global: property/asyncProperty take what the user specify using the beforeEach/afterEach methods
- If only global beforeEach: property/asyncProperty take it but replace it if the user specify one using the beforeEach method
- If only global asyncBeforeEach: property throws / asyncProperty takes it but replaces it if the user specify one using the beforeEach method
- If global beforeEach and asyncBeforeEach: property/asyncProperty throw

src/check/property/AsyncProperty.generic.ts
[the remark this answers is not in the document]
brapifra: Oh I see what you mean now. So in an example like this:
```typescript
configureGlobal({
    beforeEach: () => console.log('global beforeEach'),
    asyncAfterEach: () => console.log('global afterEach')
});

fc.asyncProperty(() => {}).afterEach(() => console.log('local afterEach'));
```
It should print:
```
global beforeEach
global AfterEach
local afterEach
```
Right?

src/check/runner/configuration/GlobalParameters.ts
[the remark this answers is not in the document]
brapifra: As mentioned in https://github.com/dubzzz/fast-check/pull/900#discussion_r473277266, I think that it is valuable to not replace the global hooks when new ones have been defined with the beforeEach/afterEach methods.
If I have this case:
```typescript
configureGlobal({
    beforeEach: () => console.log('cleanup'),
});

fc.property(() => {}).beforeEach(() => console.log('beforeEach'));
```
I want both hooks to be run. Otherwise I'm going to have to call `cleanup` every time I use the `beforeEach` method in a property/asyncProperty.
Although I understand your point, an API like the one you suggest wouldn't solve my problem unfortunately :sweat_smile:

src/check/property/AsyncProperty.generic.ts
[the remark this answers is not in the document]
dubzzz: My idea would have been the following:

```js
configureGlobal({
    beforeEach: () => console.log('global beforeEach'),
    asyncAfterEach: () => console.log('global afterEach')
});

fc.asyncProperty(() => {}).afterEach(() => console.log('local afterEach'));
```

Would print:
```
global beforeEach
local afterEach
```
_as global afterEach would have been overriden at call site level_

**AND I need to think about your idea:**

>  it is valuable to not replace the global hooks when new ones have been defined with the beforeEach/afterEach methods

Initially global overrides have been created as a syntaxic sugar to help users define only once a behaviour while it allows to override it locally. If we go for your proposal it means that in the case of beforeEach hooks, override will no be possible. I need to think abour it and come back to you (I will have a look to what Jest or Mocha do on this subject).

Meanwhile if we go in that direction I think that the ordering should be something like : 
- global before
- local before
- local after
- global after

Here are other examples:

<details>
<summary>Global beforeEach, Global afterEach</summary>

```js
configureGlobal({
    beforeEach: () => console.log('global beforeEach'),
    afterEach: () => console.log('global afterEach')
});

fc.property(() => {});
// global beforeEach
// global afterEach

fc.asyncProperty(() => {});
// global beforeEach
// global afterEach

fc.property(() => {}).afterEach(() => console.log('local afterEach'));
// global beforeEach
// local afterEach

fc.asyncProperty(() => {}).afterEach(() => console.log('local afterEach'));
// global beforeEach
// local afterEach
```
</details>

<details>
<summary>Global beforeEach, Global asyncAfterEach</summary>

```js
configureGlobal({
    beforeEach: () => console.log('global beforeEach'),
    asyncAfterEach: () => console.log('global afterEach')
});

fc.property(() => {});
// OPT1 throws: Async hooks are not supported for synchronous properties. You may override it locally to prevent this error message
// OPT2 throws: Async hooks are not supported for synchronous properties. Please remove your global override or convert it to a synchronous hook or switch to an asynchronous property

fc.asyncProperty(() => {});
// global beforeEach
// global afterEach

fc.property(() => {}).afterEach(() => console.log('local afterEach'));
// OPT1
// global beforeEach
// local afterEach
// OPT2 throws: Async hooks are not supported for synchronous properties. Please remove your global override or convert it to a synchronous hook or switch to an asynchronous property

fc.asyncProperty(() => {}).afterEach(() => console.log('local afterEach'));
// global beforeEach
// local afterEach
```
</details>

<details>
<summary>Global afterEach, Global asyncAfterEach</summary>

```js
configureGlobal({
    afterEach: () => console.log('global afterEach'),
    asyncAfterEach: () => console.log('global afterEach')
});

fc.property(() => {});
// throws: Cannot define both sync and async version of the same hook

fc.asyncProperty(() => {});
// throws: Cannot define both sync and async version of the same hook
```
</details>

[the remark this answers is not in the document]
dubzzz: Concerning your point, what do you think of something like?

```js
fc.asyncProperty(() => {})
  .afterEach((previousAfterEach) => {
    previousAfterEach();
    console.log('local afterEach');
  });
```

or

```js
fc.asyncProperty(() => {})
  .afterEach(() => console.log('local afterEach'), { behaviour: 'keep-existing' });
// behaviour being one of 'keep-existing', 'override' (the default)...
```

or

```js
fc.asyncProperty(() => {})
  .afterEachKeepExisting(() => console.log('local afterEach'));
```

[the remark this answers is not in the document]
brapifra: Nice! Thanks for thinking about it :slightly_smiling_face: 
I personally find the API of option 1 a bit nicer. However, I think option 2 would give the flexibility of being able to modify the default behaviour via global config:
```typescript
fc.configureGlobal({ defaultBehaviourOfPropertyHooks: 'keep-existing' });
```
But maybe it's over-complicating the API, not sure... I'll leave it to you to decide! :smile:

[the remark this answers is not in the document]
dubzzz: Let's go for option 1! Actually it was my favourite when I wrote the comment and you seem to like it too:

```js
fc.asyncProperty(() => {})
  .afterEach((previousAfterEach) => {
    previousAfterEach();
    console.log('local afterEach');
  });
```

brapifra:
I've made most of the requested changes. Let me know what you think @dubzzz !
The only thing that I didn't implement was the error throwing when a synchronous property is run and there is some async hook set up globally. Maybe I'm missing something but... why would we want to do that?
IMO synchronous property should simply ignore those async hooks instead of throwing an error. e.g.
```typescript
configureGlobal({
     asyncBeforeEach: () => {
          console.log("I just want to run this with my async properties and I don't want to break the sync ones");
     }
})
```

src/check/property/AsyncProperty.generic.ts
```diff
@@ -44,16 +47,28 @@ export interface IAsyncPropertyWithHooks<Ts> extends IAsyncProperty<Ts> {
...
+  constructor(readonly arb: Arbitrary<Ts>, readonly predicate: (t: Ts) => Promise<boolean | void>) {
+    const {
+      asyncBeforeEach,
+      asyncAfterEach,
+      beforeEach = AsyncProperty.dummyGlobalHook,
+      afterEach = AsyncProperty.dummyGlobalHook,
+    } = readConfigureGlobal() || {};
+    this.globalBeforeEachHook = asyncBeforeEach || beforeEach;
+    this.globalAfterEachHook = asyncAfterEach || afterEach;
+  }
```
dubzzz, line 65: As users are now powerful enough to tell the framework how to merge multiple hooks together, I would make no differences between global and local hooks. And then let the user does what she/he wants with consecutive declarations of beforeEach. In other words, the global hook is just used to initialize beforeEachHook but it does not have any specific priviledges.

I believe a code like that:
```js
configureGlobal({ asyncBeforeEach: () => console.log('1') })

fc.asyncProperty(...)
  .beforeEach(prev => {
    prev();
    console.log('2');
  })
  .beforeEach(prev => {
    prev();
    console.log('3');
  })
  .beforeEach(prev => {
    console.log('4');
    prev();
    console.log('5');
  })
```

Should resulting into a hook that would do:
```js
() => {
    console.log('4');
    console.log('1');
    console.log('2');
    console.log('3');
    console.log('5');
}
```

In other words, I'd update the initialization part as follow:

```suggestion
  static dummyHook: AsyncPropertyHookFunction = (globalHook) => globalHook();
  private beforeEachHook: AsyncPropertyHookFunction;
  private afterEachHook: AsyncPropertyHookFunction;
  constructor(readonly arb: Arbitrary<Ts>, readonly predicate: (t: Ts) => Promise<boolean | void>) {
    const { asyncBeforeEach,  asyncAfterEach,  beforeEach,  afterEach  } = readConfigureGlobal() || {};
    if (asyncBeforeEach !== undefined && beforeEach !== undefined) {
      // Concerning those checks they need to be introduce as soon as we release the version.
      // As a library authors, it's easier to relax constraints that add new ones (breaking change)
      throw new Error('Illegal');
    }
    if (asyncAfterEach !== undefined && afterEach !== undefined) {
      throw new Error('Illegal');
    }
    this.beforeEachHook = asyncBeforeEach || beforeEach || AsyncProperty.defaultHook;
    this.afterEachHook = asyncAfterEach || afterEach || AsyncProperty.defaultHook;
  }
```

The beforeEach method would then have to be updated as follow:
```js
  beforeEach(hookFunction: AsyncPropertyHookFunction): AsyncProperty<Ts> {
    const previousBeforeEachHook = this.beforeEachHook;
    this.beforeEachHook = () => hookFunction(previousBeforeEachHook);
    return this;
  }
```

src/check/property/Property.generic.ts
```diff
@@ -3,14 +3,15 @@ import { Arbitrary } from '../arbitrary/definition/Arbitrary';
...
+import { readConfigureGlobal, GlobalPropertyHookFunction } from '../runner/configuration/GlobalParameters';
 
 /**
  * Type of legal hook function that can be used to call `beforeEach` or `afterEach`
  * on a {@link IPropertyWithHooks}
  *
  * @public
  */
-export type PropertyHookFunction = () => void;
+export type PropertyHookFunction = (globalHookFunction: GlobalPropertyHookFunction) => void;
```
dubzzz, line 14:
```suggestion
export type PropertyHookFunction = (previousHookFunction: PropertyHookFunction) => void;
```

src/check/property/Property.generic.ts
```diff
@@ -59,16 +60,23 @@ export interface IPropertyWithHooks<Ts> extends IProperty<Ts> {
...
+  static dummyGlobalHook: GlobalPropertyHookFunction = () => {};
+  static defaultHook: PropertyHookFunction = (globalHook) => globalHook();
+  private readonly globalBeforeEachHook: GlobalPropertyHookFunction;
+  private readonly globalAfterEachHook: GlobalPropertyHookFunction;
+  private beforeEachHook: PropertyHookFunction = Property.defaultHook;
+  private afterEachHook: PropertyHookFunction = Property.defaultHook;
+  constructor(readonly arb: Arbitrary<Ts>, readonly predicate: (t: Ts) => boolean | void) {
+    const { beforeEach = Property.dummyGlobalHook, afterEach = Property.dummyGlobalHook } = readConfigureGlobal() || {};
+    this.globalBeforeEachHook = beforeEach;
+    this.globalAfterEachHook = afterEach;
```
dubzzz, line 72: As for the async version:
```suggestion
    const { asyncBeforeEach,  asyncAfterEach,  beforeEach,  afterEach  } = readConfigureGlobal() || {};
    if (asyncBeforeEach !== undefined) {
      // Why this check?
      // - Users might have been told that a clean-up hook was globally set. So write her/his property
      //   thinking that the hook would be there. But the hook was actually defined on async versions only so it passes
      // - Consistency with the rest of the philosophy of the lib and where it goes concerning sync vs async
      // - As a library authors, it's easier to relax constraints that add new ones (breaking change). So putting it now
      // would at least require a change given there are good reasons not to do that. Not putting it would be impossible
      // to change
      throw new Error('Illegal');
    }
    if (asyncAfterEach !== undefined) {
      throw new Error('Illegal');
    }
    this.beforeEachHook = beforeEach || Property.defaultHook;
    this.afterEachHook = afterEach || Property.defaultHook;
```

src/check/property/AsyncProperty.generic.ts
```diff
@@ -3,14 +3,17 @@ import { Arbitrary } from '../arbitrary/definition/Arbitrary';
...
 /**
  * Type of legal hook function that can be used to call `beforeEach` or `afterEach`
  * on a {@link IAsyncPropertyWithHooks}
  *
  * @public
  */
-export type AsyncPropertyHookFunction = (() => Promise<unknown>) | (() => void);
+export type AsyncPropertyHookFunction =
+  | ((globalHookFunction: GlobalAsyncPropertyHookFunction) => Promise<unknown>)
+  | ((globalHookFunction: GlobalAsyncPropertyHookFunction) => void);
```
dubzzz, line 16:
```suggestion
export type AsyncPropertyHookFunction =
  | ((previousHookFunction: AsyncPropertyHookFunction) => Promise<unknown>)
  | ((previousHookFunction: AsyncPropertyHookFunction) => void);
```

src/check/runner/configuration/GlobalParameters.ts
```diff
@@ -4,11 +4,21 @@ import { Parameters } from './Parameters';
...
+  asyncAfterEach?: GlobalAsyncPropertyHookFunction;
+};
+
 /**
  * Type describing the global overrides
  * @public
  */
-export type GlobalParameters = Pick<Parameters<unknown>, Exclude<keyof Parameters<unknown>, 'path' | 'examples'>>;
+export type GlobalParameters = Pick<Parameters<unknown>, Exclude<keyof Parameters<unknown>, 'path' | 'examples'>> &
+  ExtraGlobalParameters;
```
dubzzz, line 21: Directly merge Extra* in the GlobalPaarmeters as follow:

```suggestion
export type GlobalParameters = Pick<Parameters<unknown>, Exclude<keyof Parameters<unknown>, 'path' | 'examples'>> &
  {
  beforeEach?: GlobalPropertyHookFunction;
  afterEach?: GlobalPropertyHookFunction;
  asyncBeforeEach?: GlobalAsyncPropertyHookFunction;
  asyncAfterEach?: GlobalAsyncPropertyHookFunction;
};
```

It would be better in terms of generated doc.

src/check/runner/configuration/GlobalParameters.ts
```diff
@@ -4,11 +4,21 @@ import { Parameters } from './Parameters';
 /** @internal */
 const globalParametersSymbol = Symbol.for('fast-check/GlobalParameters');
 
+export type GlobalPropertyHookFunction = () => void;
+export type GlobalAsyncPropertyHookFunction = (() => Promise<unknown>) | (() => void);
```
dubzzz, line 8: As any exported type that are used by types or functions exposed to the end users they will need to be marked as public, documented and exported in fast-check-default.ts file.

dubzzz (commented):
Once again thanks a lot for this huge contribution. I added some comments and suggestions to update the code and offer more flexibility and sanity checks to the end-users

src/check/property/AsyncProperty.generic.ts
brapifra: `previousHookFunction` can't be an `AsyncPropertyHookFunction`. This function doesn't accept any input arguments. That's why I used `GlobalAsyncPropertyHookFunction` instead.

src/check/property/Property.generic.ts
brapifra: Fair enough! I'll change it

src/check/property/Property.generic.ts
brapifra: Same here (https://github.com/dubzzz/fast-check/pull/900#discussion_r478601102)

src/check/property/AsyncProperty.generic.ts
brapifra: Oh, okay. I didn't know you wanted to go this route. Sounds good to me! :slightly_smiling_face:

src/check/property/AsyncProperty.generic.ts
```diff
@@ -44,10 +47,27 @@ export interface IAsyncPropertyWithHooks<Ts> extends IAsyncProperty<Ts> {
...
-  constructor(readonly arb: Arbitrary<Ts>, readonly predicate: (t: Ts) => Promise<boolean | void>) {}
+  static dummyHook: GlobalAsyncPropertyHookFunction = () => {};
+  private beforeEachHook: GlobalAsyncPropertyHookFunction;
+  private afterEachHook: GlobalAsyncPropertyHookFunction;
+  constructor(readonly arb: Arbitrary<Ts>, readonly predicate: (t: Ts) => Promise<boolean | void>) {
+    const { asyncBeforeEach, asyncAfterEach, beforeEach, afterEach } = readConfigureGlobal() || {};
+
+    if (asyncBeforeEach !== undefined && beforeEach !== undefined) {
+      throw Error(
+        'Global "asyncBeforeEach" and "beforeEach" parameters can\'t be set at the same time when running async properties'
```
brapifra, line 58: Happy to change these error messages if needed :slightly_smiling_face:

test/unit/check/property/AsyncProperty.spec.ts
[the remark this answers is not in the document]
dubzzz: Just to make sure, prefer a `afterEach` here

src/check/runner/configuration/GlobalParameters.ts
```diff
@@ -5,11 +5,28 @@ import { Parameters } from './Parameters';
...
 
+/**
+ * Type describing the global overrides
+ * @public
+ */
+export type GlobalParameters = Pick<Parameters<unknown>, Exclude<keyof Parameters<unknown>, 'path' | 'examples'>> & {
+  beforeEach?: GlobalPropertyHookFunction;
+  afterEach?: GlobalPropertyHookFunction;
+  asyncBeforeEach?: GlobalAsyncPropertyHookFunction;
+  asyncAfterEach?: GlobalAsyncPropertyHookFunction;
```
dubzzz, line 28: Add a quick documentation to explain what are `beforeEach` and others?

Suggestion:
```suggestion
  /**
  Specify a function that will be called before each execution of a property.
  It behaves as-if you manually called `beforeEach` method on all the properties you execute with fast-check.
  
  The function will be used for both {@link fast-check#property} and {@link fast-check#asyncProperty}.
  This global override should never be used in conjunction with `asyncBeforeEach`.
  */
  beforeEach?: GlobalPropertyHookFunction;
  /**
  Specify a function that will be called after each execution of a property.
  It behaves as-if you manually called `afterEach` method on all the properties you execute with fast-check.
  
  The function will be used for both {@link fast-check#property} and {@link fast-check#asyncProperty}.
  This global override should never be used in conjunction with `asyncAfterEach`.
  */
  afterEach?: GlobalPropertyHookFunction;
  /**
  Specify a function that will be called before each execution of an asynchronous property.
  It behaves as-if you manually called `beforeEach` method on all the asynchronous properties you execute with fast-check.
  
  The function will be used only for {@link fast-check#asyncProperty}. It makes synchronous properties created by {@link fast-check#property} unable to run.
  This global override should never be used in conjunction with `beforeEach`.
  */
  asyncBeforeEach?: GlobalAsyncPropertyHookFunction;
  /**
  Specify a function that will be called after each execution of an asynchronous property.
  It behaves as-if you manually called `afterEach` method on all the asynchronous properties you execute with fast-check.
  
  The function will be used only for {@link fast-check#asyncProperty}. It makes synchronous properties created by {@link fast-check#property} unable to run.
  This global override should never be used in conjunction with `afterEach`.
  */
  asyncAfterEach?: GlobalAsyncPropertyHookFunction;
```

test/unit/check/property/Property.spec.ts
dubzzz:
```suggestion
  afterEach(() => resetConfigureGlobal());
```

I think it would be safer to always clean properly after the run. Cleaning before do not ensure we clean stuff for next tests.

dubzzz (commented):
@brapifra Thank you so much for your work and time. Just a few remarks before we merge it. Really really thank for the time you already spent for this PR 👍

dubzzz:
Perfect 👍 

Great job! Thanks a lot for the Pull Request!
I'll try to release a new minor including this new feature very soon.

PR #938, opened 2020-08-25
Fix NaN passed to max in fc.date()

brandon-leapyear:
Quick fix where using `fc.date({ max: new Date('asdf') })` would hang.

Side question: why does it hang? Should `fc.integer()` check if the bounds are `NaN` and error accordingly?

## In a nutshell

❌ New feature
✔️ Fix an issue
❌ Documentation improvement
❌ Other: *please explain*

(✔️: yes, ❌: no)

## Potential impacts

If you expect some noticeable impacts with your change please explain.

Here are some examples of impacts that might be reported into this section:
- Generated values impact,
- Shrink values impact,
- Performance impact,
- Typings impact

### DIFF
--- a/src/check/arbitrary/DateArbitrary.ts
+++ b/src/check/arbitrary/DateArbitrary.ts
@@ -13,7 +13,7 @@ export function date(constraints?: { min?: Date; max?: Date }): Arbitrary<Date> 
   const intMin = constraints && constraints.min ? constraints.min.getTime() : -8640000000000000;
   const intMax = constraints && constraints.max ? constraints.max.getTime() : 8640000000000000;
   if (Number.isNaN(intMin)) throw new Error('fc.date min must be valid instance of Date');
-  if (Number.isNaN(intMin)) throw new Error('fc.date max must be valid instance of Date');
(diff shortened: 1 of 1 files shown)


brandon-leapyear:
Ah if `max` is `NaN`, it looks like `MaxAcceptedRandom` is `NaN`, which means this condition will always be false, which means the while loop will never exit

https://github.com/dubzzz/pure-rand/blob/22da32428fd5dc65dfa194bc882e4930d9a3c66c/src/distribution/UniformIntDistribution.ts#L41-L44

dubzzz (approved):
Good catch!

dubzzz:
> Should fc.integer() check if the bounds are NaN and error accordingly?

> this condition will always be false, which means the while loop will never exit

@brandon-leapyear 

There is maybe something we could fix there: either in pure-rand or on fast-check's side 🤔

For the moment, I don't have a strong opinion.

My first feeling is: I'd more go for the check on fast-check's side as it seems to be more a bad usage of pure-rand than a bug of pure-rand.

dubzzz:
@brandon-leapyear Version 2.2.1 includes your fix 🥳 Thanks a lot 👍

PR #977, opened 2020-09-11
🏷️ Mark fc.pre as an asserts function in typings

### LINKED ISSUE #558: Mark fc.pre as assert function in TS
TypeScript 3.7 · TypeScript https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions

In fast-check `fc.pre` can be used to stop the execution of a property if some preconditions are not validated. 

```typescript
// v is number | null
fc.pre(v !== null)
// while v cannot be null there, TypeScript stills consider that it can be
```

As adding this feature might not be backward compatible, releasing it might be delayed into later majors of fast-check.

dubzzz:
Fixes #558

## In a nutshell

✔️ New feature
❌ Fix an issue
❌ Documentation improvement
❌ Other: *please explain*

(✔️: yes, ❌: no)

## Potential impacts

Typings impact: fc.pre is now properly flagged as asserts

### DIFF
--- a/src/check/precondition/Pre.ts
+++ b/src/check/precondition/Pre.ts
@@ -5,7 +5,7 @@ import { PreconditionFailure } from './PreconditionFailure';
  * @param expectTruthy - cancel the run whenever this value is falsy
  * @public
  */
-export function pre(expectTruthy: boolean): asserts expectTruthy {
+export function pre(expectTruthy: boolean): asserts expectTruthy is true {
   if (!expectTruthy) {
     throw new PreconditionFailure();
   }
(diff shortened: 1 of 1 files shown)


dubzzz:
In order to use it, the assertion has to be defined on a function and not a method:
```ts
// compiles!
function(a :number, b :number) {
  const pre: typeof fc.pre = fc.pre;
  pre(a!==0 && b!==0);
}

// fails!
// Assertions require every name in the call target to be declared with an explicit type annotation.ts(2775)
function(a :number, b :number) {
  fc.pre(a!==0 && b!==0);
}

// fails!
// Assertions require every name in the call target to be declared with an explicit type annotation.ts(2775)
// 'pre' needs an explicit type annotation
function(a :number, b :number) {
  const pre = fc.pre;
  fc.pre(a!==0 && b!==0);
}
```

See https://github.com/microsoft/TypeScript/issues/36931

PR #986, opened 2020-09-17
✨ Move to explicit `minLength`/`maxLength` constraints on `array`

dubzzz:
Before this PR, defining constraints on the size of an array required to use one of the following signatures:
- `fc.array(arb, maxLength)`
- `fc.array(arb, minLength, maxLength)`

This PR depreciates those signatures for another one `fc.array(arb, {minLength, maxLength})` which is more explicit.

The reasons behind this change are:
1. when seeing a call like `fc.array(arb, 10)` it was difficult to understand the meaning of the second argument: is is for the max? for the min?
2. no signature to only specify a min length, specifying a min required the user to also specify a max
3. difficult to use this kind of signature with `fc.set`, `fc.uint32array`... or even worst `fc.object`

The PR comes with a codemod to convert automatically implicit `minLength` and `maxLength` to the newly added object expression syntax.

Related to #89

## In a nutshell

✔️ New feature
❌ Fix an issue
❌ Documentation improvement
❌ Other: *please explain*

(✔️: yes, ❌: no)

## Potential impacts

Generated values impact:
- `fc.func`
- `fc.lorem`


dubzzz:
Related to #992

Issue #96, opened 2018-05-18
TypedArray Arbitraries?


bitjson:
Fantastic work on this library! I’ve been hoping to find a good excuse to try it on a real project for a while now.

What’s the cleanest/fastest way to generate varying lengths of random TypedArrays?

One feature that would be really helpful might be some concise, built-in arbitraries for generating TypedArrays (or even just some as-efficient-as possible examples in the readme).

dubzzz:
Hello @bitjson ,

In order to generate _varying lengths of random TypedArray_ you can use one of the `fc.array` signature.

The one to choose depends on the data you want to generate:

- you don't have any constraints on the length: you can use either (1) `fc.array(yourTypeArbitrary)` or (2) `fc.array(yourTypeArbitrary, maxSize: number)`. Signature (1) builds arrays of size between 0 and 10 (inclusive) in order to avoid having too large entries and causing them to run too long. Signature (2) can be used if you are ok to run your code on longer entries
- you want arrays with a maximal length of `maxLength`: you can use `fc.array(yourTypeArbitrary, maxSize: number)`
- you want arrays with a minimal size of `minLength`: `fc.array(yourTypeArbitrary, minSize: number, maxSize: number)`. Current signature forces the user to specify a maximal length too. The problem should be overcome by https://github.com/dubzzz/fast-check/issues/89 but I am still not sure on the way I will handle it

As an example, let's say you want to test a `sort` algorithm, you may generate arrays of integers using `fc.array(fc.integer())` that way:

```js
import * as assert from 'assert';
import * as fc from 'fast-check';
import { stableSort } from './implem';

describe('stable sort', () => {
  it('should produce an ordered array', () =>
    fc.assert(
      fc.property(fc.array(fc.integer()), data => {
        const sorted = stableSort(data, (a, b) => a - b);
        for (let idx = 1; idx < sorted.length; ++idx) {
          assert.ok(sorted[idx - 1] <= sorted[idx]);
        }
      })
    ));
});
```

More on the sort algorithm with fast-check at https://github.com/dubzzz/fast-check-examples/tree/master/test/stable%20sort

I might need to add something on it on the documentation ;)
Does it answers your needs?

bitjson:
Thanks for the quick and thorough response! Sorry, I should have linked before – I'm thinking about arbitraries for ES2015 [TypedArray Views](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Typed_arrays).

[...]

PR #1212, opened 2020-12-03
✨ Add typed arrays arbitraries

dubzzz:
## In a nutshell

Related to #96

BigInt versions of the typed arrays will not be added for the moment as they would imply a major bump as they may impact users with errors like (even if users don't use it at all):

```txt
Error: node_modules/fast-check/lib/types/check/arbitrary/TypedArrayArbitrary.d.ts(83,88): error TS2304: Cannot find name 'BigInt64Array'.
Error: node_modules/fast-check/lib/types/check/arbitrary/TypedArrayArbitrary.d.ts(88,89): error TS2304: Cannot find name 'BigUint64Array'.
```

Extracted from run: https://github.com/dubzzz/fast-check/runs/1527020437

✔️ New feature
❌ Fix an issue
❌ Documentation improvement
❌ Other: *please explain*

(✔️: yes, ❌: no)

## Potential impacts

None.
Just adding a few new arbitraries.


documentation/Arbitraries.md
```diff
@@ -43,6 +43,7 @@ fc.assert(
 - [Combinators](#combinators)
   - [Simple](#simple)
   - [Array](#array)
+  - [Typed Array](#typed-array)
```
dubzzz, line 46: Should it be in Combinators section? Might be better in Numeric as it does not offer any Combinators capabilities 🤔

PR #1250, opened 2020-12-12
💚 Increase stability of e2e on floatNext/doubleNext

dubzzz:
## In a nutshell

❌ New feature
❌ Fix an issue
❌ Documentation improvement
✔️ Other: *flaky*

(✔️: yes, ❌: no)

## Potential impacts

None

### DIFF
--- a/test/e2e/arbitraries/DoubleNextArbitrary.spec.ts
+++ b/test/e2e/arbitraries/DoubleNextArbitrary.spec.ts
@@ -4,7 +4,8 @@ const seed = Date.now();
 
 describe(`DoubleNextArbitrary (seed: ${seed})`, () => {
   describe('doubleNext', () => {
-    const numRuns = 10000;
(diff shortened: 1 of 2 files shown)


test/e2e/arbitraries/DoubleNextArbitrary.spec.ts
```diff
@@ -4,7 +4,8 @@ const seed = Date.now();
 
 describe(`DoubleNextArbitrary (seed: ${seed})`, () => {
   describe('doubleNext', () => {
-    const numRuns = 10000;
+    const limitedNumRuns = 1000;
+    const numRuns = 25000;
```
dubzzz, line 8: We more than double the number of generated values to limit the risk of negative results (in some cases we may not be lucky and generating the extreme values might take longer than expected).

In addition to that we added a test to ensure we can still at least generate one of those values in a very limited set of runs.

PR #1265, opened 2020-12-16
⚡️ Switch from Object.keys to for..in in record

dubzzz:
## In a nutshell

> The Object.keys() method returns an array of a given object's own enumerable properties, in the same order as that provided by a for...in loop (the difference being that a for-in loop enumerates properties in the prototype chain as well).
_From https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys_

Very slight improvement expected as it will do less allocations. Measurements detected no major change.

❌ New feature
❌ Fix an issue
❌ Documentation improvement
✔️ Other: *performance records*

(✔️: yes, ❌: no)

## Potential impacts

Record


dubzzz:
Change more or less reverted by #1321

dubzzz:
Moreover this change was problematic as it iterated over illegal keys (keys coming from the prototype chain)

Issue #1305, opened 2020-12-30
Seed CI jobs


dubzzz:
## 💡 Idea

Adding a seed onto CI jobs, might clearly help in case of infinite loops blocking the main thread.

The `runId` or something derived from it might be used for that purpose.

PR #1306, opened 2020-12-30
💬 Prefer unique symbols over `Symbol.for`

dubzzz:
## In a nutshell

❌ New feature
❌ Fix an issue
❌ Documentation improvement
✔️ Other: *refactoring*

(✔️: yes, ❌: no)

## Potential impacts

Symbols changed but no problem expected except if you users are importing different versions of fast-check at the same time. In that case, symbols will differ between the versions.

### DIFF
--- a/test/unit/check/runner/configuration/GlobalParameters.spec.ts
+++ b/test/unit/check/runner/configuration/GlobalParameters.spec.ts
@@ -3,6 +3,7 @@ import {
   readConfigureGlobal,
   resetConfigureGlobal,
 } from '../../../../../src/check/runner/configuration/GlobalParameters';
+import * as fc from '../../../../../lib/fast-check';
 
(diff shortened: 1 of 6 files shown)


test/unit/check/runner/configuration/GlobalParameters.spec.ts
```diff
@@ -20,4 +21,11 @@ describe('GlobalParameters', () => {
     resetConfigureGlobal();
     expect(readConfigureGlobal()).toBe(undefined);
   });
+  it('should use distinct global configurations for distinct instances of fast-check'() => {
```
dubzzz, line 24:
```suggestion
  it('should use distinct global configurations for distinct instances of fast-check', () => {
```

dubzzz:
Required for #1305

PR #1352, opened 2021-01-15
doc about filter performance

kazchimo:
## Description
`Arbitrary.filter` can harm the performance of generations of values, and this is shown in [the advanced doc](https://github.com/dubzzz/fast-check/blob/master/documentation/AdvancedArbitraries.md#filter-values) but not in the code doc.
This is a very important and fundamental issue when use property based testing libraries so should be written in the code doc explicitly.
The doc content is borrowed from the advanced doc description.

## In a nutshell

❌ New feature
❌ Fix an issue
✔️ Documentation improvement
❌ Other: *please explain*


dubzzz (approved):

dubzzz:
Thanks a lot for the improvement!

PR #1372, opened 2021-01-19
⚡️ Switch towards a faster shrinker for bigint

dubzzz:
## In a nutshell

❌ New feature
❌ Fix an issue
❌ Documentation improvement
✔️ Other: *faster shrinker*

(✔️: yes, ❌: no)

## Potential impacts

Shrinker for bigint


dubzzz:
For the moment, I use the same implementation as the one for integer but it might not be the best option.

PR #1395, opened 2021-01-24
fix: correct misspellings & other errors in English

phormio:


dubzzz (approved):
Thanks a lot for the fixes

PR #1408, opened 2021-01-31
Avoid using unsafe-eval in implementation of getGlobal

wereHamster:
The current implementation of getGlobal is not compatible with CSP (due to its use of `Function('return this')`).

## In a nutshell

✔️ Fix an issue

### DIFF
--- a/src/utils/globalThis.ts
+++ b/src/utils/globalThis.ts
@@ -1,3 +1,5 @@
+// tslint:disable
+
 // See https://github.com/paulmillr/es6-shim/commit/2367e0953edd01ae9a5628e1f47cf14b0377a7d6
 // and https://github.com/tc39/proposal-global/blob/269dd7c646451e72e3e3edc573b57ce766aac86a/README.md#rationale
 
--- a/src/utils/globalThis.ts
+++ b/src/utils/globalThis.ts
@@ -1,5 +1,9 @@
 // tslint:disable
 
+declare const self: any
+declare const window: any
+declare const global: any
+
 // See https://github.com/paulmillr/es6-shim/commit/2367e0953edd01ae9a5628e1f47cf14b0377a7d6
 // and https://github.com/tc39/proposal-global/blob/269dd7c646451e72e3e3edc573b57ce766aac86a/README.md#rationale
 
--- a/src/utils/globalThis.ts
+++ b/src/utils/globalThis.ts
@@ -10,23 +10,25 @@ declare const global: any;
  * Internal polyfill for `globalThis`
  * @internal
  */
-export let getGlobal = () => {
-  const globalThis = (function () {
-    // the only reliable means to get the global object is
-    // `Function('return this')()`
-    // However, this causes CSP violations in Chrome apps.
-    if (typeof self !== 'undefined') {
-      return self;
-    }
-    if (typeof window !== 'undefined') {
-      return window;
-    }
-    if (typeof global !== 'undefined') {
-      return global;
-    }
-    throw new Error('unable to locate global object');
-  })();
+/** @internal */
+const internalGlobalThis: any = (function () {
+  // the only reliable means to get the global object is
+  // `Function('return this')()`
+  // However, this causes CSP violations in Chrome apps.
+  if (typeof self !== 'undefined') {
+    return self;
+  }
+  if (typeof window !== 'undefined') {
+    return window;
+  }
+  if (typeof global !== 'undefined') {
+    return global;
+  }
+  throw new Error('unable to locate global object');
+})();
 
-  getGlobal = () => globalThis;
-  return globalThis;
-};
+/**
+ * Internal polyfill for `globalThis`
+ * @internal
+ */
+export const getGlobal = () => internalGlobalThis;
--- a/src/utils/globalThis.ts
+++ b/src/utils/globalThis.ts
@@ -1,13 +1,21 @@
-// Implementation of globalThis derived from https://github.com/tc39/proposal-global/blob/master/polyfill.js
-
-/** @internal */
-// tslint:disable
-const internalGlobalThis: any = (function (global) {
-  return global.globalThis ? global.globalThis : global;
-})(typeof this === 'object' ? this : Function('return this')());
+// See https://github.com/paulmillr/es6-shim/commit/2367e0953edd01ae9a5628e1f47cf14b0377a7d6
+// and https://github.com/tc39/proposal-global/blob/269dd7c646451e72e3e3edc573b57ce766aac86a/README.md#rationale
 
 /**
  * Internal polyfill for `globalThis`
  * @internal
  */
-export const getGlobal = () => internalGlobalThis;
+export let getGlobal = () => {
+  const globalThis = (function () {
+    // the only reliable means to get the global object is
+    // `Function('return this')()`
+    // However, this causes CSP violations in Chrome apps.
+    if (typeof self !== 'undefined') { return self; }
+    if (typeof window !== 'undefined') { return window; }
+    if (typeof global !== 'undefined') { return global; }
+    throw new Error('unable to locate global object');
+  })();
+  
+  getGlobal = () => globalThis
+  return globalThis;
+};
--- a/src/utils/globalThis.ts
+++ b/src/utils/globalThis.ts
@@ -1,11 +1,10 @@
-// tslint:disable
-
-declare const self: any
-declare const window: any
-declare const global: any
-
 // See https://github.com/paulmillr/es6-shim/commit/2367e0953edd01ae9a5628e1f47cf14b0377a7d6
 // and https://github.com/tc39/proposal-global/blob/269dd7c646451e72e3e3edc573b57ce766aac86a/README.md#rationale
+
+// tslint:disable
+declare const self: any;
+declare const window: any;
+declare const global: any;
 
 /**
  * Internal polyfill for `globalThis`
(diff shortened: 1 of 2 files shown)


src/utils/globalThis.ts
```diff
@@ -1,13 +1,32 @@
...
+declare const window: any;
+declare const global: any;
 
 /**
  * Internal polyfill for `globalThis`
  * @internal
  */
-export const getGlobal = () => internalGlobalThis;
+export let getGlobal = () => {
+  const globalThis = (function () {
```
dubzzz, line 14: I'd rather keep this globalThis outside of the function otherwise its computation will be done over and over.

Here is an illustration in codesandbox.
See: https://codesandbox.io/s/flamboyant-field-zopcx?file=/src/globalThis.spec.ts

Explanation of the codesandbox:
- Go in the test tab.
- I tweaked the code to add a console log just to confirm that it would be executed twice while I expect the result to be cached.

src/utils/globalThis.ts
```diff
@@ -1,13 +1,32 @@
...
+    }
+    if (typeof global !== 'undefined') {
+      return global;
+    }
+    throw new Error('unable to locate global object');
+  })();
+
+  getGlobal = () => globalThis;
+  return globalThis;
+};
```
dubzzz, line 32:
```suggestion
/** @internal */
const internalGlobalThis: any = (function () {
  // the only reliable means to get the global object is
  // `Function('return this')()`
  // However, this causes CSP violations in Chrome apps.
  if (typeof self !== 'undefined') {
    return self;
  }
  if (typeof window !== 'undefined') {
    return window;
  }
  if (typeof global !== 'undefined') {
    return global;
  }
  throw new Error('unable to locate global object');
})();

/**
 * Internal polyfill for `globalThis`
 * @internal
 */
export const getGlobal = () => internalGlobalThis;
```

In order to avoid the double-evaluation.

src/utils/globalThis.ts
```diff
@@ -1,13 +1,32 @@
...
+    if (typeof self !== 'undefined') {
+      return self;
+    }
+    if (typeof window !== 'undefined') {
+      return window;
+    }
+    if (typeof global !== 'undefined') {
+      return global;
+    }
+    throw new Error('unable to locate global object');
```
dubzzz, line 27: Just wondering: Is there a risk to fall in this case?

wereHamster: I don't know of any JS runtime which would fall into this case.

src/utils/globalThis.ts
wereHamster: The tests work for me.

The code should be executed only once because `getGlobal` replaces itself with `() => globalThis` after running the first time. Think of it as a lazily-evaluated `globalThis` where the detection code runs only when the `getGlobal` function is actually used. If it's never called, the detection code (which potentially throws an Error) never runs.

dubzzz: Actually, the replacement will not work in the case of an import. What you'll import is the lazy function, it'll never be updated.

dubzzz: Alternatively, we can just don't do the optim. After all, keeping the global is pretty useless given the length and complexity of code.

wereHamster: According to ES Module semantics the let binding should update: https://stackoverflow.com/a/32558929

dubzzz: Yes, I know. But only if not exported. Let's take this code:

```js
import {getGlobal} from './'
const initialGetGlobal = getGlobal;
getGlobal();
assert.equal(getGlobal, initialGetGlobal); // will not fail, getGlobal will not be changed even with the let trick
```

I'll come back with a repro outside of codesandbox

wereHamster: Non-compliant ES module bundler 🤷 . I committed your suggestion.

dubzzz: It seems that you are right. 

The problem will not be with esm but cjs (I still support it). An user that does `var { getGlobal } = require("./global.js")` will relaunch the computation on each call without using the cached one.

Just for the record here is the repro for CJS:

```js
// global.js
"use strict";

var _getGlobal = function getGlobal() {
  var globalThis = (function () {
    console.log("iife");
    if (typeof self !== "undefined") return self;
    if (typeof window !== "undefined") return window;
    if (typeof global !== "undefined") return global;
    throw new Error("unable to locate global object");
  })();

  exports.getGlobal = _getGlobal = function getGlobal() {
    return globalThis;
  };

  return globalThis;
};

exports.getGlobal = _getGlobal;

// play.js
"use strict";
const { getGlobal } = require("./global.js");

const initialGetGlobal = getGlobal;
getGlobal();
console.log(initialGetGlobal === getGlobal);
getGlobal();

// output of: 
// $ node play.js 
iife
true
iife
```

No problem with ESM, as you said:

```js
// global.mjs
export let getGlobal = () => {
  const globalThis = (function () {
    console.log("iife");
    if (typeof self !== "undefined") return self;
    if (typeof window !== "undefined") return window;
    if (typeof global !== "undefined") return global;
    throw new Error("unable to locate global object");
  })();
  getGlobal = () => globalThis;
  return globalThis;
};

// play.mjs
import { getGlobal } from "./global.mjs";

const initialGetGlobal = getGlobal;
getGlobal();
console.log(initialGetGlobal === getGlobal);
getGlobal();

// output of:
// $ node play.mjs
iife
false
```

Anyway, I prefer the version with an extracted code, it looks easier.

dubzzz:
Thanks you so much for this fix and so sorry about the esm discussion.
Just in case you need the fix: do you want me to release it? Do you have a precise minor version you want?

Your PR remind me that I should clearly fix the Netlify step of my CI as it will fail for non admins.

wereHamster:
It would be nice to have a release with this patch. I'm currently using 2.4.0 but I'm not using anything advanced so I could very well update to the latest (2.12.x).

dubzzz:
Ok, I'm gonna release the patch on 2.12.x for the moment (I changed my CI from Travis to GH Actions between 2.4.x and 2.12.x, backport will be not easy I think).

I'll let you now once released (maybe only tonight French time).

dubzzz:
@wereHamster Version 2.12.1 is on its way. No breaking change expected between 2.4.0 and 2.12.1

wereHamster:
There was one non-BC change, I got failures in the use of  `fc.constant('Dashboard')`. It seems previously the type was inferred as `fc.Arbitrary<'Dashboard'>` while with the new version it's `fc.Arbitrary<string>`

dubzzz:
🤔 I'll have a look to it. Clearly not expected, thanks for the report

wereHamster:
Maybe it's related to nested records:

```
interface Query {
  __typename: "Query",
  data: {
    __typename: "X" | "Y"
  }
}

const arb = fc.record<Query>({
  __typename: fc.constant("Query"), // infers ok
  data: fc.record({
    __typename: fc.constant<"X">("X"), // needs type param
  }),
})
```

wereHamster:
TypeScript 3.9.7

dubzzz:
Another trick would be to use: `fc.constant("X" as const)`

Pretty strange, I have it working on my side:

```ts
// index.ts
import fc from "fast-check";
interface Query {
  __typename: "Query";
  data: {
    __typename: "X" | "Y";
  };
}
const arb = fc.record<Query>({
  __typename: fc.constant("Query"),
  data: fc.record({
    __typename: fc.constant("X"),
  }),
});
```

With command: `tsc index.ts --target es6 --moduleResolution node`
And fast-check@2.12.1, typescript@3.9.7

wereHamster:
Can you try with this?

```
interface Query {
  __typename: 'Query'
  data?: { __typename: 'X' }
}

export const arb = fc.record<Query>({
  __typename: fc.constant('Query'),
  data: fc.record({
    __typename: fc.constant('X'),
  }),
})
```

It seems the optionality of the `data` field causes the problem. If I change `data?: …` to `data: …` then there is no error.

dubzzz:
So strange, it works too 🤔 

With this setup:
```sh
$ cat node_modules/fast-check/package.json | grep version
  "version": "2.12.1"

$ cat node_modules/typescript/package.json | grep version
  "version": "3.9.7"
```

dubzzz:
By the way, we may add your typing check in https://github.com/dubzzz/fast-check/blob/master/test/type/main.ts just to prevent it to change in the future... It checks regressions in typings against several versions of TypeScript (but it was not there in 2.4.0 as I used an other tool for that purpose in the past).

wereHamster:
Can you try with strict (`tsc --strict`)?

wereHamster:
(btw I was wrong about my TS version, it's 4.0.3)

dubzzz:
Ok, I can reproduce with `--strict`. 

It seems that the regression comes from version 2.11.0. I'll check why it changed.

dubzzz:
I am investigating this issue. I detected that the regression has been introduced by https://github.com/dubzzz/fast-check/commit/2684269ec564503e4158c774239b49c2258fc718. Still checking was happened and how it should be fixed.

dubzzz:
@wereHamster I finally found a fix for the regression you sent me. Please let me now if it would fully fix the issue you noticed. If so, I'll backport the fix in 2.11.x (version that introduced the regeression) and 2.12.x

- PR for the fix: https://github.com/dubzzz/fast-check/pull/1465
- Instructions to install the version locally: https://github.com/dubzzz/fast-check/pull/1465#issuecomment-775491790

dubzzz:
Here is a quick update related to this issue:
- unsafe-eval fix has been back-ported to all versions of fast-check >=2.7.0
- the typing regression on record has been fixed for 2.11.x and 2.12.x (the only versions including the issue)

Issue #1492, opened 2021-02-13
Plan for fast-check@3 (Not yet scheduled) 


dubzzz:
Next major version of fast-check may include:
- [x] drop support for TypeScript <4.0, by doing this `tuple`, `property` could be properly typed
- [x] drop `genericTuple` that will not be needed anymore once `tuple` will be able to handle variadic parameters properly
- [x] drop deprecated signatures and move to constraints-based once. The case of `integer` or `bigInt` is still unclear
- [ ] ~worst scenario by default, fast-check is supposed to find bugs by itself not to be guide to them, as a consequence the default should cover worst cases and users should disable them whenever not applicable (see `double` with flag `next`)~
- [ ] ~ready for tree shaking, adopt a better file structure. Even if tree shaking is not a must have for fast-check, the file structure should be reworked~

Other possible ideas: https://twitter.com/ndubien/status/1366704239825866752?s=19

dubzzz:
Released!

Issue #992, opened 2020-09-20
RFC - Unify signatures accross arbitraries


dubzzz:
## 💡 Idea

**Aim**

A single way to customize arbitaries of fast-check.

Contrary to existing: `fc.integer(1, 10)` (where 1 is for min and 10 for max) vs `fc.object({maxKeys: 5})` (and not `fc.object(5)`)

**Existing signatures**

Today, the signatures of the built-in arbitraries coming with fast-check can be separated into two categories:
- the ones with overloads
- the ones with constraints as an objects

For the overloading ones you can think about `fc.array`. It accepts three main overloads: `fc.array(arb)`, `fc.array(maxLength)` and `fc.array(minLength, maxLength)`.

For the constraints-based ones, you can think about `fc.webUrl`. It only has two signatures: `fc.webUrl()` and `fc.webUrl(constraints)`.

**History**

Most of the legacy arbitraries have been defined with an overloading strategy.

Over time this strategy proved weak: it did not scale to the increasing number of options required by new arbitraries, it appeared costly to develop (as overloading does not exist in JavaScript) and it was and is ambiguous for users.

_Why costly to develop?_

Let's suppose we want to develop something like `fc.array` with the overloads: `fc.array(arb)`, `fc.array(maxLength)` and `fc.array(minLength, maxLength)`. Our code would be something like:

```js
function array(arb, aLength, bLength) {
  const minLength = bLength !== undefined ? aLength : 0;
  const maxLength = bLength !== undefined ? bLength : aLength !== undefined ? aLength : 10;
  return new ArrayArbitrary(arb, minLength, maxLength);
}
```

This one is pretty simple as it only exposes two parameters but we already see how complex it can be to just know what would be the `minLength` or `maxLength` to apply.

_Why ambiguous?_

The second main concern is ambiguity. For someone not knowing the API it is not very easy to know what means `2` in the following signatures `fc.array(fc.nat(), 2)` or `fc.array(fc.nat(), 2, 2)` or `fc.array(fc.nat(), 2, 3)`.

As a consequence even in the tests of fast-check itself, we have been using `fc.array(arb, minLength, maxLength)` signature when just wanted to set the `maxLength` (while we could have used a shorter version: `fc.array(arb, maxLength)`).

Signatures for `fc.set` might be even more ambiguous.

_Difficult to express everything_

[...]

PR #1495, opened 2021-02-14
💥 Remove deprecated signatures of `fc.float`/`fc.double`

dubzzz:
## In a nutshell

Related to #1492 
Follow-up of #992

❌ New feature
❌ Fix an issue
❌ Documentation improvement
✔️ Other: *remove deprecated*

(✔️: yes, ❌: no)

## Potential impacts

Remove deprecated signatures of `fc.float` and of `fc.double`.


src/check/arbitrary/ObjectArbitrary.ts
```diff
@@ -397,7 +397,7 @@ function jsonSettings(stringArbitrary: Arbitrary<string>, constraints?: number |
   const values = [
     boolean(),
     maxSafeInteger(),
-    double({ next: true, noDefaultInfinity: true, noNaN: true }),
+    double({ noDefaultInfinity: true, noNaN: true }).filter((d) => !Object.is(d, -0)),
```
dubzzz, line 400: Fix has to be done on `main` branch too. It should also be back-ported as the change (from `double()` to `double({next:true})`) introduced a breaking change.

test/unit/utils/stringify.spec.ts
```diff
@@ -53,12 +53,9 @@ describe('stringify', () => {
   }
   it('Should be equivalent to JSON.stringify for JSON compliant objects', () =>
     fc.assert(
-      fc.property(
-        fc.anything({ values: [fc.boolean(), fc.integer(), fc.double(), fc.fullUnicodeString(), fc.constant(null)] }),
-        (obj) => {
-          expect(stringify(obj)).toEqual(JSON.stringify(obj));
-        }
-      )
+      fc.property(fc.unicodeJsonObject(), (obj) => {
```
dubzzz, line 56: Fix has to be done on `main` branch too

src/check/arbitrary/ObjectArbitrary.ts
dubzzz: It does not seem that `-0` is an illegal value.

Nonetheless, there is a strange behaviour in JavaScript with negative zeros and `JSON.*` methods:
- `JSON.parse('{"a": -0}')` preserves `-0`
- `JSON.stringify({a: -0})` does not and produces `{"a":0}`

dubzzz: See https://tools.ietf.org/html/rfc7159#page-6

PR #1504, opened 2021-02-15
💥 Remove deprecated signatures of `fc.lorem`

dubzzz:
## In a nutshell

Related to #1492 
Follow-up of #992

❌ New feature
❌ Fix an issue
❌ Documentation improvement
✔️ Other: *remove deprecated*

(✔️: yes, ❌: no)

## Potential impacts

Remove deprecated signatures of `fc.lorem`.

### DIFF
--- a/perf/tasks.js
+++ b/perf/tasks.js
@@ -1,6 +1,6 @@
 exports.runComplexFailure = function (fc) {
   const loremIpsum = fc.record({
-    text: fc.lorem(100),
+    text: fc.lorem({ maxCount: 100 }),
     type: fc.constant('x'),
     attrs: fc.constant({}),
     markup: fc.option(
--- a/src/check/arbitrary/LoremArbitrary.ts
+++ b/src/check/arbitrary/LoremArbitrary.ts
@@ -222,7 +222,7 @@ const loremWord = () =>
  * @public
  */
 function lorem(constraints: LoremConstraints = {}): Arbitrary<string> {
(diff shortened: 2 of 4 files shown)


perf/tasks.js
```diff
@@ -1,6 +1,6 @@
 exports.runComplexFailure = function (fc) {
   const loremIpsum = fc.record({
-    text: fc.lorem(100),
+    text: fc.lorem({ maxCount: 100 }),
```
dubzzz, line 3: Fix has to be done on `main` branch too

src/check/arbitrary/LoremArbitrary.ts
```diff
@@ -213,54 +213,18 @@ const loremWord = () =>
...
- * @param constraints - Constraints to be applied onto the generated value
- *
- * @remarks Since 2.5.0
- * @public
- */
-function lorem(constraints: LoremConstraints): Arbitrary<string>;
-function lorem(...args: [] | [number] | [number, boolean] | [LoremConstraints]): Arbitrary<string> {
-  const maxWordsCount = typeof args[0] === 'object' ? args[0].maxCount : args[0];
-  const sentencesMode = typeof args[0] === 'object' ? args[0].mode === 'sentences' : args[1];
-  const maxCount = maxWordsCount || 5;
```
dubzzz, line 261: This was actually a bug! The max value specified by the user was overridden when passing 0.

PR #1522, opened 2021-02-17
📝 Do not package with comments (except d.ts)

dubzzz:
## In a nutshell

It seems that VSCode uses the JSDoc coming with d.ts file. Same for WebStorm.

As a consequence it does not make really sense to package fast-check with all those jsdoc comments again and again in JS files if they are never read.

❌ New feature
❌ Fix an issue
✔️ Documentation improvement
❌ Other: *please explain*

(✔️: yes, ❌: no)

## Potential impacts

Documentation


dubzzz:
@github-actions deploy

PR #1535, opened 2021-02-18
🐛 fix nodejs import bug

osdevisnot:
## In a nutshell

❌ New feature
✔️ Fix an issue
❌ Documentation improvement
❌ Other: *please explain*

(✔️: yes, ❌: no)

## Potential impacts

If you expect some noticeable impacts with your change please explain.

- Usage Impact - As [documented](https://nodejs.org/api/esm.html#esm_import_specifiers), modern veersion of node.js needs an explicit `import` specifier to be able to use ESM format in node.js

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,8 @@
   "exports": {
     ".": {
       "require": "./lib/fast-check.js",
-      "default": "./lib/esm/fast-check.js"
+      "default": "./lib/esm/fast-check.js",
+      "import": "./lib/esm/fast-check.js"
     }
   },
   "module": "lib/esm/fast-check.js",
--- a/package.json
+++ b/package.json
@@ -7,8 +7,8 @@
   "exports": {
     ".": {
       "require": "./lib/fast-check.js",
-      "default": "./lib/esm/fast-check.js",
-      "import": "./lib/esm/fast-check.js"
+      "import": "./lib/esm/fast-check.js",
+      "default": "./lib/esm/fast-check.js"
     }
   },
   "module": "lib/esm/fast-check.js",
--- a/test/esm/esbuild-with-import/package.json
+++ b/test/esm/esbuild-with-import/package.json
@@ -1,7 +1,7 @@
 {
   "scripts": {
     "build:browser": "esbuild main.js --bundle --platform=browser --outfile=dist/main.browser.js",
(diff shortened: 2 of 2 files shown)


osdevisnot:
This PR is absolete.

dubzzz:
Actually I planned to at it too. But it looks that it works without it as I specified 'default'

osdevisnot:
Yeah I didn't realize default is used for import when import is missing.

dubzzz:
Actually it may fix the issue I got with esbuild 🤔

dubzzz:
From what I have seen in the CI run, default should be used after import (it should come last).

If your suggestion works then the CI run testing CJS/ESM should be marked as failure because of esbuild (current test checks that I get CommonJS and not module)

package.json
```diff
@@ -7,7 +7,8 @@
   "exports": {
     ".": {
       "require": "./lib/fast-check.js",
-      "default": "./lib/esm/fast-check.js"
+      "default": "./lib/esm/fast-check.js",
+      "import": "./lib/esm/fast-check.js"
```
dubzzz, line 11:
```suggestion
      "import": "./lib/esm/fast-check.js",
      "default": "./lib/esm/fast-check.js"
```

It seems that Webpack does not like when default is not at the end.

osdevisnot:
> If your suggestion works then the CI run testing CJS/ESM should be marked as failure because of esbuild (current test checks that I get CommonJS and not module)

I am not entirely familiar with CI setup here. But let me know if this needs any other change.

Also, since you are running the generated output with `node.js`, it is definitely not an esbuild bug. If the bug persists after this change, we might need to set `"type":"module"` in the `package.json` of the test directory to resolve this.

dubzzz:
Actually, the issue is only due to the fact that esbuild does not support exports yet. So cjs will be imported when using node platform as esbuild uses main field in that case (does not parse overrides suggested in exports)

dubzzz:
Given exports are ignored, the issue is still there.

Given the fix does not change the situation on esbuild and might break others*, I'll not merge it.

_Most of the time, it is highly recommended to do a major bump when changing exports. While this change should be impact less, I prefer not merging it as esbuild state is unchanged._

Thanks a lot for the suggestion 😊

PR #1537, opened 2021-02-19
WIP Better recursive structures

dubzzz:
This Pull Request is not suitable to be merged. It will be re-split into multiple smaller ones in the future.

The idea is to POC stuff around the recursive structures and add supports for clever shrinking for them. The proposal assessed by this PR is to introduced a notion of cross-shrink that can be used by users to specify if `option`, `oneof` and `frequency` have the right to try to shrink onto other arbitraries.

_*as `option` has not been covered yet so they are replaced by `fc.oneof(fc.constant(null), arb)` in the following snippets_
_**the current api placed constraints as the first argument in `oneof` and `frequency` for typings reasons. They will be moved at the end in next major as fast-check will drop support for old versions of TypeScript._

It could be used to declare recursive structures like:

```js
const recursionConstraints = {depthBiasFactor: 0.5, depthContext: 'tree'};
//const recursionConstraints = {maxDepth: 3, depthContext: 'tree'}; // to build fc.memo with fc.letrec

const treeArb = fc.letrec(tie => ({
  tree: fc.record({
    value: fc.integer({min: 1}),
    left: fc.option(tie('tree'), recursionConstraints),
    right: fc.option(tie('tree'), recursionConstraints),
  })
})).tree;

// Printing a sample of values
console.log(fc.sample(treeArb, 3));

// Running against a falsy property to check shrinker
function traverse(node) {
  if (node === null) return [];
  return [node.value, ...traverse(node.left), ...traverse(node.right)];
}
fc.assert(
  fc.property(treeArb, (tree) => {
    const t = traverse(tree);
    return t.reduce((acc, cur) => acc + cur, 0) < 3;
  })
);

// Or another one
fc.assert(
  fc.property(treeArb, (tree) => {
    const t = traverse(tree);
    return t.length === new Set(t).size;
  })
);

// Or another one
fc.assert(
  fc.property(treeArb, (tree) => {
    const t1 = traverse(tree.left);
    const t2 = traverse(tree.right);
    return t1.length === t2.length;
  })
);
```


dubzzz:
@github-actions deploy

dubzzz:
@github-actions deploy

dubzzz:
@github-actions deploy

dubzzz:
@github-actions deploy

dubzzz:
In the context of the [shrinking-challenge](https://github.com/jlink/shrinking-challenge), fast-check performed poorly on the calculator one. Calculator was a challenge involving recursive data-structures.

The first benefit of this PR regarding this challenge is that for the end-user it seems that the syntax will be more friendly:
```diff
   const { expression } = fc.letrec((tie) => ({
     expression: fc.oneof(
---    fc.integer(), // 3x fc.integer()
---    fc.integer(), // otherwise fast-check might generate infinite structures
---    fc.integer(), // 3 cases over 5 are terminal case, the 2 others multiply by 2 the number of terms
+++    { withCrossShrink: true, depthBiasFactor: 0.5 }, // this argument might be optional in next majors
+++    fc.integer(),
       fc.tuple(fc.constant("+"), tie("expression"), tie("expression")),
       fc.tuple(fc.constant("/"), tie("expression"), tie("expression"))
     ),
   }));
```

The second benefit is that now it shrinks a bit more. The [previous status](https://github.com/jlink/shrinking-challenge/blob/18991fd74e833d793d195f5b2adec611b7e45442/pbt-libraries/fast-check/reports/calculator.md) was chaotic with 97 distinct results and no great leader. The new one is a bit more promising with 63 distinct values and some clear leaders (only 100 runs so 2% is not a real leader):
* ``["/",0,["/",0,-1]]`` (11.00%)
* ``["/",0,["/",0,1]]`` (8.00%)
* ``["/",0,["/",["/",0,-1],-1]]`` (4.00%)
* ``["+",["/",0,["/",0,1]],0]`` (4.00%)
* ``["+",["/",0,["/",0,-1]],0]`` (4.00%)
* ``["/",0,["/",["/",0,1],-1]]`` (3.00%)
* ``["/",0,["/",["/",0,1],1]]`` (3.00%)
* ``["/",["/",0,["/",0,-1]],1]`` (2.00%)
* ``["/",0,["/",["+",0,0],-1]]`` (2.00%)
* ``["+",["+",["/",0,["/",0,1]],0],0]`` (2.00%)

PR #1572, opened 2021-03-02
Improve types of tie fn in letrec arbitrary

### LINKED ISSUE #1164: Typing improvement for `letrec` combinator
## 🚀 Feature Request

Improve typing of `tie` function in `letrec` combinator by expanding the type signature with the following, allowing it to read from the generic `T`.

```typescript
tie: {
    <K extends keyof T>(key: K): Arbitrary<T[K]>;
    (key: string): Arbitrary<unknown>;
}

// so the full signature becomes:
export declare function letrec<T>(
  builder: (tie: {
    <K extends keyof T>(key: K): Arbitrary<T[K]>;
    (key: string): Arbitrary<unknown>;
  }) => { [K in keyof T]: Arbitrary<T[K]> },
): { [K in keyof T]: Arbitrary<T[K]> };
```

## Motivation

The way how `tie` function of `letrec` combinator is typed means that we have to explicitly cast result of each call to that function. This way it is only needed to specify all return types of all arbitraries in `T`.

## Example

For example, reusing letrec tree from docs. 

```typescript
import * as fc from 'fast-check';

const fst = fc.letrec((tie) => ({
  tree: fc.oneof(tie('node'), tie('leaf'), tie('leaf')),
  node: fc.record({
    left: tie('tree'),
    right: tie('tree'),
  }),
  leaf: fc.nat(),
}));

type Tree = number | { left: Tree; right: Tree };
const snd = fc.letrec<{
  tree: Tree;
  node: { left: Tree; right: Tree };
  leaf: number;
}>((tie) => ({
  tree: fc.oneof(tie('node'), tie('leaf'), tie('leaf')),
  node: fc.record({
    left: tie('tree'),
    right: tie('tree'),
  }),
  leaf: fc.nat(),
}));
```

In the first example, without explicit type specification, it all still defaults to unknown. However, in the second example with explicit type specification, everything becomes well typed.

qlonik:
Closes #1164

I've updated documentation and example usage in `BinaryTreeArbitrary.ts` file. I'm not sure what else should be updated.

## In a nutshell

❌ New feature
✔️ Fix an issue
✔️ Documentation improvement
❌ Other: *please explain*

(✔️: yes, ❌: no)


src/check/arbitrary/LetRecArbitrary.ts
```diff
@@ -74,7 +79,7 @@ function isLazyArbitrary(arb: Arbitrary<any> | undefined): arb is LazyArbitrary
  * @public
  */
 export function letrec<T>(
-  builder: (tie: (key: string) => Arbitrary<unknown>) => { [K in keyof T]: Arbitrary<T[K]> }
+  builder: (tie: TieFunction<T>) => { [K in keyof T]: Arbitrary<T[K]> }
```
dubzzz, line 82: From what I see from the tests only targeting typings (see https://github.com/dubzzz/fast-check/blob/main/test/type/main.ts), the change does not seem to be backward compatible in its current shape.

We have errors like:
```txt
Error: main.ts(291,3): error TS2345: Argument of type '"No recursion \"letrec\""' is not assignable to parameter of type '{ expected: { a: Arbitrary<number>; b: Arbitrary<string>; }; got: {}; }'.
Error: main.ts(298,3): error TS2345: Argument of type '"Recursive \"letrec\""' is not assignable to parameter of type '{ expected: { a: Arbitrary<number>; b: Arbitrary<unknown>; }; got: {}; }'.
Error: main.ts(305,3): error TS2345: Argument of type '"Invalid recursion \"letrec\""' is not assignable to parameter of type '{ expected: { a: Arbitrary<number>; b: Arbitrary<unknown>; }; got: {}; }'.
```

qlonik: <del>This is very unfortunate. I've tried investigating and debugging it, but I have ran out of time. As far as I can see, as soon as generic `T` is used somewhere, it starts participating in the TypeScript inference. Maybe, if it was possible to mark an argument as excluded from participating in inference, it would work (but this is just a guess). There is a pretty old issue about that https://github.com/microsoft/TypeScript/issues/14829 too. However, when I tried some of the ideas in that issue, they do not work. There is weirdness that if you don't use `tie` function in the callback, then everything works, but as soon as parameter to callback is specified, it breaks.</del>

EDIT:
As a last ditch effort, I tried to put Cb into generic, and it seems to work? I hope... :) 
The issue that might be is that someone could explicitly overwrite that Cb generic, but that might be noted as an unsupported use-case.

qlonik: Okay, so now only 3.2 fails in CI, I'm not sure why, it seems that locally 3.2 passes for me?

qlonik: @dubzzz Do you have any ideas why only 3.2 might be failing? Thanks.

dubzzz: No idea for the moment, I'll try to investigate by checkouting the branch locally 🤔 (really sorry, I missed your previous message)

dubzzz:
If only 3.2 fails, the PR could possibly make it for 3.x of fast-check. I'll investigate the 3.2 issue

dubzzz:
I confirm I can successfully reproduce the issue detected by the CI:
image

To do so, I basically reproduced the setup used by the CI:
- `yarn --frozen-lockfile`
- `yarn build:prod-ci`
- `yarn pack`
- `mv fast-check-*.tgz fast-check.tgz`
- `cd test/type`
- `yarn add ../../fast-check.tgz`
- `sed -i 's/ts-expect-error/ts-ignore/g' *.ts`
- `sed -i 's/fc-require-ts-3.4/ts-ignore/g' *.ts`
- `npx -p typescript@3.2 tsc`

I'm trying to understand what makes TS complains...

src/check/arbitrary/LetRecArbitrary.ts
```diff
@@ -75,8 +80,8 @@ function isLazyArbitrary(arb: Arbitrary<any> | undefined): arb is LazyArbitrary
  * @remarks Since 1.16.0
  * @public
  */
-export function letrec<T>(
-  builder: (tie: (key: string) => Arbitrary<unknown>) => { [K in keyof T]: Arbitrary<T[K]> }
+export function letrec<T, Cb extends TieFunction<T> = TieFunction<T>>(
```
dubzzz, line 83: It seems that defining an overload helps to fix the issue reported by TS 3.2 🤔 

```ts
function letrec<T>(builder: (tie: (key: string) => Arbitrary<unknown>) => { [K in keyof T]: Arbitrary<T[K]>; }): { [K in keyof T]: Arbitrary<T[K]>; };
function letrec<T, Cb extends TieFunction<T> = TieFunction<T>>(builder: (tie: Cb) => { [K in keyof T]: Arbitrary<T[K]>; }): { [K in keyof T]: Arbitrary<T[K]>; };
```

dubzzz: Actuallly does not seem to work well 😢 

I tried to check the typings by using:
```ts
expectType<{ a: fc.Arbitrary<number>; b: fc.Arbitrary<string>; c: fc.Arbitrary<string> }>()(
  fc.letrec<{ a: number; b: string; c: string }>((tie) => ({
    a: fc.nat(),
    b: fc.string(),
    c: tie('b'),
  })),
  'Invalid "letrec"'
);
```

But it fails to type it...

dubzzz:
@qlonik I synced my 3.x branches of fast-check with latest release (2.17), by targeting next-2_17_0 you'll open a PR targeting the next major. It should remove the issue linked to 3.2 as I'll not support it anymore in 3.x.

As moved lots of code since your PR, do you want me to handle the merge conflict? Just want to make sure it would be ok with you if I update the PR, or tell me if you prefer to do it on your side.

qlonik:
Done :rocket:. I'm sorry it took some time. 

For now I didn't put another commit extracting `TieFunction` as generic `Cb`. I hope that it will work well without that change, since that change was rather weird.

Something weird happened with checks, I'm guessing because of me changing the base branch. Could you please rerun those checks?

src/arbitrary/letrec.ts
```diff
@@ -2,6 +2,11 @@ import { LazyArbitrary } from './_internals/LazyArbitrary';
 import { Arbitrary } from '../check/arbitrary/definition/Arbitrary';
 import { convertFromNext, convertToNext } from '../check/arbitrary/definition/Converters';
 
+export interface TieFunction<T> {
```
dubzzz, line 5: - Add a small JSdoc, just a few words + the `@public` at the end of the block.
- Import and Re-Export it from src/fast-check-default.ts

documentation/Arbitraries.md
```diff
@@ -3075,6 +3075,26 @@ fc.letrec(tie => ({
...
+  leaf: number;
+}>((tie) => ({
+  tree: fc.oneof(tie('node'), tie('leaf'), tie('leaf')),
+  node: fc.record({
+    left: tie('tree'),
+    right: tie('tree'),
+  }),
+  leaf: fc.nat(),
+}));
+```
```
dubzzz, line 3097: Instead of the example in the doc Arbitrary.md mostly explaining the primitives from a JS point of view, I'd rather update the example in the `@example` of `letrec` to add the typing trick.

Something like:
```txt
 * // Specify recursive type explicitly, since TS cannot infer recursive types
 * type Tree = number | [Tree, Tree];
 * // Explicitly specify types of each recursive constructors
 * const { tree } = fc.letrec<{
 *   tree: Tree;
 *   node: [Tree, Tree];
 *   leaf: number;
 * }>(tie => ({
 *   tree: fc.oneof({depthFactor: 0.5}, tie('leaf'), tie('node')),
 *   node: fc.tuple(tie('tree'), tie('tree')),
 *   leaf: fc.nat()
 * }));
 * // tree is 50% of node, 50% of leaf
 * // the ratio goes in favor of leaves as we go deeper in the tree (thanks to depthFactor)
```

As the pattern `fc.name<ExpectedType>` is mode and more spreading, having it only inside the jsdoc looks legit.

dubzzz (commented):
Thank you so much for the update. I put some comments and launched the tests (GitHub blocked the execution, it asked me for a manual validation).

_Notes regarding the 3.x version._
It's not ready yet. I have a 3.x release channel on the tag `fast-check@next` but for the moment it is mostly a `2.x` with breaking changes (the ones marked with `@deprecated` today). I hope to be able to release it before the end of the year but it implies many structural changes to be fully ready (some of them have already been released behind the scene since 2.15 others are on-going). Each 2.x version will have it's next-2_x branch (they always come with all the changes of the previous next).

dubzzz:
@qlonik Sorry for the delay, v3 took me longer than expected to set up. But it's closer. I retried you change locally I made some minor modifications to it and reached this stuff that seems to work without any breaking change 🤔 

```ts
type LetRecValue<T> = {
  [K in keyof T]: Arbitrary<T[K]>;
};

type Not<T> = T extends true ? false : true;
type And<T, U> = T extends true ? (U extends true ? true : false) : false;
type IsNever<T> = [T] extends [never] ? true : false;
type Extends<T, U> = T extends U ? true : false;
type ExtendsString<T> = Extends<T, string> extends boolean
  ? boolean extends Extends<T, string>
    ? true
    : false
  : false; // Extends<T, string> is: false for unknown but boolean for any
type IsUnknown<T> = And<And<Not<IsNever<T>>, Extends<T, unknown>>, And<Extends<unknown, T>, Not<ExtendsString<T>>>>;
type TypedTieFunction<T> = <K extends keyof T>(key: K) => Arbitrary<T[K]>;
type TypedBuilderFunction<T> = (tie: TypedTieFunction<T>) => { [K in keyof T]: Arbitrary<T[K]> };

declare function letrec<T>(...params: IsUnknown<T> extends true ? [TypedBuilderFunction<T>, never] : [TypedBuilderFunction<T>]): LetRecValue<T>;
declare function letrec<T>(builder: (tie: (key: string) => Arbitrary<unknown>) => { [K in keyof T]: Arbitrary<T[K]> }): LetRecValue<T>;
```

I'll probably re-open a PR and add you as contributor so that you get granted for it too: as lots of the hard work has been done by your contribution.

dubzzz:
And even shorter with this one:

```ts
type LetRecValue<T> = {
  [K in keyof T]: Arbitrary<T[K]>;
};
type TypedTieFunction<T> = <K extends keyof T>(key: K) => Arbitrary<T[K]>;
type TypedBuilderFunction<T> = (tie: TypedTieFunction<T>) => { [K in keyof T]: Arbitrary<T[K]> };
declare function letrec<T>(...params: T extends {} ? [TypedBuilderFunction<T>] : [TypedBuilderFunction<T>, never]): LetRecValue<T>;
declare function letrec<T>(builder: (tie: (key: string) => Arbitrary<unknown>) => { [K in keyof T]: Arbitrary<T[K]> }): LetRecValue<T>;
```

PR #1586, opened 2021-03-06
Add a way to ignore already covered cases in property check

### LINKED ISSUE #1576: Unnecessary repetitions
## 🐛 Bug Report

Fast-check shouldn't run more tests that are necessary to check all cases 

## To Reproduce

Steps to reproduce:
```
const arb = fc.record({
    left: fc.boolean(),
    right: fc.boolean(),
});

it('should run test only 4 times', () => {
    fc.assert(fc.property(arb, (data) => {
        console.log(data);
    }));
});
```
If I run this test, it will log data 100 times (default count)

## Expected behavior

Fast-check shouldn't repeat cases and run only 4

## Your environment

| fast-check            |  2.13.0          |
| node                  |  14.16.0          |
| TypeScript           |  4.2.2          |

Djaler:
For some tests possible cases count is very small, and we don't need to repeat them again and again. This feature will allow detecting repetitions and skip them. Resolves #1576

## In a nutshell

✔️ New feature
❌ Fix an issue
❌ Documentation improvement
❌ Other: *please explain*

(✔️: yes, ❌: no)

## Potential impacts

Performance impact due to a decrease of the property check runs.

### DIFF
--- a/documentation/Runners.md
+++ b/documentation/Runners.md
@@ -110,6 +110,8 @@ export interface Parameters<T = void> {
     interruptAfterTimeLimit?: number; // optional, interrupt test execution after a given time limit
                         // in milliseconds (relies on Date.now): disabled by default
     markInterruptAsFailure?: boolean; // optional, mark interrupted runs as failure: disabled by default
+    ignoreEqualValues?: boolean; // optional, do not repeat runs with already covered cases: disabled by default
+    					// This is useful when arbitrary has a limited number of variants
     reporter?: (runDetails: RunDetails<T>) => void; // optional, custom reporter replacing the default one
                         // reporter is responsible for throwing in case of failure, as an example default one throws
                         // whenever `runDetails.failed` is true but it is up to you
--- a/src/check/runner/DecorateProperty.ts
+++ b/src/check/runner/DecorateProperty.ts
@@ -3,11 +3,12 @@ import { SkipAfterProperty } from '../property/SkipAfterProperty';
 import { TimeoutProperty } from '../property/TimeoutProperty';
 import { UnbiasedProperty } from '../property/UnbiasedProperty';
 import { QualifiedParameters } from './configuration/QualifiedParameters';
+import { IgnoreEqualValuesProperty } from '../property/IgnoreEqualValuesProperty';
 
 /** @internal */
 type MinimalQualifiedParameters<Ts> = Pick<
   QualifiedParameters<Ts>,
-  'unbiased' | 'timeout' | 'skipAllAfterTimeLimit' | 'interruptAfterTimeLimit'
+  'unbiased' | 'timeout' | 'skipAllAfterTimeLimit' | 'interruptAfterTimeLimit' | 'ignoreEqualValues'
 >;
 
 /** @internal */
@@ -22,5 +23,8 @@ export function decorateProperty<Ts>(
     prop = new SkipAfterProperty(prop, Date.now, qParams.skipAllAfterTimeLimit, false);
   if (qParams.interruptAfterTimeLimit != null)
     prop = new SkipAfterProperty(prop, Date.now, qParams.interruptAfterTimeLimit, true);
+  if (qParams.ignoreEqualValues) {
+    prop = new IgnoreEqualValuesProperty(prop);
+  }
   return prop;
 }
--- a/src/check/runner/configuration/Parameters.ts
+++ b/src/check/runner/configuration/Parameters.ts
@@ -92,6 +92,13 @@ export interface Parameters<T = void> {
    */
   markInterruptAsFailure?: boolean;
   /**
+   * Do not repeat runs with already covered cases.
+   * This is useful when arbitrary has a limited number of variants.
+   *
+   * NOTE: Values are compared by equality of fc.stringify results.
+   */
+  ignoreEqualValues?: boolean;
+  /**
    * Way to replay a failing property directly with the counterexample.
    * It can be fed with the counterexamplePath returned by the failing test (requires `seed` too).
    * @remarks Since 1.0.0
--- a/src/check/property/IgnoreEqualValuesProperty.ts
+++ b/src/check/property/IgnoreEqualValuesProperty.ts
@@ -1,0 +1,22 @@
+import { IRawProperty } from './IRawProperty';
+import { Random } from '../../random/generator/Random';
+import { Shrinkable } from '../arbitrary/definition/Shrinkable';
+import { stringify } from '../../utils/stringify';
+
+export class IgnoreEqualValuesProperty<Ts, IsAsync extends boolean> implements IRawProperty<Ts, IsAsync> {
+  private coveredCases: Map<string, ReturnType<IRawProperty<Ts, IsAsync>['run']>> = new Map();
+
+  constructor(readonly property: IRawProperty<Ts, IsAsync>) {}
+
+  isAsync = (): IsAsync => this.property.isAsync();
+  generate = (mrng: Random, runId?: number): Shrinkable<Ts> => this.property.generate(mrng, runId);
+  run = (v: Ts): ReturnType<IRawProperty<Ts, IsAsync>['run']> => {
+    const stringifiedValue = stringify(v);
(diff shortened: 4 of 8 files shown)


Djaler:
One question is - shouldn't we enable this flag by default?

src/check/property/SkipEqualValuesWithComparatorProperty.ts
```diff
@@ -0,0 +1,19 @@
+import { IRawProperty } from './IRawProperty';
+import { Random } from '../../random/generator/Random';
+import { Shrinkable } from '../arbitrary/definition/Shrinkable';
+
+export class SkipEqualValuesWithComparatorProperty<Ts, IsAsync extends boolean> implements IRawProperty<Ts, IsAsync> {
+  private coveredValues: Ts[] = [];
```
dubzzz, line 6: Using a `Set` would be better in terms of performances. Array structure will be too costly for very high number of runs

src/check/property/SkipEqualValuesWithComparatorProperty.ts
```diff
@@ -0,0 +1,19 @@
...
+
+  constructor(readonly property: IRawProperty<Ts, IsAsync>, readonly comparator: (a: Ts, b: Ts) => boolean) {}
+
+  isAsync = (): IsAsync => this.property.isAsync();
+  generate = (mrng: Random, runId?: number): Shrinkable<Ts> => this.property.generate(mrng, runId);
+  run = (v: Ts): ReturnType<IRawProperty<Ts, IsAsync>['run']> => {
+    if (this.coveredValues.some((coveredValue) => this.comparator(coveredValue, v))) {
+      return null as any;
```
dubzzz, line 14: Return success status instead (no `as` required if return type is OK)

src/check/runner/DecorateProperty.ts
```diff
@@ -22,5 +24,12 @@ export function decorateProperty<Ts>(
     prop = new SkipAfterProperty(prop, Date.now, qParams.skipAllAfterTimeLimit, false);
   if (qParams.interruptAfterTimeLimit != null)
     prop = new SkipAfterProperty(prop, Date.now, qParams.interruptAfterTimeLimit, true);
+  if (qParams.skipEqualValues) {
+    if (qParams.valuesComparator) {
```
dubzzz, line 28: Code could always pass in that path and let the constructor deal with defaulting the comparator if needed

src/check/runner/configuration/Parameters.ts
```diff
@@ -91,6 +91,19 @@ export interface Parameters<T = void> {
...
+   * NOTE: values comparison can be configured with 'valuesComparator'.
+   * By default values are compared by equality of fc.stringify results.
+   */
+  skipEqualValues?: boolean;
+  /**
+   * Replace the default way to compare values by a custom one
+   *
+   * Should be used with 'skipEqualValues'
+   */
+  valuesComparator?: (a: T, b: T) => boolean;
```
dubzzz, line 106: Not sure of the naming 🤔 As this parameter is useless when the other is not used I'm wondering if we could expose it differently

dubzzz (commented):
Quick review from my mobile, so I just covered it quickly. Concerning the idea to enable it by default, I think we should not. Today the feature breaks some invariants of fast-check, numRuns is not respected when you toggle it so putting it by default would be problematic. Moreover the cache plus compare process comes with an extra cost most of user will not want to pay (most of the time not needed).

Regarding the custom comparator, I'm wondering if we want to offer this option to the user. One more option in the API is one more option to maintain so at first glance I'd only offer a default comparison based on stringify. It makes finding duplicates easier as a simple Set with O(1) look-up could be used instead of a O(n) look-up running potentially costly comparisons (stringify is costly if we call it over and over, as deepEquality algorithms).

dubzzz:
Regarding the naming of the parameter, we should not call it skip. Skip has already a special meaning in fast-check. It means that the run was not executed but replaced by another.

For instance when using `fc.pre` in your property, executions leading to falsy `fc.pre` will be skipped but the number of runs you asked will always be fulfilled.

We could maybe explicit the fact that the execution will be consumed even if we do not run anything.

src/check/property/SkipEqualValuesWithComparatorProperty.ts
Djaler: image
I didn't see how to return success status in other way than null.

dubzzz: I thought back about it. Actually, the current PR not really "skip" the run (with fast-check meaning) but rather do not run the code of the property. As it does not skip it has to return a status: either success or failure, and actually it can be both of them. I'd say that the best option will be to return the returned value associated to this stringified value. We could for instance have something like:

```js
// alreadySeenValuesAndReturn: Map<string, ReturnType<IRawProperty...[run]>

const stringifiedValue = stringified(v)
if (this.alreadySeenValuesAndReturn.has(stringifiedValue))
  return this.alreadySeenValuesAndReturn.get(stringifiedValue)!;

const out = this.property.run(v);
this.alreadySeenValuesAndReturn.set(stringifiedValue, out)
return out;
```

Why that? Because with this code we could bypass a failed run (it may happen to replay some already tested values during shrinking process, some of those replays may be related to red runs, in that case in order to keep a clear error report, the run should not be marked as success and return the original version).

My snippet above is not perfect (mostly for sync property throwing stuff, the code should rethrow it to be consistent*). *the feature is like 'do not re-run user code but behave as if it was executed' 

_from my mobile_

dubzzz:
By the way, I forgot to say: thank you so much for the PR 😁

Djaler:
You're welcome ;)

dubzzz (commented):
Some minor remarks, but looks great!
I'm still looking for a naming but the one you suggested seems cool.

I suggested some and I'm still trying to think about others.

As remarks are mostly regarding the naming, let just check what would be the best choice before you update the code.

dubzzz:
@github-actions deploy

src/check/runner/configuration/Parameters.ts
```diff
@@ -91,6 +91,13 @@ export interface Parameters<T = void> {
    * @remarks Since 1.19.0
    */
   markInterruptAsFailure?: boolean;
+  /**
+   * Do not repeat runs with already covered cases.
+   * This is useful when arbitrary has a limited number of variants.
+   *
+   * NOTE: Values are compared by equality of fc.stringify results.
+   */
+  ignoreEqualValues?: boolean;
```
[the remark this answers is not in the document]
dubzzz: Actually let's go for it. Naming is good and if needed we can add more details in the JSdoc in the future.

documentation/Runners.md
```diff
@@ -1,215 +1,217 @@
-# [:house:](../README.md) Runners
```
dubzzz, line 1: Can you double check while so many lines have been marked as changed in this file? It would be easier/better to have the change limited to the impacted lines themselves in terms of git history.

It might be a change in the line ending 🤔

dubzzz (commented):
Perfect, one last comment regarding the fact that one file has changed on all lines

documentation/Runners.md
Djaler: Yeah, looks my IDE makes this automatically

Djaler:
Also, about default behavior. Can you check this? https://jqwik.net/docs/current/user-guide.html#exhaustive-generation
jqwik is a tool to do property-based testing in Java.
I tried to test the same property (with two booleans) using it, and jqwik runs it only 4 times, as I expect

dubzzz (approved):

dubzzz:
@Djaler Thanks a lot for the contribution. I'll have a look into jqwik to check how they do.

Meanwhile, I'm wondering if we should also add a skipEqualValues that really skips the runs that have already been tried (except failed ones where it should return the error), it would make users able to have the number of tests they requested 🤔

ignore would be an option to just discard the runs while skip would preserve the number of runs (potentially leading to skip-related errors if the generation of non equal values is costly).

Just an idea for the moment 💡

PR #1678, opened 2021-04-01
✨ Introduce NextArbitrary, new way to define arbitraries

dubzzz:
## In a nutshell

Defining new arbitraries from scratch in fast-check might be complex. One of the reason behind this complexity is mostly the fact that generators are responsible to generate a shrinkable value. Other frameworks tend to prefer having an API exposing two methods `generate` and `shrink`. fast-check initially went in that direction but it was impossible to shrink values issued by oneof with such approach... impossible until we came with the concept of `ArbitraryWithContextualShrink`. It somehow showed that having a dedicated `shrink` method was possible but required some more information to be passed.

`NextArbitrary`, replacement of `Arbitrary` in v3 (next major), is the new way to implement such arbitraries. Documentation will be added in coming PRs.

✔️ New feature
❌ Fix an issue
❌ Documentation improvement
❌ Other: *please explain*

(✔️: yes, ❌: no)

## Potential impacts

None


src/check/arbitrary/definition/Converters.ts
```diff
@@ -0,0 +1,57 @@
...
+    return arb.arb;
+  }
+  assertIsNextArbitrary(arb);
+  return new ConverterFromNext(arb, legacyShrunkOnceContext);
+}
+
+/**
+ * Convert an instance of Arbitrary to an instance of NextArbitrary
+ * @param arb - The instance to be converted
+ * @public
```
dubzzz, line 49:
```suggestion
 * @param arb - The instance to be converted
 * @remarks Since 2.15.0
 * @public
```

src/check/arbitrary/definition/Converters.ts
```diff
@@ -0,0 +1,57 @@
...
+  }
+  assertIsNextArbitrary(arb);
+  return new ConverterFromNext(arb);
+}
+
+/**
+ * Convert an instance of NextArbitrary to an instance of ArbitraryWithContextualShrink
+ * @param arb - The instance to be converted
+ * @param legacyShrunkOnceContext - Default context to be returned when shrunk once
+ * @public
```
dubzzz, line 24:
```suggestion
 * @param legacyShrunkOnceContext - Default context to be returned when shrunk once
 * @remarks Since 2.15.0
 * @public
```

src/check/arbitrary/definition/Converters.ts
```diff
@@ -0,0 +1,57 @@
+import { Arbitrary, assertIsArbitrary } from './Arbitrary';
+import { ArbitraryWithContextualShrink } from './ArbitraryWithContextualShrink';
+import { ConverterFromNext } from './ConverterFromNext';
+import { ConverterToNext } from './ConverterToNext';
+import { assertIsNextArbitrary, NextArbitrary } from './NextArbitrary';
+
+/**
+ * Convert an instance of NextArbitrary to an instance of Arbitrary
+ * @param arb - The instance to be converted
+ * @public
```
dubzzz, line 10:
```suggestion
 * @param arb - The instance to be converted
 * @remarks Since 2.15.0
 * @public
```

src/check/arbitrary/definition/NextArbitrary.ts
```diff
@@ -0,0 +1,387 @@
...
+   *
+   * @remarks Since 2.15.0
+   */
+  abstract generate(mrng: Random): NextValue<T>;
+
+  /**
+   * Check if a given value could have been generated
+   * through this instance of arbitrary.
+   *
+   * @param value - Value to be assessed
```
dubzzz, line 28:
```suggestion
   * through this instance of arbitrary.
   *
   * In general, `canGenerate` is not designed to be called for each `shrink` but rather on very special cases.
   * Its usage most be restricted to `canGenerate` or in the rare* context of a `shrink` method being called without
   * any context. In this ill-formed case of `shrink`, `canGenerate` could be used or called if needed.
   *
   * *we fall in that case when fast-check is asked to shrink a value that has been provided manually by the user,
   *  in other words: a value not coming from a call to `generate` or a normal `shrink`.
   *
   * @param value - Value to be assessed
```

src/check/arbitrary/definition/NextArbitrary.ts
```diff
@@ -0,0 +1,387 @@
...
+   * @returns `true` if and only if the value could have been generated by this instance
+   *
+   * @remarks Since 2.15.0
+   */
+  abstract canGenerate(value: unknown): value is T;
+
+  /**
+   * Generate a value of type `T` along with its context (if any)
+   * based on the provided random number generator.
```
dubzzz, line 37:
```suggestion
   * Shrink a value of type `T`, may rely on the context previously provided to shrink efficiently
```

src/check/arbitrary/definition/NextArbitrary.ts
```diff
@@ -0,0 +1,387 @@
...
+   *
+   * Must never be called with possibly valid values.
+   * A call to `canGenerate` must be done first if value is possibly not compatible with the current instance.
```
dubzzz, line 40:
```suggestion
   * Must never be called with possibly invalid values.
   * Consider calling `canGenerate` first if you are not sure of the origin of the value.
```

PR #1756, opened 2021-04-17
✨ Add support for bias in `clone` (and `dedup`)

dubzzz:
## In a nutshell

✔️ New feature
❌ Fix an issue
❌ Documentation improvement
❌ Other: *please explain*

(✔️: yes, ❌: no)

## Potential impacts

Impact on Generated values for `clone` and `dedup`


src/check/arbitrary/CloneArbitrary.ts
```diff
@@ -55,6 +55,10 @@ class CloneArbitrary<T> extends Arbitrary<T[]> {
       cur = its.map((it) => it.next());
     }
   }
+
+  withBias(freq: number): CloneArbitrary<T> {
+    return new CloneArbitrary(this.arb.withBias(freq), this.numValues);
```
dubzzz, line 60: In terms of performances using `biasWrapper` may have been a boost, but as we will migrate this arbitrary to `NextArbitrary` it's not needed to change that.

PR #1952, opened 2021-06-03
🚧⚡️ Reduce number for intermediate closures in generate-path

dubzzz:
Several closures were built by fast-check each time we wanted to produce a value. They may come with a price, so let's try to remove them and see what will be the resulted performance improvement.

**_Category:_**

- [ ] ✅ Add or update tests
- [ ] 🐛 Fix a bug
- [ ] 🏷️ Add or update types
- [x] ⚡️ Improve performance
- [ ] _Other(s):_ ...

**_Potential impacts:_**

- [x] Generated values: _no impact expected but as it changes the generate path it may impact anything passing through it_
- [ ] Shrink values
- [x] Performance
- [ ] Typings
- [ ] _Other(s):_ ...

### DIFF
--- a/src/check/runner/Runner.ts
+++ b/src/check/runner/Runner.ts
@@ -1,4 +1,4 @@
-import { stream } from '../../stream/Stream';
+import { Stream, stream } from '../../stream/Stream';
 import { Shrinkable } from '../arbitrary/definition/Shrinkable';
 import { PreconditionFailure } from '../precondition/PreconditionFailure';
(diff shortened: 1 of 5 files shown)


dubzzz:
@github-actions deploy

dubzzz:
It is not an option given the fact that we don't want to generate any intermediate values when running assertions against path+seed. A more complex option would be needed to handle such case properly. Additionally such 'improvement' has not prove any speed ups during benchmarks so we can close it for now.

PR #1969, opened 2021-06-07
🚀 create overload constantFrom

kazchimo:
Add constantFrom variant.
Now we can only create an arb that produces values like this.

```typescript
type Color = "white" | "black"

const arb = constantFrom<Color[]>("white", "black") // Arbitrary<Color>
```

But type parameter `Color[]` is not intuitive.
With this PR feature we can create an arb from simple type annotation.

```typescript
type Color = "white" | "black"

const arb = constantFrom<Color>("white", "black") // Arbitrary<Color>
```

> NOTE:
You cannot create union type of literal type arbs without a type annotation.
> ```typescript
> type Color = "white" | "black"
> 
> const arb = constantFrom("white", "black") // Arbitrary<string>
> ```

**_Category:_**

- [ ] ✅ Add or update tests
- [ ] 🐛 Fix a bug
- [x] 🏷️ Add or update types
- [ ] ⚡️ Improve performance
- [ ] _Other(s):_ ...

**_Potential impacts:_**

- [ ] Generated values
- [ ] Shrink values
- [ ] Performance
- [x] Typings
- [ ] _Other(s):_ ...

### DIFF
--- a/src/arbitrary/constantFrom.ts
+++ b/src/arbitrary/constantFrom.ts
@@ -16,7 +16,19 @@ type Elem<T> = T extends any[] ? T[number] : T;
  * @public
  */
 function constantFrom<T>(...values: T[]): Arbitrary<T>;
+
+/**
+ * For one `...values` values - all equiprobable
+ *
+ * **WARNING**: It expects at least one value, otherwise it should throw
+ *
+ * @param values - Constant values to be produced (all values shrink to the first one)
+ *
+ * @remarks Since 0.0.12
+ * @public
+ */
 function constantFrom<TArgs extends any[] | [any]>(...values: TArgs): Arbitrary<TArgs[number]>;
+
 function constantFrom<TArgs extends any[] | [any] | any>(...values: Arrayfy<TArgs>): Arbitrary<Elem<TArgs>> {
   if (values.length === 0) {
     throw new Error('fc.constantFrom expects at least one parameter');
--- a/src/arbitrary/constantFrom.ts
+++ b/src/arbitrary/constantFrom.ts
@@ -17,7 +17,7 @@ type Elem<T> = T extends any[] ? T[number] : T;
  * @remarks Since 0.0.12
  * @public
  */
-function constantFrom<T>(...values: T[]): Arbitrary<T>;
+function constantFrom<T = never>(...values: T[]): Arbitrary<T>;
 
 /**
  * For one `...values` values - all equiprobable
--- a/src/arbitrary/constantFrom.ts
+++ b/src/arbitrary/constantFrom.ts
@@ -2,14 +2,8 @@ import { Arbitrary } from '../check/arbitrary/definition/Arbitrary';
 import { convertFromNext } from '../check/arbitrary/definition/Converters';
 import { ConstantArbitrary } from './_internals/ConstantArbitrary';
 
-/**
- * For one `...values` values - all equiprobable
- *
- * **WARNING**: It expects at least one value, otherwise it should throw
- *
- * @param values - Constant values to be produced (all values shrink to the first one)
- */
-function constantFrom<T>(...values: T[]): Arbitrary<T>;
+type Arrayfy<T> = T extends any[] ? T : T[];
+type Elem<T> = T extends any[] ? T[number] : T;
 
 /**
  * For one `...values` values - all equiprobable
@@ -21,11 +15,13 @@ function constantFrom<T>(...values: T[]): Arbitrary<T>;
  * @remarks Since 0.0.12
  * @public
  */
-function constantFrom<TArgs extends any[] | [any]>(...values: TArgs): Arbitrary<TArgs[number]> {
+function constantFrom<T>(...values: T[]): Arbitrary<T>;
+function constantFrom<TArgs extends any[] | [any]>(...values: TArgs): Arbitrary<TArgs[number]>;
+function constantFrom<TArgs extends any[] | [any] | any>(...values: Arrayfy<TArgs>): Arbitrary<Elem<TArgs>> {
   if (values.length === 0) {
     throw new Error('fc.constantFrom expects at least one parameter');
   }
-  return convertFromNext(new ConstantArbitrary(values));
+  return convertFromNext(new ConstantArbitrary(values)) as Arbitrary<Elem<TArgs>>;
 }
 
 export { constantFrom };
--- a/src/arbitrary/constantFrom.ts
+++ b/src/arbitrary/constantFrom.ts
@@ -2,7 +2,9 @@ import { Arbitrary } from '../check/arbitrary/definition/Arbitrary';
 import { convertFromNext } from '../check/arbitrary/definition/Converters';
 import { ConstantArbitrary } from './_internals/ConstantArbitrary';
 
+/** @internal */
 type Arrayfy<T> = T extends any[] ? T : T[];
+/** @internal */
 type Elem<T> = T extends any[] ? T[number] : T;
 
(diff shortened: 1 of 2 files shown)


dubzzz:
There is actually a way to have proper typings in such case using the following syntax: `constantFrom(...["white", "black"] as const)`. But it is definitely not perfect and probably less intuitive than `constantFrom<Color>("white", "black")` for most users.

Moving from `Color[]` to `Color` would require an API breaking change and should be postponed for the next major version, but it could definitely be a good idea. Please let me know if you reach something fine. We could definitely plan to merge it for the next major release (along other breaking changes).

kazchimo:
> There is actually a way to have proper typings in such case using the following syntax: constantFrom(...["white", "black"] as const). But it is definitely not perfect and probably less intuitive than constantFrom<Color>("white", "black") for most users.

Year that's right.

Now we support both syntax of `constantFrom<Color[]>("white", "black")` and  `constantFrom<Color>("white", "black")`.
Is this cause a API breaking change?
Users still can use less intuitive syntax like `constantFrom<T[]>`.

But I'm not confident with type declarations because they have generics like `<TArgs extends any[] | [any] | any>` and use force cast with `as`.

src/arbitrary/constantFrom.ts
```diff
@@ -12,9 +15,13 @@ import { ConstantArbitrary } from './_internals/ConstantArbitrary';
  * @remarks Since 0.0.12
  * @public
  */
-export function constantFrom<TArgs extends any[] | [any]>(...values: TArgs): Arbitrary<TArgs[number]> {
+function constantFrom<T>(...values: T[]): Arbitrary<T>;
+function constantFrom<TArgs extends any[] | [any]>(...values: TArgs): Arbitrary<TArgs[number]>;
```
dubzzz, line 19: api-extractor complains about the missing JSDoc, please add the same as the one above. I'd probably keep the same `since 0.0.12` as in terms of API it keeps being the same 🤔 (so 100% the same JSDoc should work fine)

dubzzz:
I just did some manual test to ensure it did not break anything. There are some very small changes but it seems pretty backward compatible.

<details>
<summary>Here are my tests: <a href="https://www.typescriptlang.org/play?#code/PTAECEFEDEHkCVICgkBMCmBjANgQwE7qgBmArgHaYAuAlgPbmiYMDOVu5V0+dAtgIIAeACr98Acxah0ADyrpyqKRwCeAbQC6oAD6g1qjQD4AFADpzAN1zZS6FgC5QoiSwCUjgAo9eNFuhFikmrkpLwARuj4RgDcKCCg-NDCkPAoVCoADkRi+LgqxCoihqAAvE7ScgpKoKqaoAD85Y7CmrHpWaCQ2Oi8RaXlsvKKyuTqWo0tIeGRWs2xGDgERGSUtAxMrOyc3HzgRWaW1rYOTpruoF58vv7ChvNYeIQkFNT0jMzkbBxc3nvOkhUhtValpdPpRkYDqYrDY7M1Am5PN5rgEXMFQhEonc4mAAMIACUguIA0gBlFAfNg1ACM-UpWx+fH4xlcoHignI6AskWKAFpecUqAALXygBhEUURKjyfBIemgMK0sr0747XjgFlssCCCgAa3IdAA7uQ+QLQDQqBtSNhUKAMjwwrgwtgVAqiFRDehcLr0LaqHQSHR8GEaLbMNZsFJDRahXRSJbVBtPgzA-hQMKiLhUBhbSwaOJyLgqKRCBTNjUAEx0zaq7zM6kAGlAFdccvLYSrypr21+xkbzdbbeTNQAzNXk7WmcZNE21BWNKz2VNMXVdJzufhNIYh1SwmOuxOe7tpxpZ-PF9rl5Et6B+YKRVJxeapFKZWXh7gACzjr5HgRQ6c1GpU89HnLRcCkelXFbeUwm-A9f0ZdUALUICQLnDRwMgzZoPfKlcAAVh-Bk1SEK9N0hfsWx3S0wiIhCSN+DkMWvSim2o2CCM7JNELVPZyJMKjB3lXAADZiMnARmOmCit2nYCzwXGiFXEhjJP4ljZMhNDFJg9tRO4lU-w0mS5J00ClOU3AAHYJL-Zk1AAcmpRz0PPZSwlstTjOnZzXN0oA">playground</a></summary>

```ts
// BEFORE

declare function constantFromA<TArgs extends any[] | [any]>(...values: TArgs): Promise<TArgs[number]>;

// AFTER

type Arrayfy<T> = T extends any[] ? T : T[];
type Elem<T> = T extends any[] ? T[number] : T;
declare function constantFromB<T>(...values: T[]): Promise<T>;
declare function constantFromB<TArgs extends any[] | [any]>(...values: TArgs): Promise<TArgs[number]>;

// CHECKS

const a1 = constantFromA() // <never> --> this one is better
const b1 = constantFromB() // <unknown> --> it could probably be tweaked to forbid calls without any constant for the added signature

const a2 = constantFromA(1, 2)
const b2 = constantFromB(1, 2)

const a3 = constantFromA([], [2]) // <number[] | never[]>
const b3 = constantFromB([], [2]) // <number[]> --> this one is better

const a4 = constantFromA(...([[1], [2]] as const))
const b4 = constantFromB(...([[1], [2]] as const))

const a5 = constantFromA<number[]>(1, 2)
const b5 = constantFromB<number[]>(1, 2)
const b52 = constantFromB<number>(1, 2)

const a6 = constantFromA<number[][]>([1], [2])
const b6 = constantFromB<number[][]>([1], [2])
const b62 = constantFromB<number[]>([1], [2])

const a7 = constantFromA(['1'], [2])
const b7 = constantFromB(['1'], [2])
```
</details>

There is probably one case to address as previous behaviour was a bit better on this one:

```ts
const a1 = constantFromA() // current
// <never> --> this one is better

const b1 = constantFromB() // yours
// <unknown> --> it could probably be tweaked to forbid calls without any constant for the added signature
```

src/arbitrary/constantFrom.ts
```diff
@@ -2,6 +2,21 @@ import { Arbitrary } from '../check/arbitrary/definition/Arbitrary';
 import { convertFromNext } from '../check/arbitrary/definition/Converters';
 import { ConstantArbitrary } from './_internals/ConstantArbitrary';
 
+type Arrayfy<T> = T extends any[] ? T : T[];
+type Elem<T> = T extends any[] ? T[number] : T;
```
dubzzz, line 6: I usually add this annotation to avoid any risk of unwanted publication of the type. Those ones being fully internals, we can safely use it there.

```suggestion
/** @internal */
type Arrayfy<T> = T extends any[] ? T : T[];
/** @internal */
type Elem<T> = T extends any[] ? T[number] : T;
```

src/arbitrary/constantFrom.ts
```diff
@@ -2,6 +2,21 @@ import { Arbitrary } from '../check/arbitrary/definition/Arbitrary';
...
+ * For one `...values` values - all equiprobable
+ *
+ * **WARNING**: It expects at least one value, otherwise it should throw
+ *
+ * @param values - Constant values to be produced (all values shrink to the first one)
+ *
+ * @remarks Since 0.0.12
+ * @public
+ */
+function constantFrom<T>(...values: T[]): Arbitrary<T>;
```
dubzzz, line 18: I think you wanted to add `=never` to cover the last (not so) _problematic_ typing.

```suggestion
function constantFrom<T = never>(...values: T[]): Arbitrary<T>;
```

dubzzz (approved):

PR #2309, opened 2021-09-21
⚡️ Lighter implementation for `anything` arbitrary

dubzzz:
Rewrite `anything` based on `letrec` instead of `memo`. `letrec` is less expensive in terms of memory consumption and as expensive to compute than `memo`-based versions.

Snapshots have been updated as they will not 100% identical in terms of generated values. While `memo` will end with a `oneof` with a single entry, `letrec` will always deal with `oneof` with all the arbitraries. The main difference will be at the end of the tree: `letrec` will take something coming from the first arbitrary no matter the others. On his side `memo` will still call `nextInt` but with a single entry so resulting to zero (see https://github.com/dubzzz/fast-check/blob/main/src/arbitrary/_internals/FrequencyArbitrary.ts#L69).

**_Category:_**

- [ ] ✅ Add or update tests
- [ ] 🐛 Fix a bug
- [ ] 🏷️ Add or update types
- [x] ⚡️ Improve performance
- [ ] _Other(s):_ ...

**_Potential impacts:_**

- [x] Generated values
- [ ] Shrink values
- [ ] Performance
- [ ] Typings
- [ ] _Other(s):_ ...


dubzzz:
@github-actions deploy

dubzzz:
Follow performance tests at https://github.com/dubzzz/fast-check-benchmarks/actions/runs/1263626732

dubzzz:
Or an even more intense run at https://github.com/dubzzz/fast-check-benchmarks/actions/runs/1263631034

PR #2464, opened 2021-11-04
🚚 Prepare rename of `jsonObject` into `jsonValue`

### LINKED ISSUE #2457: `jsonObject` name is misleading (better name: `jsonValue`)
`jsonObject` name implies `typeof x === 'object'` when in reality it generates any JSON-representable value (`object`, `string`, `number`, `null`).

A better name for this arbitrary would be `jsonValue`.

https://github.com/dubzzz/fast-check/blob/dbf7d08b174857103d73b678fc0031f12de0e812/src/arbitrary/jsonObject.ts#L9-L19

dubzzz:
Just a new name for `jsonObject` and `unicodeJsonObject` as suggested into #2457.

Fixes #2457

**_Category:_**

- [ ] ✅ Add or update tests
- [ ] 🐛 Fix a bug
- [ ] 🏷️ Add or update types
- [ ] ⚡️ Improve performance
- [ ] _Other(s):_ ...

**_Potential impacts:_**

- [ ] Generated values
- [ ] Shrink values
- [ ] Performance
- [ ] Typings
- [ ] _Other(s):_ ...


dubzzz:
Before merging it we probably want to see what we want to do with `maxDepth`. Do we keep defaulting it to 2 or do we leverage the `depthFactor` on it (see oneof, frequency)?

PR #2515, opened 2021-11-29
💥 Switch to the new API of `Arbitrary` and remove old variants

dubzzz:
This was the aim of this whole introduction of `NextArbitrary`: being able to move to a better API for our `Arbitrary`. Now it's true!

Still a lot to renames required in order to have everything ok for v3 but we are closer from the final target.

**_Category:_**

- [ ] ✅ Add or update tests
- [ ] 🐛 Fix a bug
- [ ] 🏷️ Add or update types
- [ ] ⚡️ Improve performance
- [x] _Other(s):_ ...

**_Potential impacts:_**

- [ ] Generated values
- [ ] Shrink values
- [ ] Performance
- [ ] Typings
- [ ] _Other(s):_ ...


src/check/property/AsyncProperty.ts
```diff
@@ -23,9 +22,8 @@ function asyncProperty<Ts extends [unknown, ...unknown[]]>(
   }
   const arbs = args.slice(0, args.length - 1) as { [K in keyof Ts]: Arbitrary<Ts[K]> };
   const p = args[args.length - 1] as (...args: Ts) => Promise<boolean | void>;
-  const mappedArbs = arbs.map((arb): typeof arb =>
-    convertFromNext(new AlwaysShrinkableArbitrary(convertToNext(arb)))
-  ) as typeof arbs;
+  arbs.forEach(assertIsArbitrary);
```
dubzzz, line 25: Not sure we want to assert from that place 🤔
We used to assert from `tuple` but I don't know if it was better or not

test/unit/check/arbitrary/ArrayArbitrary.spec.ts
```diff
@@ -1,71 +0,0 @@
-import * as fc from '../../../../lib/fast-check';
```
dubzzz, line 1: Migrate this test

test/unit/check/model/commands/CommandsArbitrary.spec.ts
```diff
@@ -1,270 +0,0 @@
-import prand from 'pure-rand';
```
dubzzz, line 1: Migrate this test

test/unit/stubs/arbitraries.ts
```diff
@@ -1,7 +1,7 @@
 import { Arbitrary } from '../../../src/check/arbitrary/definition/Arbitrary';
```
dubzzz, line 1: Migrate users to mocks or local stubs

src/check/property/AsyncProperty.ts
dubzzz: Let's keep the check here for the moment!

PR #2600, opened 2022-01-02
⚡️ Improve performance of `set`

dubzzz:
Only applies for instances of `set` not overriding the default comparator. Any override of the compator will lead to degradated performances. The performance gap will mostly be visible with large arrays.

**_Category:_**

- [ ] ✅ Add or update tests
- [ ] 🐛 Fix a bug
- [ ] 🏷️ Add or update types
- [x] ⚡️ Improve performance
- [ ] _Other(s):_ ...

**_Potential impacts:_**

- [x] Generated values: Not expected, but code forks between no user-defined comparison function (with new code) and the other, we may have unwantingly introduced a regression there.
- [ ] Shrink values
- [x] Performance
- [ ] Typings
- [ ] _Other(s):_ ...

### DIFF
--- a/src/arbitrary/_internals/helpers/StrictlyEqualSet.ts
+++ b/src/arbitrary/_internals/helpers/StrictlyEqualSet.ts
@@ -1,0 +1,45 @@
+import { CustomSet } from '../interfaces/CustomSet';
+
+/**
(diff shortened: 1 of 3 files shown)


dubzzz:
please deploy

PR #2750, opened 2022-02-22
📝 Reorder assert overloads to render better documentation

### LINKED ISSUE #2749: The signature and async behavior of assert should be clarified
##  Bug Report

https://dubzzz.github.io/fast-check/index.html#assert says that `assert` returns a promise and "Has to be awaited", but [actual behavior](https://github.com/dubzzz/fast-check/blob/350d517/src/check/runner/Runner.ts#L196-L200) seems to be more nuanced—it returns a promise only if the property being tested is async, but otherwise finishes synchronously and returns undefined.

### Synchronous
```sh
$ node -e 'const fc = require("fast-check");                                                       
  let count = 0;
  const result = fc.assert(
    fc.property(fc.nat(), n => { count++ })
  );

  console.log(
    `${count} invocations before fc.assert returned ${result}`
  );
  Promise.resolve().then(
    () => console.log(`${count} invocations at end of turn`)
  );

  const then = result && result.then;
  if (typeof then === "function") {
    then.call(result,
      v => console.log(`${count} invocations at resolution`),
      err => console.error(err)
    );
  }'
100 invocations before fc.assert returned undefined
100 invocations at end of turn
```

### Asynchronous
```sh
$ node -e 'const fc = require("fast-check");
  let count = 0;
  const result = fc.assert(
    fc.asyncProperty(fc.nat(), n => { count++ })
  );

  console.log(
    `${count} invocations before fc.assert returned ${result}`
  );
  Promise.resolve().then(
    () => console.log(`${count} invocations at end of turn`)
  );

  const then = result && result.then;
  if (typeof then === "function") {
    then.call(result,
      v => console.log(`${count} invocations at resolution`),
      err => console.error(err)
    );
  }'
0 invocations before fc.assert returned [object Promise]
1 invocations at end of turn
100 invocations at resolution
```

## Expected behavior

Documentation should accurately describe behavior, which should be possible by editing comments in [Runner.ts](https://github.com/dubzzz/fast-check/blob/350d517/src/check/runner/Runner.ts#L153-L194) (and possibly just reordering, documentation appears to be derived from the first signature).

## Your environment

| fast-check            | 2.22.0     |
| node                  | 16.10.0    |

gibson042:
Fixes #2749

**_Category:_**

- [ ]  Introduce new features
- [x]  Add or update documentation
- [ ]  Add or update tests
- [ ]  Fix a bug
- [ ]  Add or update types
- [ ]  Improve performance
- [ ] _Other(s):_ ...

**_Potential impacts:_**

- [ ] Generated values
- [ ] Shrink values
- [ ] Performance
- [ ] Typings
- [x] _Other(s):_ [API reference documentation](https://dubzzz.github.io/fast-check/index.html#assert)

### DIFF
--- a/src/check/runner/Runner.ts
+++ b/src/check/runner/Runner.ts
@@ -156,24 +156,6 @@ function check<Ts>(rawProperty: IRawProperty<Ts>, params?: Parameters<Ts>): unkn
  * It can be called directly from describe/it blocks of Mocha.
  * No meaningful results are produced in case of success.
  *
- * WARNING: Returns a promise to be awaited if the property is asynchronous
- *
- * @param property - Synchronous or asynchronous property to be checked
- * @param params - Optional parameters to customize the execution
- *
- * @remarks Since 0.0.7
- * @public
- */
-function assert<Ts>(
-  property: IRawProperty<Ts> & Exclude<IRawProperty<Ts>, IAsyncProperty<Ts> | IProperty<Ts>>,
-  params?: Parameters<Ts>
-): Promise<void> | void;
-/**
- * Run the property, throw in case of failure
- *
- * It can be called directly from describe/it blocks of Mocha.
- * No meaningful results are produced in case of success.
- *
  * WARNING: Has to be awaited
  *
  * @param property - Asynchronous property to be checked
@@ -196,6 +178,21 @@ function assert<Ts>(property: IAsyncProperty<Ts>, params?: Parameters<Ts>): Prom
  * @public
  */
 function assert<Ts>(property: IProperty<Ts>, params?: Parameters<Ts>): void;
+/**
+ * Run the property, throw in case of failure
+ *
+ * It can be called directly from describe/it blocks of Mocha.
+ * No meaningful results are produced in case of success.
+ *
+ * WARNING: Returns a promise to be awaited if the property is asynchronous
+ *
+ * @param property - Synchronous or asynchronous property to be checked
+ * @param params - Optional parameters to customize the execution
+ *
+ * @remarks Since 0.0.7
+ * @public
+ */
+function assert<Ts>(property: IRawProperty<Ts>, params?: Parameters<Ts>): Promise<void> | void;
 function assert<Ts>(property: IRawProperty<Ts>, params?: Parameters<Ts>): unknown {
   const out = check(property, params);
   if (property.isAsync()) return (out as Promise<RunDetails<Ts>>).then(asyncReportRunDetails);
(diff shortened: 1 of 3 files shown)


dubzzz:
Seems that the change introduces a breaking change 🤔

gibson042:
@dubzzz Fixed (:crossed_fingers:)

src/check/runner/Runner.ts
```diff
@@ -154,45 +154,48 @@ function check<Ts>(rawProperty: IRawProperty<Ts>, params?: Parameters<Ts>): unkn
...
  * @param params - Optional parameters to customize the execution
  *
  * @remarks Since 0.0.7
  * @public
  */
-function assert<Ts>(property: IAsyncProperty<Ts>, params?: Parameters<Ts>): Promise<void>;
+function assert<Ts>(
+  property: IRawProperty<Ts> & Exclude<IRawProperty<Ts>, IAsyncProperty<Ts> | IProperty<Ts>>,
+  params?: Parameters<Ts>
+): Promise<void> | void;
```
dubzzz, line 170: I'm not sure to understand the need behind this move of the overload higher in the ordering. It seems that with the current setup (the one before your PR) the overloads gets properly shown in IDEs like VSCode.

Here is what I get for `asyncProperty`:
image

And for `property`:
image

Your problem mostly be related to things being written in the current documentation and not with the ordering of those overloads.

Additionally as it worked well up to now, and to limit risks of regressions I'd not add complexity (Exclude... were not needed in the past, they make typings harder to read for non-ts users) on the typings for not needed reasons.

dubzzz (changes requested):
Not sure to get the reason of the PR. In the documentation only the `asyncProperty` version is marked as "WARNING: Has to be awaited". The sync version does not have the comment and is the one shown in the IDE.

Both are generated in the documentation because they have to be. So yes the message "WARNING: Has to be awaited" is shown but only for the signature with async which is expected.

gibson042:
Oh, I didn't realize that the signatures were independently selectable. This is more like an API reference page usability issue, but I'm not sure where to fix it. In the meantime, I've updated this PR to restore the current overload signature order.

gibson042:
@dubzzz Addressed with custom CSS.

**light mode**
light

**dark mode**
dark

dubzzz:
Will check the PR in the coming days, sorry for the delay. Coming back soon

dubzzz (approved):

dubzzz:
Sorry for the delay, once again thank for the update plus the dark mode. I just relaunched the tests just in case (mostly for prettier, tsdoc and other scripts that may complain for things like comments...).

PR #2803, opened 2022-03-17
🐛🔧 Add `"types"` to `"exports"`

Osman-Sodefa:
When trying to use `fast-check` with TypeScript where  `moduleResolution` is set to `Node12` or `NodeNext` we get the following error:

> Could not find a declaration file for module 'fast-check'. '.../node_modules/fast-check/lib/esm/fast-check.js' implicitly has an 'any' type.
> Try `npm i --save-dev @types/fast-check` if it exists or add a new declaration (.d.ts) file containing `declare module 'fast-check';`ts(7016)

This change fixes the issue.

I think this might be related: https://github.com/microsoft/TypeScript/issues/46334

**_Category:_**

- [ ] ✅ Add or update tests
- [x] 🐛 Fix a bug
- [] 🏷️ Add or update types
- [ ] ⚡️ Improve performance
- [ ] _Other(s):_ ...

**_Potential impacts:_**

- [ ] Generated values
- [ ] Shrink values
- [ ] Performance
- [ ] Typings
- [ ] _Other(s):_ ...

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
   "exports": {
     ".": {
       "require": "./lib/fast-check.js",
+      "types": "./lib/types/fast-check.d.ts",
       "default": "./lib/esm/fast-check.js"
     }
   },


dubzzz:
Thank you so much for the fix @Osman-Sodefa! I'll probably need to cover it as I did for other esm-related stuff in https://github.com/dubzzz/fast-check/tree/main/test/esm.

Is there a specific version you need the fix in? I mean which version do you need to include the fix?

dubzzz:
2.23.1 is on its way 🚀

Osman-Sodefa:
> I'll probably need to cover it as I did for other esm-related stuff in https://github.com/dubzzz/fast-check/tree/main/test/esm.

Though I think it's more TS related than ESM related.

dubzzz:
I mean I put all the environment related issues into this directory. It's probably not well labeled 😂

Osman-Sodefa:
I have enabled `skipLibCheck` in my `tsconfig.json`; that's why I didn't get that error.

Issue #2780, opened 2022-03-11
How do I test concurrency for an API?


younes-io:
## 💬 Question and Help

I am developing a booking app and I want to use the [`fast-check`][1] library to simulate race conditions in my API. Basically, I want to simulate the use case where many users would call the same API to book the same table at the same time :

Suppose this is my API : `POST /book` with a request body like :

```json

{
   "table": 1,
   "time": "20:00"
}

```

My [`koa`][2] API looks something like this:

```ts

router
    .post('/book', (ctx) => {
        const { table, time } = ctx.request.body;
        const { isBooked } = await checkIfAvailable(table,time).body; // calls MongoDB
        if(!isBooked){
			ctx.body = {
				table,
				time,
				booked: true
			};	
        } else {
			ctx.throw("This table is already booked", 404)
		}
        
    })


```

My test looks like this :

```ts

mockedCheckIfAvailable.mockImplementation(
	s.scheduleFunction(async (table,time) => {
		return  booking;
	}),
);

const  res  =  await  request
	.post('/book')
	.set('Content-Type', 'application/json')
	.set('Accept', 'application/json')
	.send(booking);
expect(res).toBe(true);

```

I read the [Race Conditions][3] docs but I can't get around how to apply it.

  [1]: https://github.com/dubzzz/fast-check
  [2]: https://github.com/koajs/koa
  [3]: https://github.com/dubzzz/fast-check/blob/main/documentation/RaceConditions.md

dubzzz:
Hey,

You probably want to have a look into those examples on race conditions: https://github.com/dubzzz/fast-check/tree/main/example/005-race

They should be useful to see how it can be applied onto "real" examples. Just in case it can help, I also gave a talk at React Europe to introduce the technique: https://www.youtube.com/watch?v=GIPbY75-lEo.

younes-io:
Good evening @dubzzz ,

Thank you for the link. I have already watched that video, and I also read most of the docs of the library.
The weird part is when I put the API call inside the `fc.asyncProperty()`, the test passes even though my assertion is `expect(s.report()).toBe(5);`... it always passes
When I put the API call out of the `fc.asyncProperty()` block, the API call is made, and the test fails (which is what I expect).

So, my question is: does `fast-check` have an impact on API calls, mocked servers ?

By the way, this how I mock the server:

```ts
import supertest from 'supertest';
import app from '../src/app';

const request = supertest(app.callback());

await request
        .post('/api/v1/book')
[...]

PR #2807, opened 2022-03-17
✨ Add ability to wait until the end of a given task from `fc.scheduler`

dubzzz:
Related to #2780

While the current implementation of `fc.scheduler` is offering all the minimal requirements to check any asynchronous piece of software, it can be difficult to use against test solutions like `supertest` that do not trigger the API request immediately and rather start them into some kind of differed trigger. For those kind of behaviours, the recommended approach (before this PR), would be to manually make sure that the API request has been issued before calling any of the `waitOne` or `waitAll`. While it works, it makes the test more difficult to read and rather complex to get as it implies retry logics and `Promise`-based tricks to wait enough and see the requests.

The current pull request adds a new built-in that should highly ease such needs by providing a `waitFor` able to wait until the full resolution of a task while triggering scheduled tasks that may occur later or on the go. It is not as safe as the usual `waitAll` and `waitOne` as it depends on async operations possibly depending on some random (or more precisely something not under the control of fast-check) code.

**_Category:_**

- [ ] ✅ Add or update tests
- [ ] 🐛 Fix a bug
- [ ] 🏷️ Add or update types
- [ ] ⚡️ Improve performance
- [ ] _Other(s):_ ...

**_Potential impacts:_**

- [ ] Generated values
- [ ] Shrink values
- [ ] Performance
- [ ] Typings
- [ ] _Other(s):_ ...


dubzzz:
please deploy

younes-io:
Thank you @dubzzz !

PR #2813, opened 2022-03-25
📝 Add examples based on supertest

dubzzz:
**_Category:_**

- [x] 📝 Add or update documentation
- [ ] ✅ Add or update tests
- [ ] 🐛 Fix a bug
- [ ] 🏷️ Add or update types
- [ ] ⚡️ Improve performance
- [ ] _Other(s):_ ...

**_Potential impacts:_**

- [ ] Generated values
- [ ] Shrink values
- [ ] Performance
- [ ] Typings
- [ ] _Other(s):_ ...

### DIFF
--- a/example/005-race/supertest/app.spec.ts
+++ b/example/005-race/supertest/app.spec.ts
@@ -1,0 +1,119 @@
+import fc from 'fast-check';
+import request from 'supertest';
+
+import { app, dropDeactivatedInternal } from './src/app';
+//import { app, dropDeactivatedInternal } from './src/appBug';
+import * as DbMock from './src/db';
+import { User } from './src/db';
+
+if (!fc.readConfigureGlobal()) {
+  // Global config of Jest has been ignored, we will have a timeout after 5000ms
+  // (CodeSandbox falls in this category)
+  fc.configureGlobal({ interruptAfterTimeLimit: 4000 });
+}
+
+describe('app', () => {
+  it('should be able to call multiple /drop-deactivated at the same time', async () => {
+    await fc.assert(
+      fc.asyncProperty(
+        fc.uniqueArray(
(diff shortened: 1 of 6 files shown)


example/005-race/supertest/app.spec.ts
```diff
@@ -0,0 +1,119 @@
...
+          { selector: (u) => u.id }
+        ),
+        fc.integer({ min: 1, max: 5 }),
+        fc.scheduler({
+          // In the case of supertest, as a user you MUST define a custom `act` function
+          // if you want to run multiple queries at the same time. The following one adds
+          // some extra timers so that supertest can push promises in the queue in time.
+          // Second test is the very same test but outside of supertest.
+          act: async (f) => {
+            await new Promise((r) => setTimeout(r, 0));
```
dubzzz, line 34: While it works well for this precise setup I'm wondering if it is not a bit lucky. What seems to happen (to be confirmed) is that the first query is triggered and starts its call to the db. At this point, if no delay based on timeouts is done as we do with this act, then the first db call for query 1 (alone when scheduled) will resolve and the next db call linked to quey 1 will go next. Once query 1, will be over query 2 will start. With the delay based on timeout we somehow allow the second query to start, but it's already too late as it will never compete with query 1 on the first db access. There is probably something to do there to allow user to customize kind of delays they want to apply at scheduling time. 🤔

dubzzz: I'm wondering if a way to mimic supertest would somehow be to run the call with some kind of delays.

dubzzz: Act is wrapping waitOne and related before they even starts to look for pending requests, so the act-based solution should be the way to go for supertest 👌

PR #2826, opened 2022-03-30
🐛 Narrow return type of readConfigureGlobal

### LINKED ISSUE #2825: `readConfigureGlobal()` is typed as being able to return undefined
## 🐛 Bug Report

It looks like [`readConfigureGlobal()` is typed as returning `GlobalParameters | undefined`](https://github.com/dubzzz/fast-check/blob/main/src/check/runner/configuration/GlobalParameters.ts#L118), but from what I can tell it's not possible for the global parameters to be undefined. The [parameters start out as an empty object](https://github.com/dubzzz/fast-check/blob/main/src/check/runner/configuration/GlobalParameters.ts#L5) and can only be set to a [non-undefined `GlobalParameters`](https://github.com/dubzzz/fast-check/blob/main/src/check/runner/configuration/GlobalParameters.ts#L109) or [reset back to an empty object](https://github.com/dubzzz/fast-check/blob/main/src/check/runner/configuration/GlobalParameters.ts#L128).

## Expected behavior

I would expect the return type of `readConfigureGlobal()` to be `GlobalParameters`.

TomerAberbach:
Closes #2825

### DIFF
--- a/src/check/runner/configuration/GlobalParameters.ts
+++ b/src/check/runner/configuration/GlobalParameters.ts
@@ -115,7 +115,7 @@ export function configureGlobal(parameters: GlobalParameters): void {
  * @remarks Since 1.18.0
  * @public
  */
-export function readConfigureGlobal(): GlobalParameters | undefined {
+export function readConfigureGlobal(): GlobalParameters {
   return globalParameters;
 }
 


dubzzz:
Thanks a lot for the fix. I'll wait for test to run and I'll probably launch a new release (minor) to include it along with other changes.

Issue #2736, opened 2022-02-18
Sampling a property throws an error inside fast-check: "this.arb.withBias is not a function"


nth-commit:
## 🐛 Bug Report

I want to be able to change the assert function to a sample and see what values might be input into my test. 

However, when I sample a property, I get the following error:

```
    TypeError: this.arb.withBias is not a function

      45 |
      46 | test('reproduction', () => {
    > 47 |   fc.sample(fc.property(fc.integer(), () => {})).forEach(console.log)
         |      ^
      48 | })
      49 |

      at ConverterToNext.generate (node_modules/fast-check/lib/check/arbitrary/definition/ConverterToNext.js:33:55)
      at Property.generate (node_modules/fast-check/lib/check/property/Property.generic.js:28:32)
      at node_modules/fast-check/lib/check/runner/Tosser.js:9:28
      at node_modules/fast-check/lib/check/runner/Sampler.js:26:62
      at mapHelper (node_modules/fast-check/lib/stream/StreamHelpers.js:19:15)
          at mapHelper.next (<anonymous>)
      at Object.sample (node_modules/fast-check/lib/check/runner/Sampler.js:33:16)
      at Object.<anonymous> (test/connectDisconnectModel.test.ts:47:6)
```

## To Reproduce

`fc.sample(fc.property(fc.integer(), () => {})).forEach(console.log)`

## Expected behavior

Doesn't throw, prints out all the delicious generated parameters.

## Your environment

| Packages / Softwares  | Version(s) |
| --------------------- | ---------- |
| fast-check            |  2.22.0          |
| node                  | 16.3.0           |
| TypeScript*           |  4.5.5          |

*Only for TypeScript's users

nth-commit:
This seems to be a workaround:

```typescript
const property = fc.property(fc.integer(), () => {})
// @ts-ignore
property.withBias = () => property

fc.sample(property, {}).forEach(console.log)
```

So either:
- Properties needs a no-op implementation of `withBias`
- Properties dispatches any `withBias` calls to all the input generators
- `ConverterToNext.generate` checks for the existence of a `withBias` method before trying to call it

dubzzz:
Woups this is indeed not expected. Good catch 😱

dubzzz:
Still haven't checked it further, but it's in my radar

dubzzz:
Will be fixed with v3

dubzzz:
Fixed in v3

PR #2942, opened 2022-05-06
💥 Change the API of a property to rely on the modern one

dubzzz:
This PR is a revamped version of #2498 which targetted the alpha version of v3.

The aim is to replace our legacy API for `Property` and `AsyncProperty` by the one created for our next generation of arbitraries. The PR simplify replace the legacy `Property` by its new version previously called `NextProperty`.

**_Category:_**

- [ ] ✅ Add or update tests
- [ ] 🐛 Fix a bug
- [ ] 🏷️ Add or update types
- [ ] ⚡️ Improve performance
- [x] _Other(s):_ Breaking change
  
  

**_Potential impacts:_**

- [ ] Generated values
- [ ] Shrink values
- [ ] Performance
- [ ] Typings
- [x] _Other(s):_ API change for property type (most of the time not touched by the end-user but as the API is public nothing prevents users from using it and defining their own wrappers)


src/check/runner/Sampler.ts
```diff
@@ -16,10 +14,10 @@ import { pathWalk } from './utils/PathWalker';
 function toProperty<Ts>(
   generator: IRawProperty<Ts> | Arbitrary<Ts>,
   qParams: QualifiedParameters<Ts>
-): INextRawProperty<Ts> {
+): IRawProperty<Ts> {
   const prop = !Object.prototype.hasOwnProperty.call(generator, 'isAsync')
     ? new Property(generator as Arbitrary<Ts>, () => true)
-    : convertToNextProperty(generator as IRawProperty<Ts>);
+    : (generator as IRawProperty<Ts>);
```
dubzzz, line 20: This line is probably the reason of the issue #2736 reported by @nth-commit. I'll either patch it in v3 and v2 or just v3 🤔 But the `as` is definitely the source of the issue.

PR #2968, opened 2022-05-20
🏷️ Better typings for `letrec`

dubzzz:
As suggested in issue #1164 and then in the PR #1572, typings of `letrec` could have been improved a lot. And this is basically the idea of this PR that was highly inspired by #1572 and fixed the few last remaining issues.

**_Category:_**

- [ ] ✅ Add or update tests
- [ ] 🐛 Fix a bug
- [x] 🏷️ Add or update types
- [ ] ⚡️ Improve performance
- [ ] _Other(s):_ ...
  
  

**_Potential impacts:_**

- [ ] Generated values
- [ ] Shrink values
- [ ] Performance
- [x] Typings
- [ ] _Other(s):_ ...

### DIFF
--- a/src/arbitrary/letrec.ts
+++ b/src/arbitrary/letrec.ts
@@ -70,7 +70,7 @@ export type LetrecLooselyTypedBuilder<T> = (tie: LetrecLooselyTypedTie) => Letre
  * @public
  */
 // eslint-disable-next-line @typescript-eslint/ban-types
-export function letrec<T>(builder: T extends {} ? LetrecTypedBuilder<T> : never): LetrecValue<T>;
+export function letrec<T>(builder: T extends Record<string, unknown> ? LetrecTypedBuilder<T> : never): LetrecValue<T>;
 /**
  * For mutually recursive types
  *
--- a/src/arbitrary/letrec.ts
+++ b/src/arbitrary/letrec.ts
@@ -17,7 +17,10 @@ export type LetrecValue<T> = {
  * @remarks Since 3.0.0
  * @public
  */
-export type LetrecTypedTie<T> = <K extends keyof T>(key: K) => Arbitrary<T[K]>;
+export interface LetrecTypedTie<T> {
+  <K extends keyof T>(key: K): Arbitrary<T[K]>;
+  (key: string): Arbitrary<unknown>;
+}
 /**
  * Strongly typed type for the `builder` function passed to {@link letrec}.
  * You may want also want to use its loosely typed version {@link LetrecLooselyTypedBuilder}.
--- a/src/arbitrary/letrec.ts
+++ b/src/arbitrary/letrec.ts
@@ -69,7 +69,6 @@ export type LetrecLooselyTypedBuilder<T> = (tie: LetrecLooselyTypedTie) => Letre
  * @remarks Since 1.16.0
  * @public
  */
-// eslint-disable-next-line @typescript-eslint/ban-types
 export function letrec<T>(builder: T extends Record<string, unknown> ? LetrecTypedBuilder<T> : never): LetrecValue<T>;
 /**
  * For mutually recursive types
--- a/src/arbitrary/letrec.ts
+++ b/src/arbitrary/letrec.ts
@@ -2,6 +2,52 @@ import { LazyArbitrary } from './_internals/LazyArbitrary';
 import { Arbitrary } from '../check/arbitrary/definition/Arbitrary';
 
 /**
+ * Type of the value produced by {@link letrec}
+ * @remarks Since 3.0.0
+ * @public
+ */
+export type LetrecValue<T> = {
+  [K in keyof T]: Arbitrary<T[K]>;
+};
+
+/**
+ * Strongly typed type for the `tie` function passed by {@link letrec} to the `builder` function we pass to it.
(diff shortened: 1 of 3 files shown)


dubzzz:
Ping @qlonik

dubzzz:
That's pretty surprising:

```ts
const { b, ..._unused } = fc.letrec((tie) => ({
  a: fc.integer(),
  b: fc.array(tie('c')),
  c: fc.tuple(tie('d'), tie('a')),
  d: fc.string(),
}));
```

⬆️ Compiles!

```ts
const { b } = fc.letrec((tie) => ({
  a: fc.integer(),
  b: fc.array(tie('c')),
  c: fc.tuple(tie('d'), tie('a')),
  d: fc.string(),
}));
```

⬆️ Does not compile!

qlonik:
First of all, thank you for keeping track of this feature. 😄 


> That's pretty surprising:
> 
> ```ts
> const { b, ..._unused } = fc.letrec((tie) => ({
>   a: fc.integer(),
>   b: fc.array(tie('c')),
>   c: fc.tuple(tie('d'), tie('a')),
>   d: fc.string(),
> }));
> ```
> 
> ⬆️ Compiles!
> 
> ```ts
> const { b } = fc.letrec((tie) => ({
>   a: fc.integer(),
>   b: fc.array(tie('c')),
>   c: fc.tuple(tie('d'), tie('a')),
>   d: fc.string(),
> }));
> ```
> 
> ⬆️ Does not compile!

I'm very surprised with this behavior. For me both of these seem to not produce any type error in code sandbox environment. But somehow in this case where we are using untyped variant of letrec, the binding variable on the left of equal sign impacts the inferred type in the function call on the right. I haven't seen anything like this before.

```ts
  const { a, ..._unused } = fc.letrec((tie) => ({
    a: fc.integer(),
    b: fc.array(tie('c')),
    c: fc.tuple(tie('d'), tie('a')),
    d: fc.string(),
  }));
// letrec type:
// letrec<{ [x: string]: ...; a: unknown; }>(builder: fc.LetrecTypedBuilder<{ [x: string]: ...; a: unknown; }>): fc.LetrecValue<{ [x: string]: ...; a: unknown; }>

  const { b } = fc.letrec((tie) => ({
    a: fc.integer(),
    b: fc.array(tie('c')),
    c: fc.tuple(tie('d'), tie('a')),
    d: fc.string(),
  }));
// letrec type:
// letrec<{ b: unknown; }>(builder: fc.LetrecTypedBuilder<{ b: unknown; }>): fc.LetrecValue<{ b: unknown; }>

  const c = fc.letrec((tie) => ({
    a: fc.integer(),
    b: fc.array(tie('c')),
    c: fc.tuple(tie('d'), tie('a')),
    d: fc.string(),
  }));
// letrec type
/*
letrec<{
    a: number;
    b: unknown[];
    c: [unknown, unknown];
    d: string;
}>(builder: fc.LetrecLooselyTypedBuilder<{
    a: number;
    b: unknown[];
    c: [unknown, unknown];
    d: string;
}>): fc.LetrecValue<{
    a: number;
    b: unknown[];
    c: [unknown, unknown];
    d: string;
}>
*/
```

When I click go to definition of `letrec` function, for the first two use cases, TS jumps to "typed" variant of letrec, but the third one jumps to "loosely" typed variant. Somehow TS gets confused between two variants.

qlonik:
I don't think there is a way in TS to distinguish between whether generic was specified explicitly by the user or if that generic was inferred by TS itself. So I'm not sure if having these two overloads:

```ts
export function letrec<T>(builder: T extends {} ? LetrecTypedBuilder<T> : never): LetrecValue<T>;
export function letrec<T>(builder: LetrecLooselyTypedBuilder<T>): LetrecValue<T>;
```

provides enough information for compiler to distinguish between the two variants. Do you think providing just one signature like this would be sufficient?

```ts
export function letrec<T extends Record<string, unknown>>(builder: LetrecTypedBuilder<T>): LetrecValue<T>;
```

Given `LetrecTypedBuilder` includes the case `(key: string): Arbitrary<unknown>`, the `T` generic should get inferred into an object with returned keys and unknown values in the case when generic is not explicitly specified. (I'm sorry, I'm not on my computer right now, so I cannot yet verify if that works or not)

dubzzz:
I did think like you concerning the idea to merge both overloads together, tried but it failed 😔

My last version deals with most of the case and seems not to break existing usages of letrec, so I'll probably go for it for a first try.

dubzzz:
Here is the fix that made the 'does not compile' compiling: https://github.com/dubzzz/fast-check/commit/8fcbee325fd22ed2ee7c5f7ce17a1a416bd0058a

src/arbitrary/letrec.ts
```diff
@@ -1,6 +1,76 @@
...
+ * // tree is 50% of node, 50% of leaf
+ * // the ratio goes in favor of leaves as we go deeper in the tree (thanks to depthFactor)
+ * ```
+ *
+ * @param builder - Arbitraries builder based on themselves (through `tie`)
+ *
+ * @remarks Since 1.16.0
+ * @public
+ */
+// eslint-disable-next-line @typescript-eslint/ban-types
```
dubzzz, line 72:
```suggestion
```

dubzzz:
please format

PR #2976, opened 2022-05-25
🔖 Update CHANGELOG.md for 3.0.0

dubzzz:
Here is the CHANGELOG for 3.0.0... and the migration guide linked to it.

---

Here are the stats of how v3 performs compared to v2 in terms of time to generate entries: SAME.

Full logs here: https://github.com/dubzzz/fast-check-benchmarks/runs/6597467250 (runs is buggy for tests tweaking the maxLength)


dubzzz:
please deploy

PR #2989, opened 2022-06-06
👷 Make `fast-check` self reference itself as a dev dependency

dubzzz:
Leverage workspaces introduced in #2986 to stop relying on `../../lib` to import the build version of fast-check. Reference it as any other package.

Also includes #2998.

**_Category:_**

- [ ] ✅ Add or update tests
- [ ] 🐛 Fix a bug
- [ ] 🏷️ Add or update types
- [ ] ⚡️ Improve performance
- [ ] _Other(s):_ ...
  
  

**_Potential impacts:_**

- [ ] Generated values
- [ ] Shrink values
- [ ] Performance
- [ ] Typings
- [ ] _Other(s):_ ...


dubzzz:
Seems that in order to merge this PR we would need some more changes to be done first. For instance build should be build:prod, but in order to do that we should ideally drop the post-scripts or at least par of them.

PR #3012, opened 2022-06-11
Fix spelling

icylace:


dubzzz:
Thanks for the fix of the typo.

You may need to run `yarn version:bump` locally to declare a patch on fast-check and decline all others and commit the generated file.

You may also rename the PR to better follow the gitmoji convention we follow in the repository by adding the emoji 📝 and giving more context on what the typo is about.

Thanks in advance the upcoming updates and for the fix 👍

PR #3051, opened 2022-07-09
🐛 Better rounding in `statistics`

dubzzz:
Existing rounding strategy was implementation dependant as toFixed does not really comes with an explicit rounding rule.

**_Category:_**

- [ ] ✅ Add or update tests
- [x] 🐛 Fix a bug
- [ ] 🏷️ Add or update types
- [ ] ⚡️ Improve performance
- [ ] _Other(s):_ ...
  
  

**_Potential impacts:_**

- [ ] Generated values
- [ ] Shrink values
- [ ] Performance
- [ ] Typings
- [ ] _Other(s):_ ...

### DIFF
--- a/packages/fast-check/src/check/runner/Sampler.ts
+++ b/packages/fast-check/src/check/runner/Sampler.ts
@@ -69,6 +69,11 @@ function sample<Ts>(generator: IRawProperty<Ts> | Arbitrary<Ts>, params?: Parame
   return [...streamSample(generator, params)];
 }
 
+/** @internal */
+function round2(n: number): number {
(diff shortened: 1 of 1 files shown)


packages/fast-check/src/check/runner/Sampler.ts
```diff
@@ -69,6 +69,11 @@ function sample<Ts>(generator: IRawProperty<Ts> | Arbitrary<Ts>, params?: Parame
   return [...streamSample(generator, params)];
 }
 
+/** @internal */
+function round2(n: number): number {
```
dubzzz, line 73:
```suggestion
function round2(n: number): string {
```

PR #3067, opened 2022-07-22
🐛 Protect internal calls to `Object.*` from poisoning

dubzzz:
While fast-check can potentially push some code to alter existing and shared prototypes like the one of `Object`, it does not protect itself from this kind of issues.

The current PR solves part of the potential issues but does nothing regarding alterations of `Array.prototype.map`... At least not yet. It only focus on `Object`.

**_Category:_**

- [ ] ✅ Add or update tests
- [x] 🐛 Fix a bug
- [ ] 🏷️ Add or update types
- [ ] ⚡️ Improve performance
- [ ] _Other(s):_ ...
  
  

**_Potential impacts:_**

- [ ] Generated values
- [ ] Shrink values
- [ ] Performance
- [ ] Typings
- [ ] _Other(s):_ ...


dubzzz:
Add new Object.* calls done for record to the list of stable ones and test all of them against poisoning to make sure they will never break too much in the future.

PR #3102, opened 2022-08-30
🐛 Fix `safeApply` never calling original `apply`

dubzzz:
Existing version of the code was wrong as it was expecting `apply` to be directly defined on the called function while in theory it comes from the `Function.prototype`.

Given benchmarks we did on https://github.com/dubzzz/fast-check-benchmarks/blob/main/benchmark-poisoning-results.txt, we expect some performance uplift from this change:
- before: from 8,572.147 to 8,823.322 ops/sec
- after: from 15,083.692 to 15,573.392 ops/sec

Measurements on fast-check itself will be carried out before merging the fix to make sure of the performance boost.

**_Category:_**

- [ ] ✅ Add or update tests
- [x] 🐛 Fix a bug
- [ ] 🏷️ Add or update types
- [ ] ⚡️ Improve performance
- [ ] _Other(s):_ ...
  
  

**_Potential impacts:_**

- [ ] Generated values
- [ ] Shrink values
- [ ] Performance
- [ ] Typings
- [ ] _Other(s):_ ...

### DIFF
--- a/.yarn/versions/33387874.yml
+++ b/.yarn/versions/33387874.yml
@@ -1,0 +1,6 @@
+releases:
+  fast-check: patch
+
+declined:
+  - "@fast-check/ava"
+  - "@fast-check/jest"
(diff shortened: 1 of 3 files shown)


dubzzz:
Performance measurements for current main (https://github.com/dubzzz/fast-check/commit/dd5b27bc3d7c2f39b191ad4b4815f65561ab0edb) vs 3.1. vs this PR:

...

PR #3176, opened 2022-09-11
✨(poisoning) Do not track private globals

dubzzz:
Globals starting with a `_` are generally used for private stuffs such as: measuring coverage with nyc, gathering buffer related metas before dropping it to the console...

As we want poisoning to be easily pluggable without any extensive configuration, we prefer not tracking the content of such entity and just limit ourselves to check in their pointer changed or not but not deeply checking them.

**_Category:_**

- [ ] ✅ Add or update tests
- [ ] 🐛 Fix a bug
- [ ] 🏷️ Add or update types
- [ ] ⚡️ Improve performance
- [ ] _Other(s):_ ...
  
  

**_Potential impacts:_**

- [ ] Generated values
- [ ] Shrink values
- [ ] Performance
- [ ] Typings
- [ ] _Other(s):_ ...

### DIFF
--- a/packages/poisoning/src/internals/CaptureAllGlobals.ts
+++ b/packages/poisoning/src/internals/CaptureAllGlobals.ts
@@ -52,7 +52,7 @@ function captureOneRecursively(knownGlobals: AllGlobals, instance: unknown, name
       // For instance: do not track the content of globalThis.__coverage__
       continue;
     }
-    const subGlobalName = topLevel ? name + '.' + String(descriptorName) : String(descriptorName);
+    const subGlobalName = !topLevel ? name + '.' + String(descriptorName) : String(descriptorName);
     captureOneRecursively(knownGlobals, descriptor.value, subGlobalName, false);
   }
 }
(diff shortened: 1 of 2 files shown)


packages/poisoning/src/internals/CaptureAllGlobals.ts
```diff
@@ -47,14 +47,19 @@ function captureOneRecursively(knownGlobals: AllGlobals, instance: unknown, name
...
       continue;
     }
-    const subGlobalName = name !== 'globalThis' ? name + '.' + String(descriptorName) : String(descriptorName);
-    captureOneRecursively(knownGlobals, descriptor.value, subGlobalName);
+    if (topLevel && descriptorName[0] === '_') {
+      // Do not scan what's sounds like private properties dropped on globalThis
+      // For instance: do not track the content of globalThis.__coverage__
+      continue;
+    }
+    const subGlobalName = topLevel ? name + '.' + String(descriptorName) : String(descriptorName);
```
dubzzz, line 55:
```suggestion
    const subGlobalName = !topLevel ? name + '.' + String(descriptorName) : String(descriptorName);
```

PR #3264, opened 2022-10-03
:memo: Improve grammar in HowItWorks

gibson042:
**_Category:_**

- [ ]  Introduce new features
- [x]  Add or update documentation
- [ ]  Add or update tests
- [ ]  Fix a bug
- [ ]  Add or update types
- [ ]  Improve performance
- [ ] _Other(s):_ ...

### DIFF
--- a/packages/fast-check/documentation/HowItWorks.md
+++ b/packages/fast-check/documentation/HowItWorks.md
@@ -1,11 +1,11 @@
 # [:house:](../README.md) How it works?
 
-Let's take a deep dive into how property based testing frameworks work behind the scene.
+Let's take a deep dive into how property based testing frameworks work.
 Understanding how to implement a simple one should give you a better knowledge of fast-check and its internals.
 
-In this document we will describe how to build a property based testing framework (with limited functionalities) from scratch. Our home-made property based testing framework will be called `miniFc` and will rely on `pure-rand` to generate random values. It should help into understanding the big picture.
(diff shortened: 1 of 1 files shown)


dubzzz:
Thank you so much for the contribution. Will read through it in the coming days.

Remark: _No change needed regarding the failing tests. They are mostly things to be ignored in the context of updates on markdown files except the format and lint one. One complains because changes occurred in one of the packages and did not resulted into a bump of it, the other because of test coverage drop (I need to investigate this one but it's not at all related to your change as it turned flaky a few weeks ago)_

packages/fast-check/documentation/HowItWorks.md
```diff
@@ -227,7 +227,7 @@ miniFc.assert = (property, { seed = Date.now() } = {}) => {
 };
 ```
 
-In previous section, we did not cover the reason why we opted for pure random generators. In property based we want properties to be reproducible no matter the seed, no matter the hardware and no matter the time... But we also want to have independent runs for each iteration in the loop.
+The previous section did not explian the reason for using pure random generators. In property based testing, we want properties to be reproducible no matter the seed, no matter the hardware and no matter the time... But we also want to have independent runs for each iteration in the loop.
```
dubzzz, line 230: Do you mean 'explain' instead of 'explian'?

gibson042: Whoops, sure do.
```suggestion
The previous section did not explain the reason for using pure random generators. In property based testing, we want properties to be reproducible no matter the seed, no matter the hardware and no matter the time... But we also want to have independent runs for each iteration in the loop.
```

PR #3395, opened 2022-11-08
✨ Introduce new `gen` arbitrary

dubzzz:
An often raised complaint regarding property based testing is that writing the generators is complex. Indeed up-to-now generators have to be defined before hand, in other words making them depends on previous values or outputs coming from the code under tests is not really feasible or when it is it requires some extra thinking.

This new arbitrary is supposed to ease such cases to unlock more adoption for property based testing. But it comes with a cost:
- easier to write properties the wrong way
- shrinker is less efficient

The currently proposed implementation is very experimental and suffers from many drawbacks:
- ~unable to be used with other arbitraries (exclusive when used)~
- ~require caching mechanism to be implemented by the user~
- ~usage with cloneable arbitraries might not work properly~

**_Category:_**

- [ ] ✅ Add or update tests
- [ ] 🐛 Fix a bug
- [ ] 🏷️ Add or update types
- [ ] ⚡️ Improve performance
- [ ] _Other(s):_ ...
  
  

**_Potential impacts:_**

- [ ] Generated values
- [ ] Shrink values
- [ ] Performance
- [ ] Typings
- [ ] _Other(s):_ ...


dubzzz:
Actually, replay is not obvious on such arbitrary as the path depends on what has and will be generated. Indeed the set of arbitraries implied in a shrinking path may be different on each part of it. Replay would have to include these details to be possible, as we already did for commands.

moodmosaic:
I was about to create a similar issue and just came across this branch/PR ❤️ Is this going to be merged anytime soon? What's the current state?

moodmosaic:
Yes, exactly. The reason I *might* need something like this, is a tool I'm working on targeting (more or less) that category of users you mentioned. You start with a traditional test method (no parameters) and then you start adding parameters.

For each parameter you add (e.g. a parameter of type `string`) you'll get a random string, using fast-check internally, and perhaps `fc.auto` turns out to be useful here, haven't tried this branch yet.

moodmosaic:
FWIW, not sure if it really helps, the way I kind of simulated 'auto' in that specific scenario I described above, was by always passing a record. That is, the generated record uses (runtime defined) arbitraries to generate its values.

moodmosaic:
This is awesome 🎉

dubzzz:
Thanks a lot. Just have to write down the documentation for it and it will be released 🚀🤞

PR #3458, opened 2022-12-08
📝 Document how to use `context` in `examples`

tskj:
This PR addresses the issue raised [here](https://github.com/dubzzz/fast-check/issues/3459). I would prefer a proper fix in the library itself, but this at least gives the users the tools to circumvent the bug themselves!

I have added the suggested piece of code and short explanation to both parts of the docs so that they're easy to find. The idea is that users shouldn't have to stumble over this and have to dig through old issues to potentially find a fix, and can ideally be warned early enough that they circumvent the issue all together.

Please feel free to suggest better, friendlier, or clearer wording, or other places I should add this as well! 

**_Category:_**

- [x] 📝 Add or update documentation
- [ ] ✅ Add or update tests
- [ ] 🐛 Fix a bug
- [ ] 🏷️ Add or update types
- [ ] ⚡️ Improve performance
- [ ] _Other(s):_ ...
  
  

**_Potential impacts:_**

- [ ] Generated values
- [ ] Shrink values
- [ ] Performance
- [ ] Typings
- [ ] _Other(s):_ ...

### DIFF
--- a/.yarn/versions/93eb2223.yml
+++ b/.yarn/versions/93eb2223.yml
@@ -1,0 +1,2 @@
+declined:
+  - fast-check
--- a/packages/fast-check/documentation/Tips.md
+++ b/packages/fast-check/documentation/Tips.md
@@ -400,7 +400,7 @@ fc.assert(
 Please note that if you are using [custom examples](#add-custom-examples-next-to-generated-ones) in your tests, you will need the following implementation of `Context` in your examples:
 
 ```typescript
-const exampleContext = () => fc.sample(fc.context(), {numRuns:1})[0];
+const exampleContext = () => fc.sample(fc.context(), { numRuns: 1 })[0];
 ```
 
 ## Preview generated values
@@ -541,7 +541,7 @@ fc.assert(fc.property(fc.string(), fc.string(), fc.string(), myCheckFunction), {
 If you are using `Context` to [log within a predicate](#log-within-a-predicate) you will need to use the following context implementation in the examples.
 
 ```typescript
-const exampleContext = () => fc.sample(fc.context(), {numRuns:1})[0];
+const exampleContext = () => fc.sample(fc.context(), { numRuns: 1 })[0];
 
 fc.assert(fc.property(fc.string(), fc.string(), fc.context(), myCheckFunction), {
   examples: [['', '', exampleContext()]],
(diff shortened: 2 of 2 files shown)


dubzzz (commented):
Thanks for the suggestion, I'd definitely go for this one. I will merge soon

tskj:
> Thanks for the suggestion, I'd definitely go for this one. 

Great! I believe this will make users' lives much easier. 

However, I still strongly urge that eventually fast-check itself will output counter examples that are copy-pastable into the tests. As a very crude approximation, it can even output this very code itself, such as `Counterexamples: [[() => fc.sample(fc.context(), {numRuns:1})[0]]]` - there really is no reason to ask users to do this by hand!

dubzzz:
The PR does not pass, you'll need to run `yarn version:bump` on it and flag fast-check as a patch version.

Regarding the:
> However, I still strongly urge that eventually fast-check itself will output counter examples that are copy-pastable.

As the owner and maintainer of the repository **I receive urgency by no-one**. The project is made on free time without any funding today. If you really care about it, start funding the project and maybe I'll work on it. But today, it does not pay me at all and I will clearly not spend one second on such a second zone issue that will bring more harm than it fixes.

And once again, the repository being open source, if you get better ideas than someone having worked on the project for more than 5 years now, just fork it or wait.

tskj:
> The PR does not pass, you'll need to run yarn version:bump on it and flag fast-check as a patch version.

Ah, my apologies, I was confused by the CONTRIBUTING.md document, it seemed like I wouldn't have to do anything since I didn't do a code change. I have now set it to "decline" since there is no change to the code, let me know if that was wrong or if there is something else I have to do in addition!

> As the owner and maintainer of the repository I receive urgency by no-one.

I'm sorry if that is how it came across, I meant to use the phrase "to urge someone" which means "to strongly recommend".

I have no urgency for any fix as I already have my workaround, and I am indeed doing this PR only in the hopes that future people will be able to use this fine library without having to crawl through closed issues.

I do still strongly recommend, and indeed urge you to consider, to take issues with usability seriously. That is a concern of high importance for many development teams around the world, especially those who are careful about the dependencies they want to take on. In that light, having two first class features in your library that do not work together, is a bad look.

This is merely my personal opinion, I am only sharing it because I want the best for the project. What you do with that is up to you, and I respect what ever decision you come to.

dubzzz:
Seems that formatting is not passing

tskj:
Ah thanks for telling me. Not sure if that was what was breaking it since it didn't give me any error message other than the filename, but this should I work I think. Let me know if there is anything else!

PR #3467, opened 2022-12-10
✨(worker) Terminate workers if property timeout-ed

dubzzz:
Up-to-now ill workers stayed alive until the property has been executed on all the possible values. In other words, if one property was leading to infinite loops for some values, it would have created many workers consumming lots of memory and would have killed them only at the end of the whole shrinking process. It will now kill them as soon as possible to avoid keeping unneeded code running in background.

⚠️ We have to bump fast-check

**_Category:_**

- [ ] ✅ Add or update tests
- [ ] 🐛 Fix a bug
- [ ] 🏷️ Add or update types
- [ ] ⚡️ Improve performance
- [ ] _Other(s):_ ...
  
  

**_Potential impacts:_**

- [ ] Generated values
- [ ] Shrink values
- [ ] Performance
- [ ] Typings
- [ ] _Other(s):_ ...


packages/worker/test/internals/worker-pool/Lock.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+    const seenValues: number[] = [];
+    let lastAcquiredLock: AcquiredLock | undefined = undefined;
+
+    // Act
+    lock.acquire().then((acquired) => {
+      seenValues.push(1);
+      lastAcquiredLock = acquired;
+    });
+    lock.acquire().then((acquired) => {
+      seenValues.push(1);
```
dubzzz, line 27:
```suggestion
      seenValues.push(2);
```

packages/worker/test/internals/worker-pool/Lock.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+      lastAcquiredLock = acquired;
+    });
+    lock.acquire().then((acquired) => {
+      seenValues.push(1);
```
dubzzz, line 31:
```suggestion
      seenValues.push(3);
```

PR #3511, opened 2022-12-25
✨(jest) Let fast-check guide timeouts

### LINKED ISSUE #3084: Integrate jest timeouts to @fast-check/jest
## 🚀 Feature Request

Jest tests themselves have a timeout. It would be nice if these timeouts that are added into the `testProp` was used for fast check's timeout.

## Motivation

This makes it easier to know that the fast check `interruptAfterTimeLimit` will be always be under the timeout given to the full jest test itself. I suggest that it may need to be the jest timeout minus 1 second.

## Example

```
import { itProp, testProp, fc } from '@fast-check/jest';

describe('', () => {
  itProp('should run too', [fc.nat(), fc.nat()], (a, b) => {
    return a + b === b + a;
  }, 5000);
});
```

The `5000` should then be used for the `interruptAfterTimeLimit`.

dubzzz:
Up-to-now fast-check and jest had their own and separate mechanisms to handle timeouts of running tests.
On one side fast-check defined interruptAfterTimeLimit, on the other Jest defined timeout.

Trick has been to define interruptAfterTimeLimit = 0.9 * timeout, so that fast-check timeouts before jest does (0.ç just being one possible value among others, but it does not mean it will always be fast-check before jest).

This trick will not be needed anymore as when relying on `@fast-check/jest`, timeouts of jest will be redirected to fast-check so that fast-check will be responsible for any timeout related to pbt tests.

Fixes #3084

**_Category:_**

- [ ] ✅ Add or update tests
- [ ] 🐛 Fix a bug
- [ ] 🏷️ Add or update types
- [ ] ⚡️ Improve performance
- [ ] _Other(s):_ ...
  
  

**_Potential impacts:_**

- [ ] Generated values
- [ ] Shrink values
- [ ] Performance
- [ ] Typings
- [ ] _Other(s):_ ...

### DIFF
--- a/packages/jest/src/internals/TestWithPropRunnerBuilder.ts
+++ b/packages/jest/src/internals/TestWithPropRunnerBuilder.ts
@@ -18,6 +18,7 @@ export function buildTestWithPropRunner<Ts extends [any] | any[], TsParameters e
   fc: FcExtra
 ): void {
   const customParams: FcParameters<TsParameters> = { ...params };
+  // Handle seed
   if (customParams.seed === undefined) {
     const seedFromGlobals = readConfigureGlobal().seed;
     if (seedFromGlobals !== undefined) {
@@ -31,6 +32,21 @@ export function buildTestWithPropRunner<Ts extends [any] | any[], TsParameters e
       }
     }
   }
+  // Handle timeout
+  if (customParams.interruptAfterTimeLimit === undefined) {
+    // Copy global configuration of interruptAfterTimeLimit as local one
+    customParams.interruptAfterTimeLimit = fc.readConfigureGlobal().interruptAfterTimeLimit;
+  }
+  const jestTimeout = timeout !== undefined ? timeout : extractJestGLobalTimeout();
+  if (jestTimeout !== undefined) {
+    if (customParams.interruptAfterTimeLimit === undefined) {
+      // Use the timeout specified at jest's level for interruptAfterTimeLimit
+      customParams.interruptAfterTimeLimit = jestTimeout;
+    } else {
+      // Mix both jest and fc's timeouts
+      customParams.interruptAfterTimeLimit = Math.min(customParams.interruptAfterTimeLimit, jestTimeout);
+    }
+  }
 
   const promiseProp = wrapProp(prop);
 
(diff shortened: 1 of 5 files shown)


packages/jest/src/internals/TestWithPropRunnerBuilder.ts
```diff
@@ -41,6 +57,24 @@ export function buildTestWithPropRunner<Ts extends [any] | any[], TsParameters e
...
+  for (const key of Object.getOwnPropertySymbols(globalThis)) {
+    if (String(key) === stateSymbolStringValue) {
+      const jestState = (globalThis as any)[key];
+      if (jestState !== null && typeof jestState === 'object' && typeof jestState.testTimeout === 'number') {
+        return jestState.testTimeout;
+      }
+    }
+  }
+  return undefined; // no such case expected
+}
```
dubzzz, line 80: @SimenB Maybe you'll have a better and longer term suggestion for this hack?!

**What was the problem?** Jest and fast-check had both their timeouts. A common trick I used to give to people was: "use interruptAfterTimeLimit in fast-check with a value slightly inferior to the timeout of Jest". But this trick was damn broken as while fast-check will let's say wait 1000ms and jest 1100ms, there is no certainty that the timeout of fast-check will end before the one of Jest.

**What's the fix about?** In the context of `it.prop` instead of asking jest to handle the timeouts, the handling is fully degated to fast-check, so that there is not risk of overlaps. But to do so I had to find what would be the timeout on jest side... thus this hacky code.

**Is it safe today?** So far, I made it safe by testing it extensively via tests making sure fast-check will always be connected to the expected timeouts. The tests apply various configurations of Jest and timeouts and make sure fast-check extracts the right timeouts.

dubzzz: In other words, following this change, when running:

```ts
it.prop([...], async () => {....})
```

The timeout will not be handled by Jest but delegated to fast-check. It will thus be able to interrupt predicate being running and decide whether or not the interrupt was ok or not. But above all it will print the stack, the seed... while today it gets interrupted at the middle and does not output anything.

PR #3631, opened 2023-02-13
Lock pure-rand dependency to 6.0.0 instead of range

paulmillr:
This will protect against potential supply chain attacks.

Your library is used in security-critical apps. Even though lockfiles are utilized, i'd prefer a simple deterministic solution.

### DIFF
--- a/packages/fast-check/package.json
+++ b/packages/fast-check/package.json
@@ -53,7 +53,7 @@
     "node": ">=8.0.0"
   },
   "dependencies": {
-    "pure-rand": "^6.0.0"
+    "pure-rand": "6.0.0"
   },
   "devDependencies": {
     "@fast-check/poisoning": "workspace:*",


dubzzz:
pure-rand being also one of my other projects, I prefer letting the ^ to benefit from potential bug fixes, performance improvements... occuring into it. As a consequence, I'll reject your suggestion for now

paulmillr:
@dubzzz if a user needs performance improvements and bugfixes, they can bump fast-check which will bump rand. If they don't need it, they will be more secure. For example, i'm using version 3.0.0 and I see zero need in upgrading it to something newer. However pure rand is still going to be bumped automatically.

Your package update could be distributing malware for all I know. You don't sign it, and your security practices likely aren't that tough. That's fine. But for those who work in the particular field the change would help.

dubzzz:
> You don't sign it

I can probably start by this part first. Signing fast-check and pure-rand should be feasible. I just need to check what to use to sign them.

paulmillr:
Can this be re-opened?

10 months have passed and pure-rand is still a version range, awaiting for your credentials to be leaked to start distributing malware to all fast-check users.

paulmillr:
https://github.com/dubzzz/fast-check/issues/4475

PR #3775, opened 2023-04-20
🔧 Publish all packages in one workflow

dubzzz:
Up-to-now each package being published triggered its own fresh workflow. It means that in addition to the one launched for main, we also pushed one for each tag.

We want to avoid relaunching workflows in such case and just have a re-lauchable step that could read tags at commit level when executed and publish versions accordingly.

**_Category:_**

- [ ] ✅ Add or update tests
- [ ] 🐛 Fix a bug
- [ ] 🏷️ Add or update types
- [ ] ⚡️ Improve performance
- [ ] _Other(s):_ ...
  
  

**_Potential impacts:_**

- [ ] Generated values
- [ ] Shrink values
- [ ] Performance
- [ ] Typings
- [ ] _Other(s):_ ...


dubzzz:
Worth checking https://github.com/orgs/community/discussions/26639#discussioncomment-3949595

PR #3891, opened 2023-05-19
⚡️ Speed-up race-condition schedulers

dubzzz:
**:boom: Breaking change**

This PR makes schedulers faster. By being faster it means that code that was previously passing may start to fail. Additionally it may predates `waitAll` and makes it not working in some cases that relied on this lucky slow scheduler.

The new scheduler implementation will try to be as fast as possible and skip as few as possible frames. In this change scheduler takes action in places where it used not to because it was waiting for no reasons.

**_Category:_**

- [ ] ✅ Add or update tests
- [ ] 🐛 Fix a bug
- [ ] 🏷️ Add or update types
- [x] ⚡️ Improve performance
- [ ] _Other(s):_ ...
  
  

**_Potential impacts:_**

- [ ] Generated values
- [ ] Shrink values
- [ ] Performance
- [ ] Typings
- [x] _Other(s):_ Stricter, less room for sub-optimal asynchronous code


packages/fast-check/src/arbitrary/_internals/implementations/SchedulerImplem.ts
```diff
@@ -222,9 +222,10 @@ export class SchedulerImplem<TMetaData> implements Scheduler<TMetaData> {
     });
   }
 
-  async waitOne(customAct?: SchedulerAct): Promise<void> {
+  waitOne(customAct?: SchedulerAct): Promise<void> {
```
dubzzz, line 225:
```suggestion
  async waitOne(customAct?: SchedulerAct): Promise<void> {
```

packages/fast-check/src/arbitrary/_internals/implementations/SchedulerImplem.ts
```diff
@@ -222,9 +222,10 @@ export class SchedulerImplem<TMetaData> implements Scheduler<TMetaData> {
...
     const waitAct = customAct || defaultSchedulerAct;
-    await this.act(() => waitAct(async () => await this.internalWaitOne()));
+    const waitOneResult: Promise<unknown> = this.act(() => waitAct(() => this.internalWaitOne()));
+    return waitOneResult as Promise<void>;
```
dubzzz, line 228:
```suggestion
    await this.act(() => waitAct(() => this.internalWaitOne()));
```

packages/fast-check/src/arbitrary/_internals/implementations/SchedulerImplem.ts
```diff
@@ -224,7 +224,7 @@ export class SchedulerImplem<TMetaData> implements Scheduler<TMetaData> {
 
   async waitOne(customAct?: SchedulerAct): Promise<void> {
     const waitAct = customAct || defaultSchedulerAct;
-    await this.act(() => waitAct(async () => await this.internalWaitOne()));
+    await this.act(() => waitAct(() => this.internalWaitOne()));
```
dubzzz, line 227: I kept one extra await for now. I hope it could pass the tests with only one of them remaining. Anyway, the ideal thing would be to have none of them but so far it means breaking some tests so better investigating them before attempting anything dropping the last one.

By the way, before merging I should probably check how breaking such change could be for our users. The fact that dropping all the await broke our own tests proves that we should be careful with that change.

website/docs/tutorials/detect-race-conditions/snippets.mjs
```diff
@@ -78,15 +78,17 @@ export const queueCodeV5 = `export function queue(fun) {
...
+           .then(resolve, reject);
+        });
       }
       return new Promise((resolve, reject) => {
         onDone.push(() => {
           const p = fun(...args);
-          p.then(runNext, runNext);
-          p.then(resolve, reject);
+          p.then(v => { runNext(); return v; }, e => { runNext(); throw e; })
+           .then(resolve, reject);
```
dubzzz, line 91:
```suggestion
          p.then(v => { runNext(); resolve(v); }, e => { runNext(); reject(e); });
```

website/docs/tutorials/detect-race-conditions/snippets.mjs
```diff
@@ -78,15 +78,17 @@ export const queueCodeV5 = `export function queue(fun) {
     return (...args) => {
       if (!pending) {
         pending = true;
-        const p = fun(...args);
-        p.then(runNext, runNext);
-        return p;
+        return new Promise((resolve, reject) => {
+          const p = fun(...args);
+          p.then(v => { runNext(); return v; }, e => { runNext(); throw e; })
+           .then(resolve, reject);
```
dubzzz, line 84:
```suggestion
          p.then(v => { runNext(); resolve(v); }, e => { runNext(); reject(e); });
```

website/docs/tutorials/detect-race-conditions/snippets.mjs
```diff
@@ -352,7 +356,9 @@ export const extendedWithExceptionsPBTSpecCode = `import {queue} from './queue.j
           const monitoredScheduledCall = (...args) => {
             concurrentQueriesDetected ||= queryPending;
             queryPending = true;
-            return scheduledCall(...args).finally(() => (queryPending = false));
+            const out = scheduledCall(...args);
+            out.finally(() => (queryPending = false));
```
dubzzz, line 360:
```suggestion
            out.finally(() => (queryPending = false)).catch(() => {});
```

packages/fast-check/src/arbitrary/_internals/implementations/SchedulerImplem.ts
```diff
@@ -224,7 +224,8 @@ export class SchedulerImplem<TMetaData> implements Scheduler<TMetaData> {
 
   async waitOne(customAct?: SchedulerAct): Promise<void> {
```
dubzzz, line 225:
```suggestion
   waitOne(customAct?: SchedulerAct): Promise<void> {
```

website/docs/tutorials/detect-race-conditions/one-step-closer-to-real-usages.mdx
```diff
@@ -35,16 +35,12 @@ pendingQueries.push(queued(2).then((v) => seenAnswers.push(v)));
 await s.waitFor(Promise.all(pendingQueries));
 ```
 
-An alternative options would have been to use `waitAll` instead of `waitFor` but it comes with a precise requirements being that promise had to be imperatively scehduled by the time we reach it. In other words, if our code delays a little bit the call to `call` it will not wait anything. With the current implementation of `queue` it works, but with next versions of it it might not be the case anymore as queued calls might start queuing themselves behind already started and running promises.
+An alternative would have been to use `waitAll` instead of `waitFor` but it comes with a precise requirement: promises have to be imperatively scheduled by the time we reach them. In other words, if our code delays a little bit the call to `call`, the scheduler might not wait enough.
 
 As the fact that `call` has to be fired synchronously when issuing a call on a clean context is not a requirement for our current function, we can relax the constraint in our test to make evolving this implementation easier.
```
dubzzz, line 40:
```suggestion
Given the fact that `call` being fired synchronously is not a requirement for our current function, we can relax the constraint in our test to make evolving this implementation easier.
```

website/docs/tutorials/detect-race-conditions/one-step-closer-to-real-usages.mdx
```diff
@@ -35,16 +35,12 @@ pendingQueries.push(queued(2).then((v) => seenAnswers.push(v)));
 await s.waitFor(Promise.all(pendingQueries));
 ```
 
-An alternative options would have been to use `waitAll` instead of `waitFor` but it comes with a precise requirements being that promise had to be imperatively scehduled by the time we reach it. In other words, if our code delays a little bit the call to `call` it will not wait anything. With the current implementation of `queue` it works, but with next versions of it it might not be the case anymore as queued calls might start queuing themselves behind already started and running promises.
+An alternative would have been to use `waitAll` instead of `waitFor` but it comes with a precise requirement: promises have to be imperatively scheduled by the time we reach them. In other words, if our code delays a little bit the call to `call`, the scheduler might not wait enough.
```
dubzzz, line 38:
```suggestion
An alternative would have been to use `waitAll` instead of `waitFor` but it comes with a precise requirement: promises have to be already scheduled by the time we request the scheduler. In other words, if our code delays a little bit the call to `call`, the scheduler might not wait enough.
```

packages/fast-check/src/arbitrary/_internals/implementations/SchedulerImplem.ts
```diff
@@ -184,16 +183,11 @@ export class SchedulerImplem<TMetaData> implements Scheduler<TMetaData> {
...
-      };
-      // We will run current item if and only if the one preceeding it succeeds
-      // Otherwise, we mark the run as "failed" and ignore any subsequent item (including this one)
-      previouslyScheduled = previouslyScheduled.then(onNextItem, onFaultyItem);
-    }
-    // Once last item is done, the full sequence can be considered as successful
-    // If it failed (or any preceeding one failed), then the sequence should be marked as failed (already marked for others)
-    // We always handle Promise rejection of previouslyScheduled internally, in other words no error will bubble outside
-    previouslyScheduled.then(onDone, onFaultyItemNoThrow);
+        registerNextBuilder(index + 1, scheduled);
```
dubzzz, line 186: No stack overflow on this line. Recursive call occurs via thenables, we can stack them for long.

packages/fast-check/test/unit/arbitrary/_internals/implementations/SchedulerImplem.spec.ts
```diff
@@ -1564,10 +1741,7 @@ describe('SchedulerImplem', () => {
       task.then((v) => (taskResolvedValue = v));
 
       // Assert
-      while (s.count() !== 0) {
-        expect(taskResolvedValue).toBe(null);
-        await s.waitOne();
-      }
+      await s.waitAll();
```
dubzzz, line 1744: Adds an extra await, needed for this test

packages/fast-check/test/unit/arbitrary/_internals/implementations/SchedulerImplem.spec.ts
```diff
@@ -1588,10 +1762,7 @@ describe('SchedulerImplem', () => {
...
```
dubzzz, line 1765: Adds an extra await, needed for this test

PR #4059, opened 2023-07-10
⚡ Faster `decomposeFloat/Double`

zbjornson:
This replaces the (creative) loop-based implementation with a constant-time one. The number 1.234 is about 1100x faster and 1e30 is about 2300x faster.

If you're okay with this change I can do the same thing for double.

**_Category:_**

- [ ] ✅ Add or update tests
- [ ] 🐛 Fix a bug
- [ ] 🏷️ Add or update types
- [x] ⚡️ Improve performance
- [ ] _Other(s):_ ...
  
  

**_Potential impacts:_**

- [ ] Generated values
- [ ] Shrink values
- [x] Performance - faster
- [ ] Typings
- [ ] _Other(s):_ ...

### DIFF
--- a/packages/fast-check/src/arbitrary/_internals/helpers/FloatHelpers.ts
+++ b/packages/fast-check/src/arbitrary/_internals/helpers/FloatHelpers.ts
@@ -1,9 +1,5 @@
-const safeNumberIsInteger = Number.isInteger;
-const safeNumberIsNaN = Number.isNaN;
-
 const safeNegativeInfinity = Number.NEGATIVE_INFINITY;
 const safePositiveInfinity = Number.POSITIVE_INFINITY;
-const safeNaN = Number.NaN;
 
 /** @internal */
 export const MIN_VALUE_32 = 2 ** -126 * 2 ** -23;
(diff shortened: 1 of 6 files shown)


dubzzz:
Looks awesome. I'll spend some time checking the diff but I think we can go with a similar optim for doubles

zbjornson:
Thanks! I'm happy to work on the double variant also.

It looks like I broke something. How do you get unit tests to run locally? If I run `yarn workspace fast-check test`, I get a ton of TS errors.

packages/fast-check/src/arbitrary/_internals/helpers/FloatHelpers.ts
```diff
@@ -17,12 +16,18 @@ const INDEX_POSITIVE_INFINITY = 2139095040; // floatToIndex(MAX_VALUE_32) + 1;
 /** @internal */
 const INDEX_NEGATIVE_INFINITY = -2139095041; // floatToIndex(-MAX_VALUE_32) - 1
 
+const buffer = new ArrayBuffer(4);
+const view = new DataView(buffer);
```
dubzzz, line 20: **Perfect! It even works with node 8** which is the currently minimal supported version for fast-check (will be dropped soon, but so far v3 is supposed to support it).

dubzzz:
You may try:

```sh
yarn # install all deps
yarn build:all # build all packages

# usually I changedir into /packages/fast-check
# but workspace commands also work
yarn test
yarn e2e # runs against prod bundle so need `yarn build` to be called before if you change things
```

You'll also need to declare the version bump (minor for fast-check and none for others):

```sh
yarn version:bump
```

_I'd need to update my contributing guide with this setup_

zbjornson:
Thanks for the guidance there. Fixed the failing test, switched from `DataView` to `TypedArray` for slightly better perf, ran `version:bump`, and added another commit for `decomposeDouble`.

dubzzz:
It seems that format is invalid. Can you please run `yarn format` on your PR?

dubzzz (approved):

dubzzz:
Thank you so much for this awesome contribution 🥰

zbjornson:
My pleasure, thanks for the fast review.

PR #4202, opened 2023-09-05
🔥 Remove `withDeletedKeys` from `record`

dubzzz:
**:boom: Breaking change**

Initially `record` came with `withDeletedKeys`, but quickly we received the need to have a better granularity level to deal with keys to keep or not and so `requiredKeys` came.

Now, that we have it for some time, we want to drop `withDeletedKeys` as it tends to add complexity to the code and the API.

A migration path and possibly some scripts to help users to run the change might be provided in the future. For now the recommendation would be to replace: `withDeletedKeys: true` by `requiredKeys: []` and `withDeletedKeys: false` by nothing or `requiredKeys: [...here all the keys...]`.

**_Category:_**

- [ ] ✅ Add or update tests
- [ ] 🐛 Fix a bug
- [ ] 🏷️ Add or update types
- [ ] ⚡️ Improve performance
- [x] _Other(s):_ Drop deprecated feature
  
  

**_Potential impacts:_**

- [ ] Generated values
- [ ] Shrink values
- [ ] Performance
- [ ] Typings
- [x] _Other(s):_ ...

### DIFF
--- a/packages/fast-check/test/unit/arbitrary/record.spec.ts
+++ b/packages/fast-check/test/unit/arbitrary/record.spec.ts
@@ -55,36 +55,6 @@ describe('record', () => {
       ),
     ));
 
-  it('should call buildPartialRecordArbitrary with keys=[] when constraints defines withDeletedKeys=true', () =>
-    fc.assert(
-      fc.property(
-        fc.uniqueArray(keyArb, { minLength: 1 }),
-        fc.option(fc.boolean(), { nil: undefined }),
-        (keys, noNullPrototype) => {
-          // Arrange
-          const recordModel: Record<string | symbol, Arbitrary<any>> = {};
-          for (const k of keys) {
-            const { instance } = fakeArbitrary();
-            recordModel[k] = instance;
-          }
-          const { instance } = fakeArbitrary<any>();
-          const buildPartialRecordArbitrary = jest.spyOn(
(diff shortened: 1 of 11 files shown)


packages/fast-check/test/unit/arbitrary/record.spec.ts
```diff
@@ -209,18 +155,9 @@ describe('record (integration)', () => {
...
-  const extraParameters: fc.Arbitrary<Extra> = fc
-    .tuple(metaArbitrary, constraintsArbitrary)
-    .map(([metas, constraintsMeta]: [Meta[], { withDeletedKeys?: boolean; withRequiredKeys?: true }]) => {
-      if ('withRequiredKeys' in constraintsMeta) {
-        return [metas, { requiredKeys: metas.filter((m) => m.kept).map((m) => m.key) }] as [Meta[], RecordConstraints];
-      }
-      return [metas, constraintsMeta] as [Meta[], RecordConstraints];
-    });
+  const extraParameters: fc.Arbitrary<Extra> = metaArbitrary.map((metas) => {
+    return [metas, { requiredKeys: metas.filter((m) => m.kept).map((m) => m.key) }] as [Meta[], RecordConstraints];
```
dubzzz, line 159: Need to replicate the case: no requiredKeys in the constraints

PR #4205, opened 2023-09-05
✨ Support null-proto in `record`

dubzzz:
For now, users will have to explicitly ask for it. But in the next major, the null-prototype will be the default.

**_Category:_**

- [ ] ✅ Add or update tests
- [ ] 🐛 Fix a bug
- [ ] 🏷️ Add or update types
- [ ] ⚡️ Improve performance
- [ ] _Other(s):_ ...
  
  

**_Potential impacts:_**

- [ ] Generated values
- [ ] Shrink values
- [ ] Performance
- [ ] Typings
- [ ] _Other(s):_ ...

### DIFF
--- a/packages/fast-check/src/arbitrary/record.ts
+++ b/packages/fast-check/src/arbitrary/record.ts
@@ -7,7 +7,7 @@ import { EnumerableKeyOf } from './_internals/helpers/EnumerableKeysExtractor';
  * @remarks Since 0.0.12
  * @public
  */
-export type RecordConstraints<T = unknown> =
+export type RecordConstraints<T = unknown> = (
   | {
       /**
        * List keys that should never be deleted.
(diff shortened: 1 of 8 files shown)


packages/fast-check/src/arbitrary/record.ts
```diff
@@ -34,7 +34,15 @@ export type RecordConstraints<T = unknown> =
        * @deprecated Prefer using `requiredKeys: []` instead of `withDeletedKeys: true` as the flag will be removed in the next major
        */
       withDeletedKeys?: boolean;
-    };
+    }
+) & {
+  /**
+   * Do not generate objects with null prototype
```
dubzzz, line 40:
```suggestion
   * Do not generate records with null prototype
```

PR #4208, opened 2023-09-05
🔥 Enforce `run{Before/After}Each` on property

dubzzz:
**:boom: Breaking change**

The API of property has been updated in a non-breaking manner during v3. Now, let's seal that move to drop unneeded code complexity, drop useless tests... but also make the API clearer and way simpler to use and manage.

The implementations of `IRawProperty` now have to implement `runBeforeEach` and `runAfterEach`. If they themselves wrap another `IRawProperty`, they probably should at least call these methods when called on their side of them. The migration from v3, consists into implementing the two methods.

**_Category:_**

- [ ] ✅ Add or update tests
- [ ] 🐛 Fix a bug
- [ ] 🏷️ Add or update types
- [ ] ⚡️ Improve performance
- [x] _Other(s):_ Enforce new API
  
  

**_Potential impacts:_**

- [ ] Generated values
- [ ] Shrink values
- [ ] Performance
- [ ] Typings
- [x] _Other(s):_ ...


packages/worker/src/internals/NoopWorkerProperty.ts
```diff
@@ -25,4 +25,10 @@ export class NoopWorkerProperty<Ts> implements WorkerProperty<Ts> {
   run(): Promise<PreconditionFailure | PropertyFailure | null> {
     throw new Error('Method not implemented.');
   }
+  runBeforeEach(): Promise<void> {
```
dubzzz, line 28: Should be done in a separate PR

packages/ava/package.json
```diff
@@ -47,7 +47,7 @@
   },
   "homepage": "https://github.com/dubzzz/fast-check/tree/main/packages/ava#readme",
   "dependencies": {
-    "fast-check": "^3.0.0"
+    "fast-check": "^3.0.0 || ^4.0.0"
```
dubzzz, line 50: Should probably be done in another PR just at release time 🤔

packages/worker/src/internals/NoopWorkerProperty.ts
dubzzz: Done in #4467

Let's consider this change as a no change from a versioning point-of-view.

PR #4354, opened 2023-10-21
:bug: use a single types definition for require and import

sukovanej:
**_Category:_**

- [ ] ✅ Add or update tests
- [x] 🐛 Fix a bug
- [ ] 🏷️ Add or update types
- [ ] ⚡️ Improve performance
- [ ] _Other(s):_ ...
  
  

**_Potential impacts:_**

- [ ] Generated values
- [ ] Shrink values
- [ ] Performance
- [x] Typings
- [ ] _Other(s):_ ...

This change is motivated by my recent experimentation with esm + [@effect/schema](https://github.com/Effect-TS/schema) + fast-check. During the type-checking, it looks like the tsc uses esm version of the fast-check for my package (as it uses `"type": "module"`) and the cjs version of the fast-check is used for @effect/schema. Probably due to unique symbols and private methods / attributes in the fast-check .d.ts files, the type-checker reports a mismatch.

Please also check and feel free to join the discussion in the [effect discord thread](https://discord.com/channels/795981131316985866/1162783872224854167)

### DIFF
--- a/packages/fast-check/package.json
+++ b/packages/fast-check/package.json
@@ -7,12 +7,11 @@
   "exports": {
     "./package.json": "./package.json",
     ".": {
+      "types": "./lib/types/fast-check.d.ts",
       "require": {
-        "types": "./lib/types/fast-check.d.ts",
         "default": "./lib/fast-check.js"
       },
       "import": {
-        "types": "./lib/esm/types/fast-check.d.ts",
         "default": "./lib/esm/fast-check.js"
       }
     }
@@ -28,7 +27,7 @@
   "scripts": {
     "build": "yarn build:publish-cjs && yarn build:publish-esm && yarn build:publish-types && node postbuild/main.cjs",
     "build-ci": "cross-env EXPECT_GITHUB_SHA=true yarn build",
-    "build:publish-types": "tsc -p tsconfig.publish.types.json && tsc -p tsconfig.publish.types.json --outDir lib/esm/types",
+    "build:publish-types": "tsc -p tsconfig.publish.types.json",
     "build:publish-cjs": "tsc -p tsconfig.publish.json",
     "build:publish-esm": "tsc -p tsconfig.publish.json --module es2015 --moduleResolution node --outDir lib/esm && cp package.esm-template.json lib/esm/package.json",
     "typecheck": "tsc --noEmit",
--- a/packages/fast-check/postbuild/main.cjs
+++ b/packages/fast-check/postbuild/main.cjs
@@ -70,16 +70,6 @@ fs.readFile(path.join(__dirname, '../package.json'), (err, data) => {
     console.info(`Package details added onto d.ts version for cjs`);
   }
 
-  const dTsReplacement2 = replace.sync({
-    files: 'lib/esm/types/fast-check-default.d.ts',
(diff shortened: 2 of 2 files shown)


packages/fast-check/package.json
```diff
@@ -7,12 +7,11 @@
   "exports": {
     "./package.json": "./package.json",
     ".": {
+      "types": "./lib/types/fast-check.d.ts",
```
dubzzz, line 10: It would result into an invalid definition of the typings for TypeScript. Let's run the CI against it to confirm but I expect the runs of type-check backed by TypeScript with NodeNext to fail and publint to complain.

dubzzz: I think there is probably a problem on effect's side regarding it's support for esm/NodeNext.

dubzzz: But in theory (and it's not the case for @effect/schema), mjs and cjs files should never share the same d.ts definition. I had a long public chat about that specific issue on Twitter at: https://twitter.com/AndaristRake/status/1695550761600503827?t=EkApRjlL8ZmEYqZtKtCToA&s=19.

dubzzz:
@Andarist provided a clear details of what exactly happens in this specific case: https://twitter.com/AndaristRake/status/1716195259024052359?t=0jwHoJrqrfPdXATxW3hPbA&s=19

sukovanej:
Thanks a lot for dealing with this issue. I'll close the PR. Tho, I still have no idea what should be the solution, the @effect/schema with fast-check simply won't type-check if the consumer package is esm right now. 😢

dubzzz:
I'll try to see if I can find something on my end

Andarist:
Well, due to the nature of `fast-check` setup, this can just happen "anytime" in mixed CJS/ESM dependency graphs. `@effect/schema` is just a specific example of that problem. 

There are basically, two answers here (that I know of):
- `fast-check` could *not* ship a dual package. Maybe a Preconstruct-like setup could be used here. This would eliminate the dual package hazard completely. The same would be eliminated if `fast-check` would use a simpler one-format setup (where CJS still has a somewhat wider support).
- the responsibility to deal with this is pushed onto the consumers. The consumer might fight this by loading the CJS variant of `fast-check` using `createRequire` etc. 

Of course, Effect could also consider their packaging strategy here. But that would only address this specific problem of those 2 specific packages - the problem itself could arise at any time for just any other CJS package that integrates with `fast-check`.

dubzzz:
@sukovanej Did you find a way to bypass the issue? I'm trying to find ways not to force my users to have to do ugly things to use the library so I hope to come up with something that could work in the coming weeks

sukovanej:
Hey, I ended up ditching the esm for a moment as the original problem got resolved by a fix on the effect side. And I'm not facing the problem with cjs. So unfortunately, I won't be able to help much at the moment.

Maybe one note, I originally wanted to prepare a different MR there. I wanted to mark internal methods as `@internal` and strip them from the generated d.ts files. My assumption was that if the two cjs and esm type definitions were structurally identical, the type checker would be actually happy. So this might be something to consider.

dubzzz:
Stripping them from the d.ts might indeed be a good idea

PR #4384, opened 2023-10-24
💚 Attempt to patch untar step

dubzzz:
Third attempt, another philosophy now.

### DIFF
--- a/.github/workflows/build-status.yml
+++ b/.github/workflows/build-status.yml
@@ -263,6 +263,10 @@ jobs:
           name: bundles
           path: packages/
       - name: Unpack production packages
+        if: runner.os != 'windows-latest'
+        run: yarn unpack:all
+      - name: Unpack production packages (Windows only)
+        if: runner.os == 'windows-latest'
         # untar seems to fail on Windows when untaring already existing files
         run: yarn unpack:all || true
       - name: Alter internals to behave as if published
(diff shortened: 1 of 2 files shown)


.github/workflows/build-status.yml
```diff
@@ -263,7 +263,8 @@ jobs:
           name: bundles
           path: packages/
       - name: Unpack production packages
-        run: yarn unpack:all
+        # untar seems to fail on Windows when untaring already existing files
+        run: yarn unpack:all || true
```
dubzzz, line 267:
```suggestion
        if: runner.os != 'windows-latest'
        run: yarn unpack:all
      - name: Unpack production packages (Windows only)
        if: runner.os == 'windows-latest'
        # untar seems to fail on Windows when untaring already existing files
        run: yarn unpack:all || true
```

.github/workflows/build-status.yml
```diff
@@ -263,7 +263,12 @@ jobs:
           name: bundles
           path: packages/
       - name: Unpack production packages
+        if: runner.os != 'windows-latest'
         run: yarn unpack:all
+      - name: Unpack production packages (Windows only)
+        if: runner.os == 'windows-latest'
```
dubzzz, line 269:
```suggestion
        if: matrix.os != 'windows-latest'
        run: yarn unpack:all
      - name: Unpack production packages (Windows only)
        if: matrix.os == 'windows-latest'
```

PR #4584, opened 2024-01-02
👷(poisoning) Move build chain to ESM

dubzzz:
The build chain of `@fast-check/poisoning` has been CommonJS-based since day 1. With ESM moving forward in the ecosystem, it's time to move ourselves to the new standard and adapt our build chains to ESM.

Unfortunately it may have some subtle impacts on our users as our package will not be a CJS one offering a ESM fallback anymore. I will rather be the opposite: an ESM package with a fallback to CJS. It implies that we moved ESM related files closer to the root of the package (we could have kept them in esm/) and moved the CJS ones further in the file structure (we had to move them).

Another subtle impact is that it would impose our users to run at least Node ≥12.17.0.

As such we consider it as a breaking change.

**_Category:_**

- [ ] ✅ Add or update tests
- [ ] 🐛 Fix a bug
- [ ] 🏷️ Add or update types
- [ ] ⚡️ Improve performance
- [ ] _Other(s):_ ...
  
  

**_Potential impacts:_**

- [ ] Generated values
- [ ] Shrink values
- [ ] Performance
- [ ] Typings
- [ ] _Other(s):_ ...

### DIFF
--- a/packages/poisoning/package.json
+++ b/packages/poisoning/package.json
@@ -2,22 +2,22 @@
   "name": "@fast-check/poisoning",
   "description": "Set of utilities to ease detection and revert of poisoning",
   "version": "0.1.0",
-  "type": "commonjs",
+  "type": "module",
   "main": "lib/main.js",
   "exports": {
     "./package.json": "./package.json",
     ".": {
       "require": {
-        "types": "./lib/main.d.ts",
-        "default": "./lib/main.js"
+        "types": "./lib/cjs/main.d.ts",
+        "default": "./lib/cjs/main.js"
       },
       "import": {
-        "types": "./lib/esm/main.d.ts",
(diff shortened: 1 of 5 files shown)


packages/poisoning/package.json
```diff
@@ -2,33 +2,33 @@
   "name": "@fast-check/poisoning",
   "description": "Set of utilities to ease detection and revert of poisoning",
   "version": "0.1.0",
-  "type": "commonjs",
+  "type": "module",
   "main": "lib/main.js",
```
dubzzz, line 6: Should I keep it referencing the cjs version for backward compatibility (pre-node 12)? So far it does not seem to be a good option, but at the same time I don't know if this field still makes sense today...

PR #4592, opened 2024-01-02
👷 Move build chain to ESM

dubzzz:
**⚠️ Minor breaking change**

The build chain of `fast-check` has been CommonJS-based since day 1. With ESM moving forward in the ecosystem, it's time to move ourselves to the new standard and adapt our build chains to ESM.

Unfortunately it may have some subtle impacts on our users as our package will not be a CJS one offering a ESM fallback anymore. I will rather be the opposite: an ESM package with a fallback to CJS. It implies that we moved ESM related files closer to the root of the package (we could have kept them in esm/) and moved the CJS ones further in the file structure (we had to move them).

Another subtle impact is that it would impose our users to run at least Node ≥12.17.0.

As such we consider it as a breaking change. On its own it should not be a huge problem for most of the users but given it changes our minimal requirement let's put it into the next major release.

**_Category:_**

- [ ] ✅ Add or update tests
- [ ] 🐛 Fix a bug
- [ ] 🏷️ Add or update types
- [ ] ⚡️ Improve performance
- [ ] _Other(s):_ ...
  
  

**_Potential impacts:_**

- [ ] Generated values
- [ ] Shrink values
- [ ] Performance
- [ ] Typings
- [ ] _Other(s):_ ...


packages/fast-check/jest.setup.js
```diff
@@ -1,5 +1,6 @@
-const process = require('process');
-const fc = require('fast-check');
+import process from 'process';
+import fc from 'fast-check';
```
dubzzz, line 2: I suspect an issue in Jest with ESM not working the same way in setup files. As if setup files were not sharing the state with the tests themselves. I'll have to confirm that hypothesis on an hello-world example and report it to Jest if confirmed.

dubzzz:
🟢 e2e tests

PR #4709, opened 2024-02-17
Make `fc.pre` an assertion function

gruhn:
By making `fc.pre` an assertion function TypeScript can narrow types based on the condition. For example:

    const item: number | undefined = numberArray[randomIndex]

    fc.pre(item !== undefined)

    item // narrowed type: number

If `fc.pre` is not an assertion function, we have to do an additional (redundant) check to convince TypeScript that `item` is not `undefined`.

See: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions

I marked this change as "decline" but I'm not sure what's appropriate. There is no runtime impact and the API hasn't really changed either. The return type is technically still `void`.

**_Category:_**

- [ ] ✅ Add or update tests
- [ ] 🐛 Fix a bug
- [x] 🏷️ Add or update types
- [ ] ⚡️ Improve performance
- [ ] _Other(s):_ ...
  
  

**_Potential impacts:_**

- [ ] Generated values
- [ ] Shrink values
- [ ] Performance
- [x] Typings
- [ ] _Other(s):_ ...

### DIFF
--- a/.yarn/versions/bbb17765.yml
+++ b/.yarn/versions/bbb17765.yml
@@ -1,0 +1,2 @@
+releases:
+  fast-check: minor
--- a/packages/fast-check/src/check/precondition/Pre.ts
+++ b/packages/fast-check/src/check/precondition/Pre.ts
@@ -6,7 +6,7 @@ import { PreconditionFailure } from './PreconditionFailure';
  * @remarks Since 1.3.0
  * @public
  */
-export function pre(expectTruthy: boolean): void {
+export function pre(expectTruthy: boolean): asserts expectTruthy {
   if (!expectTruthy) {
     throw new PreconditionFailure();
   }


.yarn/versions/a10468f6.yml
```diff
@@ -0,0 +1,2 @@
+declined:
```
dubzzz, line 1: Should be marked as a minor bump

dubzzz:
Sounds to be a really good idea. Thanks for the PR

.yarn/versions/a10468f6.yml
gruhn: done

dubzzz:
Everything is OK on my end. I'm just waiting a little bit to merge it. I have to decide how breaking it will be... In other words : Is it a minor or major? For now, I think so but I have to be sure.

I promise you it's gonna merged soon either for next minor or for next major (currently working on packing everything for it).

gruhn:
no stress ❤️

dubzzz:
Will be available in a few minutes in 3.16.0

PR #4714, opened 2024-02-18
♻️ Rewrite of `scheduleSequence`

dubzzz:
**👀 Potentially risky**

Our initial implementation of `scheduleSequence` used to be reduce-based. It makes it hard to understand at first glance. As such and in attempt to ease our aim to make scheduling faster (see #3891), we are doing a _simple_ rewrite of it.

While it should be without any impact, playing with Promises is always risky and can easily drop some extra waiting times or add some. As such this PR will probably wait for a dedicated minor or major to be rollout.

**_Category:_**

- [ ] ✅ Add or update tests
- [ ] 🐛 Fix a bug
- [ ] 🏷️ Add or update types
- [ ] ⚡️ Improve performance
- [ ] _Other(s):_ ...
  
  

**_Potential impacts:_**

- [ ] Generated values
- [ ] Shrink values
- [ ] Performance
- [ ] Typings
- [ ] _Other(s):_ ...

### DIFF
--- a/packages/fast-check/src/arbitrary/_internals/implementations/SchedulerImplem.ts
+++ b/packages/fast-check/src/arbitrary/_internals/implementations/SchedulerImplem.ts
@@ -157,37 +157,34 @@ export class SchedulerImplem<TMetaData> implements Scheduler<TMetaData> {
     let resolveSequenceTask = () => {};
     const sequenceTask = new Promise<void>((resolve) => (resolveSequenceTask = resolve));
 
-    sequenceBuilders
-      .reduce((previouslyScheduled: PromiseLike<any>, item: SchedulerSequenceItem<TMetaData>) => {
-        const [builder, label, metadata] =
-          typeof item === 'function' ? [item, item.name, undefined] : [item.builder, item.label, item.metadata];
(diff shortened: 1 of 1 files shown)


packages/fast-check/src/arbitrary/_internals/implementations/SchedulerImplem.ts
```diff
@@ -157,37 +157,34 @@ export class SchedulerImplem<TMetaData> implements Scheduler<TMetaData> {
...
+    const onDone = () => {
+      status.done = true;
+      resolveSequenceTask();
+    };
+    const onFaulty = () => {
+      status.faulty = true;
+      resolveSequenceTask();
+    };
+
+    let previouslyScheduled = dummyResolvedPromise;
```
dubzzz, line 169: Used to be the initial version received by the reduce function. In other words, the first value of `previouslyScheduled` was `dummyResolvedPromise` (see https://github.com/dubzzz/fast-check/pull/4714/files#diff-dfe3a426c4b614c4cc2b0327598e960f3175233178422e0a3e3acaf331df50f6L180).

PR #4715, opened 2024-02-18
🐛 Better rejection handling in `scheduleSequence`

dubzzz:
**👀 Potentially risky** (risk: low)

The PR #4714 started to rewrite part of the method to make it easier to evolve. In its original implementation each scheduled item was defining its own error handling on the side and a global error handing was responsible to capture any last crash.

This new approach puts the local item-related error handling in the then-sequence and not forked from it.

In theory, it should not break anything, but as it plays with Promises we can have unwantingly added some extra then step or dropped one. As such it should probably not be released outside of a minor or major version.

**_Category:_**

- [ ] ✅ Add or update tests
- [ ] 🐛 Fix a bug
- [ ] 🏷️ Add or update types
- [ ] ⚡️ Improve performance
- [ ] _Other(s):_ ...
  
  

**_Potential impacts:_**

- [ ] Generated values
- [ ] Shrink values
- [ ] Performance
- [ ] Typings
- [ ] _Other(s):_ ...

### DIFF
--- a/packages/fast-check/src/arbitrary/_internals/implementations/SchedulerImplem.ts
+++ b/packages/fast-check/src/arbitrary/_internals/implementations/SchedulerImplem.ts
@@ -157,24 +157,24 @@ export class SchedulerImplem<TMetaData> implements Scheduler<TMetaData> {
     let resolveSequenceTask = () => {};
     const sequenceTask = new Promise<void>((resolve) => (resolveSequenceTask = resolve));
 
-    const onDone = () => {
-      status.done = true;
-      resolveSequenceTask();
-    };
-    const onFaulty = () => {
+    const onFaultyItemNoThrow = () => {
       status.faulty = true;
       resolveSequenceTask();
     };
-    const onFaultyNoop = () => {
-      /* Discarding UnhandledPromiseRejectionWarning */
-      /* No need to call resolveSequenceTask as it should already have been triggered */
+    const onFaultyItem = (error: unknown) => {
+      onFaultyItemNoThrow(); // Faulty must resolve sequence as soon as possible without any extra then
+      throw error; // Faulty must stay faulty to avoid calling next items
+    };
+    const onDone = () => {
+      status.done = true;
+      resolveSequenceTask();
     };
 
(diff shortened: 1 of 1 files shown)


packages/fast-check/src/arbitrary/_internals/implementations/SchedulerImplem.ts
```diff
@@ -184,12 +184,16 @@ export class SchedulerImplem<TMetaData> implements Scheduler<TMetaData> {
           customAct || defaultSchedulerAct,
           () => builder(),
         );
-        scheduled.catch(onFaulty);
         return scheduled;
       };
-      previouslyScheduled = previouslyScheduled.then(onNext);
+      // We will run current item if and only if the one preceeding it succeeds
+      // Otherwise, we mark the run as "failed" and ignore any subsequent item (including this one)
+      previouslyScheduled = previouslyScheduled.then(onNextItem, onFaultyItem);
```
dubzzz, line 191: The `then` explained:

- success: When item "N-1" is success, we start item "N"
- failure: When item "N-1" (or before) is failure, we mark the sequence as failed and cancel any subsequent item (by throwing)

packages/fast-check/src/arbitrary/_internals/implementations/SchedulerImplem.ts
```diff
@@ -184,12 +184,16 @@ export class SchedulerImplem<TMetaData> implements Scheduler<TMetaData> {
...
     }
-    previouslyScheduled.then(onDone, onFaultyNoop);
+    // Once last item is done, the full sequence can be considered as successful
+    // If it failed (or any preceeding one failed), then the sequence should be marked as failed (already marked for others)
+    // We always handle Promise rejection of previouslyScheduled internally, in other words no error will bubble outside
+    previouslyScheduled.then(onDone, onFaultyItemNoThrow);
```
dubzzz, line 196: The `then` explained:

- success: last item done! it's a success for the whole sequence: all items passed
- failure:
  - last item failed: we have to mark the sequence as failed
  - another one (throw): we ignore the error (it has already been marked and resolved)

PR #4717, opened 2024-02-18
⚡️ Faster scheduling of `scheduleSequence`

dubzzz:
**👀 Potentially risky** (risk: low)

With #3891, our target is to make scheduling faster. Scheduling is one of the built-in arbitrary provided by fast-check. It aims to ease detection of race-conditions by offering a simple way to detect such issues in your code.

Up-to-now the code was doing a bit too many await/then letting it give back the hand many times in a row while it could have been faster to take it back. Additionally, this past behaviour (that we are trying to fix with many PRs) used to be strange as awaiting 1 was ok, but 10-times was not. More precisely there was a number of await working and passed this number it stopped to work as usual and users started to need to use other approaches (drop the `waitAll`).

This set of PRs not only addresses a potential perf concern but also clarify what should work so that it appears more determistic for the final users.

**_Category:_**

- [ ] ✅ Add or update tests
- [ ] 🐛 Fix a bug
- [ ] 🏷️ Add or update types
- [x] ⚡️ Improve performance
- [ ] _Other(s):_ ...
  
  

**_Potential impacts:_**

- [ ] Generated values
- [ ] Shrink values
- [x] Performance
- [ ] Typings
- [ ] _Other(s):_ ...

### DIFF
--- a/packages/fast-check/src/arbitrary/_internals/implementations/SchedulerImplem.ts
+++ b/packages/fast-check/src/arbitrary/_internals/implementations/SchedulerImplem.ts
@@ -155,27 +155,30 @@ export class SchedulerImplem<TMetaData> implements Scheduler<TMetaData> {
     // Placeholder resolver, immediately replaced by the one retrieved in `new Promise`
     // eslint-disable-next-line @typescript-eslint/no-empty-function
     let resolveSequenceTask = () => {};
-    const sequenceTask = new Promise<void>((resolve) => (resolveSequenceTask = resolve));
+    const sequenceTask = new Promise<{ done: boolean; faulty: boolean }>((resolve) => {
+      resolveSequenceTask = () => resolve({ done: status.done, faulty: status.faulty });
+    });
 
     const onFaultyItemNoThrow = () => {
       status.faulty = true;
       resolveSequenceTask();
-    };
-    const onFaultyItem = (error: unknown) => {
-      onFaultyItemNoThrow(); // Faulty must resolve sequence as soon as possible without any extra then
-      throw error; // Faulty must stay faulty to avoid calling next items
     };
     const onDone = () => {
       status.done = true;
       resolveSequenceTask();
     };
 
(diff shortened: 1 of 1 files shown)


packages/fast-check/src/arbitrary/_internals/implementations/SchedulerImplem.ts
```diff
@@ -155,27 +155,30 @@ export class SchedulerImplem<TMetaData> implements Scheduler<TMetaData> {
     // Placeholder resolver, immediately replaced by the one retrieved in `new Promise`
     // eslint-disable-next-line @typescript-eslint/no-empty-function
     let resolveSequenceTask = () => {};
-    const sequenceTask = new Promise<void>((resolve) => (resolveSequenceTask = resolve));
+    const sequenceTask = new Promise<{ done: boolean; faulty: boolean }>((resolve) => {
+      resolveSequenceTask = () => resolve({ done: status.done, faulty: status.faulty });
```
dubzzz, line 159: We resolve immediately with the value without an extra `.then` step.

packages/fast-check/src/arbitrary/_internals/implementations/SchedulerImplem.ts
```diff
@@ -155,27 +155,30 @@ export class SchedulerImplem<TMetaData> implements Scheduler<TMetaData> {
...
       resolveSequenceTask();
     };
 
-    let previouslyScheduled = dummyResolvedPromise;
-    for (const item of sequenceBuilders) {
-      const [builder, label, metadata] =
-        typeof item === 'function' ? [item, item.name, undefined] : [item.builder, item.label, item.metadata];
-      const onNextItem = () => {
-        // We schedule a successful promise that will trigger builder directly when triggered
+    const registerNextBuilder = (index: number, previous: PromiseLike<unknown>) => {
```
dubzzz, line 171: This recursive code drops some unwanting awaits compared to the previous approach. As such the execution should be sightly faster with less thenables cutting the execution.

PR #4842, opened 2024-03-27
internalise private stuff, fixing dual types incompatibility

### LINKED ISSUE #4519: Problem when mixing local project moduleResolution nodenext and a 3rd party package using fast-check
## 🐛 Bug Report

```
Argument of
type 'typeof import(\"pj/effect-app/libs/node_modules/fast-check/lib/types/fast-check\")' is not assignable to parameter of
type 'typeof import(\"pj/effect-app/libs/node_modules/fast-check/lib/esm/types/fast-check\", { with: { \"resolution-mode\": \"import\" } })'
```

## Expected behavior

No error. probably there should only be one set of d.ts (unless there are some reasons there can't be):
```diff
diff --git a/CHANGELOG.md b/CHANGELOG.md
deleted file mode 100644
index f272c80981f45dc84797c50af97fe5d2348fd188..0000000000000000000000000000000000000000
diff --git a/package.json b/package.json
index 8accb7fb8662a36944083994c78e65a2e3f32a53..7f3548d14423de586bf8e1bdc99a5ad77bb05a3e 100644
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
         "default": "./lib/fast-check.js"
       },
       "import": {
-        "types": "./lib/esm/types/fast-check.d.ts",
+        "types": "./lib/types/fast-check.d.ts",
         "default": "./lib/esm/fast-check.js"
       }
     }
```
this works great so far.

## Your environment

| Packages / Softwares | Version(s) |
| -------------------- | ---------- |
| fast-check           |    3.14        |
| node                 |     18.10.0       |
| TypeScript\*         |    5.3.2        |

\*Only for TypeScript's users

patroza:
fixes #4519, #4354

### DIFF
--- a/.yarn/versions/1488efe6.yml
+++ b/.yarn/versions/1488efe6.yml
@@ -1,6 +1,8 @@
 releases:
-  "@fast-check/ava": patch
-  "@fast-check/jest": patch
-  "@fast-check/vitest": patch
-  "@fast-check/worker": patch
   fast-check: patch
+
+declined:
+  - "@fast-check/ava"
+  - "@fast-check/jest"
+  - "@fast-check/vitest"
+  - "@fast-check/worker"
--- a/.yarn/versions/1488efe6.yml
+++ b/.yarn/versions/1488efe6.yml
@@ -1,2 +1,6 @@
 releases:
+  "@fast-check/ava": patch
+  "@fast-check/jest": patch
+  "@fast-check/vitest": patch
+  "@fast-check/worker": patch
   fast-check: patch
--- a/.yarn/versions/1488efe6.yml
+++ b/.yarn/versions/1488efe6.yml
@@ -1,0 +1,2 @@
+releases:
+  fast-check: patch
--- a/packages/fast-check/src/arbitrary/_internals/helpers/DepthContext.ts
+++ b/packages/fast-check/src/arbitrary/_internals/helpers/DepthContext.ts
@@ -9,6 +9,7 @@ declare const depthIdentifierSymbol: unique symbol;
  * what they contain internally
  *
  * @remarks Since 2.25.0
+ * @public
  */
 export type DepthIdentifier = {
   /** @internal */
--- a/packages/fast-check/src/arbitrary/_internals/helpers/DepthContext.ts
+++ b/packages/fast-check/src/arbitrary/_internals/helpers/DepthContext.ts
@@ -9,9 +9,11 @@ declare const depthIdentifierSymbol: unique symbol;
  * what they contain internally
  *
  * @remarks Since 2.25.0
- * @internal
  */
-export type DepthIdentifier = { [depthIdentifierSymbol]: true };
+export type DepthIdentifier = {
+  /** @internal */
+  [depthIdentifierSymbol]: true
+};
 
 /**
  * Instance of depth, can be used to alter the depth perceived by an arbitrary
--- a/packages/fast-check/src/arbitrary/_internals/helpers/DepthContext.ts
+++ b/packages/fast-check/src/arbitrary/_internals/helpers/DepthContext.ts
@@ -1,6 +1,6 @@
 /**
  * Internal symbol used to declare an opaque type for DepthIdentifier
- * @public
+ * @internal
  */
 declare const depthIdentifierSymbol: unique symbol;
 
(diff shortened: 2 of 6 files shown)


dubzzz:
Thanks for the help, let's see if it passes the CI : 🤞

.yarn/versions/1488efe6.yml
```diff
@@ -0,0 +1,6 @@
+releases:
+  "@fast-check/ava": patch
+  "@fast-check/jest": patch
+  "@fast-check/vitest": patch
+  "@fast-check/worker": patch
```
dubzzz, line 5: You can mark non-fast-check as ignored

dubzzz: I mean all @ should be ignored/declined, they are not impacted by your change

packages/fast-check/src/check/arbitrary/definition/Value.ts
```diff
@@ -18,6 +18,7 @@ export class Value<T> {
    */
   readonly hasToBeCloned: boolean;
   /**
+   * @internal
```
dubzzz, line 21: This one seems to be on an unrelated field? 🤔

patroza:
`DepthIdentifier` is a useless type now because it's only member, the symbol is now internal.
Is this type important to be public?

packages/fast-check/src/check/arbitrary/definition/Value.ts
patroza: any private field or symbol on the public interface makes the types incompatible

patroza:
Looks like all pass except doc publishing due to no permission

dubzzz:
Seems green ✅

I'll just checkout the resulting package tomorrow morning to see how it behaves

packages/fast-check/src/check/precondition/PreconditionFailure.ts
```diff
@@ -4,7 +4,9 @@
  * @public
  */
 export class PreconditionFailure extends Error {
+  /** @internal */
   private static readonly SharedFootPrint: symbol = Symbol('fast-check/PreconditionFailure');
```
dubzzz, line 8: Probably worth changing it into a Symbol.for so that we make sure that CJS and MJS world can interoperate between each others.

patroza: private static is not on the exported interface

patroza: sadly symbol.for is also no fix;
```ts
const a = Symbol.for("a")
const b = Symbol.for("a")
const c = (_: typeof b) => 1

c(a)
//^ Argument of type 'typeof a' is not assignable to parameter of type 'typeof b'.
```

patroza: (sorry, my previous message about not being on the interface was incorrect. it's not on there because of @internal :D

packages/fast-check/src/arbitrary/_internals/helpers/DepthContext.ts
```diff
@@ -11,7 +11,10 @@ declare const depthIdentifierSymbol: unique symbol;
  * @remarks Since 2.25.0
  * @public
  */
-export type DepthIdentifier = { [depthIdentifierSymbol]: true };
+export type DepthIdentifier = {
```
dubzzz, line 14: With the suggested approach users will be able to replace instances of `DepthIdentifier` by raw `{}`. While I'm ok with relaxing the constraint for now, I'd rather put: `export type DepthIdentifier = {/* the symbol thing */} & DepthContext`.

Here is a repro of the problem: https://github.com/dubzzz/fast-check-type-hazard-repro/blob/main/main.ts#L14

Line 14 compiles with your suggestion. Ideally I'd at least force the users to come with identifiers matching context type. In other words, for end-users it will result into Identifier type is just Context (with another name), but I'm totally ok with that (until dual goes away).

packages/fast-check/src/check/precondition/PreconditionFailure.ts
patroza: yea separate would be great.
the current implementation is tested and works in the @effect/schema case

dubzzz:
> DepthIdentifier is a useless public type now because it's only member, the symbol is now internal.
Is this type important to be public and exported? An alternative could be a string literal.

Useless for now 😢 I hope to put it back once I can move back to full ESM. Hope that recent changes around CJS-MJS interrupt will help me to only ship ESM version in a few years... At that point in time I might re-enable that opaque type. For now it only helps me internally and may still be useful for users so that they reference the future opaque type in their typings (even if, nothing is opaque in it 😂).

dubzzz:
🚀 It's live on 3.17.1

PR #4871, opened 2024-04-07
✅ Move tests to Vitest

dubzzz:
My target is to be able to make fast-check ES Modules first.

I initially attempted to stick to Jest, but I failed. More precisely having Jest in ES Modules mode required: experimental flags, usage of unstable APIs, using arrow imports instead of normal ones, migrating out of spies and relying on requireActual while I'd need to importActual (aka missing API).

Unfortunately I was not able to reach an acceptable state. he main blocker being the arrow import that would impose to add some linting rule to make sure that all my tests do only arrow imports. The next blocker was rewriting all the spies to move them to mock and spies.

So I got directed to Vitest. I tried and spies were ok, import too.

**_Category:_**

- [ ] ✅ Add or update tests
- [ ] 🐛 Fix a bug
- [ ] 🏷️ Add or update types
- [ ] ⚡️ Improve performance
- [ ] _Other(s):_ ...
  
  

**_Potential impacts:_**

- [ ] Generated values
- [ ] Shrink values
- [ ] Performance
- [ ] Typings
- [ ] _Other(s):_ ...


dubzzz:
From my first measurements, with the same coverage reporter in place for both Jest and Vitest, Vitest is 33% slower. But I hope to make it faster with some cleaning and no globals mode.

PR #4924, opened 2024-04-24
🔧 Enable more advanced TS flags

dubzzz:
Based on suggestions from https://www.totaltypescript.com/tsconfig-cheat-sheet

### DIFF
--- a/tsconfig.common.json
+++ b/tsconfig.common.json
@@ -13,7 +13,6 @@
     "skipLibCheck": true,
     "strict": true,
     "moduleDetection": "force",
-    "isolatedModules": true,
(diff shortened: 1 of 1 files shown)


tsconfig.common.json
```diff
@@ -11,6 +11,9 @@
     "noFallthroughCasesInSwitch": true,
     "noUnusedLocals": true,
     "skipLibCheck": true,
-    "strict": true
+    "strict": true,
+    "moduleDetection": "force",
+    "isolatedModules": true,
+    "verbatimModuleSyntax": true
```
dubzzz, line 17:
```suggestion
    "isolatedModules": true
```

PR #4981, opened 2024-05-14
🐛 Better interrupt between multiple versions

### LINKED ISSUE #4845: Better interrupt CJS-ESM as we provide both in our bundle (Symbol.for switch)
## 🐛 Bug Report

Related to https://github.com/dubzzz/fast-check/pull/4842

Until we fully drop dual package, we should probably switch back to `Symbol.for` to drop risky things linked to symbols not being the same in different realms.

dubzzz:
fast-check was badly behaving on cloning method, custom toString and assertions linked to pre when several versions of it were running at the same time.

One of the case for that issue is ES module and CommonJS being mixed in the same test and leading fast-check to be loaded once in ESM and once in CJS.

Another case is that user relies on two versions of fast-check due to a dependency forcing a version that is not the same as the one requested by the user or by another dependency (not the same meaning not compatible semver ranges).

Fixes #4845

**_Category:_**

- [ ] ✅ Add or update tests
- [ ] 🐛 Fix a bug
- [ ] 🏷️ Add or update types
- [ ] ⚡️ Improve performance
- [ ] _Other(s):_ ...
  
  

**_Potential impacts:_**

- [ ] Generated values
- [ ] Shrink values
- [ ] Performance
- [ ] Typings
- [ ] _Other(s):_ ...

### DIFF
--- a/packages/fast-check/test/unit/check/symbols.spec.ts
+++ b/packages/fast-check/test/unit/check/symbols.spec.ts
@@ -3,7 +3,7 @@ import { cloneMethod, hasCloneMethod } from '../../../src/check/symbols';
 import * as fc from 'fast-check';
 
 describe('symbols', () => {
-  it('should declare indentical cloneMethod for distinct instances of fast-check', () => {
+  it('should declare identical cloneMethod for distinct instances of fast-check', () => {
     expect(cloneMethod).toBe(fc.cloneMethod);
   });
 });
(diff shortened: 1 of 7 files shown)


packages/fast-check/test/unit/check/symbols.spec.ts
```diff
@@ -3,8 +3,8 @@ import { cloneMethod, hasCloneMethod } from '../../../src/check/symbols';
 import * as fc from 'fast-check';
 
 describe('symbols', () => {
-  it('should declare distinct cloneMethod for distinct libraries', () => {
-    expect(cloneMethod).not.toBe(fc.cloneMethod);
+  it('should declare indentical cloneMethod for distinct instances of fast-check', () => {
```
dubzzz, line 6:
```suggestion
  it('should declare identical cloneMethod for distinct instances of fast-check', () => {
```

Issue #4162, opened 2023-08-26
Allow specifying max number of shrinks per test


AlexeyRaga:
## 🚀 Feature Request

Allow specifying max number of shrinks.
The idea is not new, other property testing framework (`QuickCheck`,`HedgeHog`, `FSCheck` in Haskell and FSharp have it)

## Motivation

Performance and friendliness with integration tests.
Also related to https://github.com/dubzzz/fast-check/issues/4161.

## Example

```typescript
fc.assert(myProperty, { numRuns: 5, maxShrinks: 5 });
```

dubzzz:
Any example of such option in the other libraries? How it can be configured compared to another test...?

On it's own reducing shrinker can be done at various level: the arbitrary itself (what we do with `noShrink`) vs the runner (what you suggest). It can also be done at different levels: how far it goes in a given level? how many levels it can run into? As actually the shrinker runs on a 2D-plan so we may want to limit one direction and not the other or a combination of both of them.

Having more context would help as it seems that you know how they do it might reduce the exploratory time required just to put a status (do/don't do) on such ticket.

---

Side-note: You may use `noShrink` or `endOnFailure`.

Like:

```ts
fc.uuid().noShrink()
```

or:

```ts
fc.assert(myProperty, { numRuns: 5, endOnFailure: true })
```

AlexeyRaga:
@dubzzz sure,

`Hedgehog`: https://hackage.haskell.org/package/hedgehog-1.4/docs/Hedgehog.html#v:withShrinks
`QuickCheck` in Haskell: https://hackage.haskell.org/package/QuickCheck-2.14.3/docs/Test-QuickCheck.html#t:Args

Alternatives are not really the same because I do not really want to give up on shrinking...

dubzzz:
As we already have `noShrink`, I think I'll try to find an option that could cover both `noShrink` and `maxShrinks` at the same time. Probably re-using `noShrink` but passing it a number for v3 and move to another name for v4.

The remaining part to decide is: what to stop shrinking as shrinker lives in a 2D-plan so 2 directions can be stopped separately or together.

dubzzz:
Maybe the existing `noShrink` should be replaced by some kind of `limitShrink({ maxDepth, maxValuesPerLevel })` or a simpler one `limitShrink(maxValues)` with `maxValues` somehow aggregating the two others

dubzzz:
[...]

PR #5006, opened 2024-05-21
✨ Add ability to limit shrink path

dubzzz:
**Description**

We want to offer a way for our users to limit the depth and overall size of the shrinking capabilities. At the moment, they can only either apply shrink or drop it totally. We want to offer a finer way to control the shrinks.

This proposal is related to #4162.

At the moment, we added an extra method for all our arbitrary but I feel that it will not be scalable. I believe we should go for yet another utility arbitrary that will be responsible to offer such capabilities. Regarding offering these capabilities at assert level itself, we should probably (option 1) do as we used to do for timeouts..., or (option 2) start thinking into some kind of plugin system. Option 1 seems to be a valid option for version 3 of fast-check, we could also start offering a noShrink utility either hidden behind limitShrink one (probably better) or directly.

**NOTE:** Not ready yet, we should move from methods to a dedicated utility.

**Advanced**

- [x] Impacts: New arbitrary


packages/fast-check/src/arbitrary/limitShrink.ts
```diff
@@ -0,0 +1,35 @@
...
+ * const dataGenerator: Arbitrary<string> = ...;
+ * const limitedShrinkableDataGenerator: Arbitrary<string> = fc.limitShrink(dataGenerator, { maxShrinks: 10 });
+ * // up to 10 shrunk values could be extracted from the resulting arbitrary
+ * ```
+ *
+ * NOTE: Although limiting the shrinking capabilities can speed up your CI when failures occur, we do not recommend this approach.
+ * Instead, if you want to reduce the shrinking time for automated jobs or local runs, consider using `endOnFailure` or `interruptAfterTimeLimit`.
+ *
+ * @returns Create another arbitrary with limited number of shrink values
+ * @remarks Since 3.20.0
```
dubzzz, line 31:
```suggestion
 * @remarks Since 3.20.0
 * @public
```

PR #5581, opened 2025-01-08
🔥 Enforce `run{Before/After}Each` on property

dubzzz:
**Description**

The API of property has been updated in a non-breaking manner during v3. Now, let's seal that move to drop unneeded code complexity, drop useless tests... but also make the API clearer and way simpler to use and manage.

The implementations of `IRawProperty` now have to implement `runBeforeEach` and `runAfterEach`. If they themselves wrap another `IRawProperty`, they probably should at least call these methods when called on their side of them. The migration from v3, consists into implementing the two methods.

**Advanced**

**:boom: Breaking change**: Change the execution flow of properties. Only impacts on advanced users building their own custom runners or their own property type.

### DIFF
--- a/.changeset/wise-pears-lay.md
+++ b/.changeset/wise-pears-lay.md
@@ -1,6 +1,5 @@
 ---
 "fast-check": major
-"@fast-check/worker": patch
 ---
 
 🔥 Enforce `run{Before/After}Each` on property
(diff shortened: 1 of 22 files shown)


packages/worker/test/internals/worker-property/WorkerPropertyFromWorker.spec.ts
```diff
@@ -31,7 +31,7 @@ describe('WorkerPropertyFromWorker', () => {
     const property = new WorkerPropertyFromWorker(arbitraries, predicate);
     const value = property.generate(mrng, 0);
     property.runBeforeEach();
-    property.run(value.value_, true);
+    property.run(value.value_);
```
dubzzz, line 34:
```suggestion
    property.run(value.value_, true);
```

.changeset/wise-pears-lay.md
```diff
@@ -0,0 +1,6 @@
+---
+"fast-check": major
+"@fast-check/worker": patch
```
dubzzz, line 3:
```suggestion
```

Issue #4416, opened 2023-11-08
Error with cause by default for v4


dubzzz:
Check node, jest, vitest, mocha, jasmine... for support to drop or not the legacy code or just part of it.

dubzzz:
**Node support:** ≥16.14 (8th February 2022)

🟢 16.14.2 / 16.15.1 / 16.20.2 / 18.0.0 / 18.18.2 / 20.10.0
🔴 12.22.12 / 14.21.3 / 16.0.0 / 16.10.0 / 16.13.2

<details>
<summary>Code snippet</summary>

```js
function g() {
  throw new Error("g failed");
}

function f() {
  try {
    g();
  } catch (err) {
    throw new Error("f failed", { cause: err });
  }
}

f();
```
</details>

PR #5590, opened 2025-01-10
💥 Error with cause by default

dubzzz:
**Description**

Until now, fast-check have been merging the content and stack of the original Error that caused the property to fail into its own Error. With the recent (2022) introduction of cause on Errors, this complex logic can be dropped in favor of the native cause mechanism.

This PR makes cause mode the default. Before this PR, toggling it was possible via `errorWithCause: true` on `fc.assert`.

Given not all test runners properly support causes attached to the Error, we offer a fallback for users willing to preserve the old behaviour. It can be toggled via `includeErrorInReport: true` on `fc.assert`.

Related to #4416.

**Advanced**

**💥 Breaking change**: The display of errors being reported by fast-check will change. Many test runners do support that but some still don't support it so a flag will have to be pass for them to display the errors correctly.

### DIFF
--- a/website/docs/migration/from-3.x-to-4.x.md
+++ b/website/docs/migration/from-3.x-to-4.x.md
@@ -50,6 +50,20 @@ fc.record(recordModel, {
 
 Related pull requests: [#5578](https://github.com/dubzzz/fast-check/pull/5578)
 
+## Update to latest v4.x
+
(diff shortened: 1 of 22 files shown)


website/docs/migration/from-3.x-to-4.x.md
```diff
@@ -50,6 +50,20 @@ fc.record(recordModel, {
 
 Related pull requests: [#5578](https://github.com/dubzzz/fast-check/pull/5578)
 
+## Update to latest v4.x
```
dubzzz, line 53:
```suggestion
## Update to v4.x
```

PR #5618, opened 2025-01-14
⚡️ Faster rewrite of `double`

dubzzz:
**Description**

When we first came up with the `double` arbitrary, `bigint` were not mature enough in JavaScript. As such, we did not want our users to be forced to bump to the very last versions of Node and polyfilled it with out own 64bits bigint implementation.

While it worked well and in an efficient way, it requires extra code and is probably less efficient (after further optimizations) than a native `bigint`. As such this PR drops it in favor of a `bigint`-based internal.

What changes for the user? Well generated values, the shrinker. And for users blocked on legacy releases of Node (<10.5 or 10.4?), they might just not be able to use this major version of fast-check.

**Advanced**

- [x] Category: ⚡️ Improve performance
- [x] Impacts: Generated values, shrinker, minimal requirements for Node


dubzzz:
Benchmark at: https://github.com/dubzzz/fast-check-benchmarks/actions/runs/13190276746

PR #5892, opened 2025-04-05
✨ Better capture scheduled tasks before running scheduling

dubzzz:
**Description**

We want our users never to worry about how we internally deal with waiting for promises to be scheduled. Given the fact there is no API (in JS) available to know if they are pending promises awaiting, we are empirically letting a few ticks running before closing the set of promises that we consider for a batch.

This PR makes sure that anytime a new promise gets added for a batch, we reset our window to wait a bit longer. It makes us safer on the fast that we are gonna catch them all.

**Advanced**

- [x] Impacts: Scheduler can schedule more tasks within the same batch, it implies that we may eventually find way more race conditions than what we used to detect. In practice this case will probably never be triggered but we prefer having it covered.

### DIFF
--- a/packages/fast-check/test/unit/arbitrary/_internals/implementations/SchedulerImplem.spec.ts
+++ b/packages/fast-check/test/unit/arbitrary/_internals/implementations/SchedulerImplem.spec.ts
@@ -828,7 +828,7 @@ describe('SchedulerImplem', () => {
       await process;
     });
 
-    it('should schedule known Promises as quickly as possible when depending on promise ticks', async () => {
+    it('should try to wait enough ticks for all Promises to be registered before scheduling any', async () => {
       // Arrange
       const awaitedTask = buildUnresolved();
       const seenTasks: unknown[][] = [];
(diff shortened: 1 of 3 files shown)


packages/fast-check/test/unit/arbitrary/_internals/implementations/SchedulerImplem.spec.ts
```diff
@@ -862,23 +862,128 @@ describe('SchedulerImplem', () => {
 
       // Assert
       expect(processFlags).toEqual(expectedFlags);
-      const batchSize = numTicksBeforeScheduling;
-      let expectedCount = batchSize + 1;
       for (let i = 0; i !== expectedFlags.length; ++i) {
         // Scheduled promises will be partially grouped into common batches, the process we play with does fire in parallel:
```
dubzzz, line 866:
```suggestion
        // Scheduled promises will be grouped into common batches, the process we play with does fire in parallel:
```

PR #6040, opened 2025-07-07
✨ Add `circular` option to `fc.letrec`

### LINKED ISSUE #6039: Add `circular: true` option to `fc.letrec`
## 🚀 Feature Request

After this [discussion](https://github.com/dubzzz/fast-check/discussions/6012) I'm thinking it'd be cool to add a `circular: true` option to `fc.letrec` that automatically reuses objects.

I'm interested in implementing this feature myself!

## Motivation

See https://github.com/dubzzz/fast-check/discussions/6012

## Example

```ts
fc.letrec(
  tie => ({
    nodeArb: fc.record({
      value: fc.nat(),
      next: tie('nodeArb'),
    }),
  }),
  { circular: true }
)
```

TomerAberbach:
**Description**

Closes #6039

- [ ] My PR adds relevant tests and they would have failed without my PR (when applicable)

**Advanced**

- [x] Category: ✨ Introduce new feature
- [x] Impacts: N/A

---

Open questions:
1. Is this really the best place/way to implement this? (see my comments on #6039)
2. See "// TODO: Do we need to clone here?" in the PR. If I _do_ need to clone, then what should I be using to do that?
3. Neither `assertProduceValuesShrinkableWithoutContext` nor `assertShrinkProducesSameValueWithoutInitialContext` pass for `circular: true`. Does that make sense? Should they pass? If so, then what am I doing wrong?

### DIFF
--- a/packages/fast-check/src/arbitrary/letrec.ts
+++ b/packages/fast-check/src/arbitrary/letrec.ts
@@ -1,8 +1,13 @@
 import { LazyArbitrary } from './_internals/LazyArbitrary';
 import type { Arbitrary } from '../check/arbitrary/definition/Arbitrary';
 import { safeHasOwnProperty } from '../utils/globals';
+import { nat } from './nat.js';
+import { record } from './record.js';
+import { array } from './array.js';
 
+const safeArrayIsArray = Array.isArray;
 const safeObjectCreate = Object.create;
+const safeObjectEntries = Object.entries;
 
 /**
  * Type of the value produced by {@link letrec}
@@ -51,49 +56,20 @@ export type LetrecLooselyTypedTie = (key: string) => Arbitrary<unknown>;
 export type LetrecLooselyTypedBuilder<T> = (tie: LetrecLooselyTypedTie) => LetrecValue<T>;
 
 /**
- * For mutually recursive types
- *
- * @example
- * ```typescript
- * type Leaf = number;
- * type Node = [Tree, Tree];
- * type Tree = Node | Leaf;
- * const { tree } = fc.letrec<{ tree: Tree, node: Node, leaf: Leaf }>(tie => ({
- *   tree: fc.oneof({depthSize: 'small'}, tie('leaf'), tie('node')),
- *   node: fc.tuple(tie('tree'), tie('tree')),
- *   leaf: fc.nat()
- * }));
- * // tree is 50% of node, 50% of leaf
- * // the ratio goes in favor of leaves as we go deeper in the tree (thanks to depthSize)
- * ```
- *
- * @param builder - Arbitraries builder based on themselves (through `tie`)
- *
- * @remarks Since 1.16.0
+ * Constraints to be applied on {@link letrec}
+ * @remarks Since 4.2.0
  * @public
  */
-export function letrec<T>(builder: T extends Record<string, unknown> ? LetrecTypedBuilder<T> : never): LetrecValue<T>;
-/**
- * For mutually recursive types
- *
- * @example
- * ```typescript
- * const { tree } = fc.letrec(tie => ({
- *   tree: fc.oneof({depthSize: 'small'}, tie('leaf'), tie('node')),
- *   node: fc.tuple(tie('tree'), tie('tree')),
- *   leaf: fc.nat()
- * }));
- * // tree is 50% of node, 50% of leaf
- * // the ratio goes in favor of leaves as we go deeper in the tree (thanks to depthSize)
- * ```
- *
- * @param builder - Arbitraries builder based on themselves (through `tie`)
- *
- * @remarks Since 1.16.0
- * @public
- */
-export function letrec<T>(builder: LetrecLooselyTypedBuilder<T>): LetrecValue<T>;
-export function letrec<T>(builder: LetrecLooselyTypedBuilder<T> | LetrecTypedBuilder<T>): LetrecValue<T> {
+export interface LetrecConstraints {
+  /**
+   * Generate objects with circular references
+   * @defaultValue false
+   * @remarks Since 4.2.0
+   */
+  circular?: boolean;
+}
+
+function nonCircularLetrec<T>(builder: LetrecLooselyTypedBuilder<T> | LetrecTypedBuilder<T>): LetrecValue<T> {
   const lazyArbs: { [K in keyof T]?: LazyArbitrary<unknown> } = safeObjectCreate(null);
   const tie = (key: keyof T): Arbitrary<any> => {
     if (!safeHasOwnProperty(lazyArbs, key)) {
@@ -115,4 +91,144 @@ export function letrec<T>(builder: LetrecLooselyTypedBuilder<T> | LetrecTypedBui
     lazyArbs[key] = lazyArb;
   }
   return strictArbs;
+}
+
+function circularLetrec<T>(builder: LetrecLooselyTypedBuilder<T> | LetrecTypedBuilder<T>): LetrecValue<T> {
+  const lazyArbs: { [K in keyof T]?: LazyArbitrary<unknown> } = safeObjectCreate(null);
+  const tie = (key: keyof T): Arbitrary<any> => {
+    if (!safeHasOwnProperty(lazyArbs, key)) {
+      // Call to hasOwnProperty ensures that the property key will be defined
+      lazyArbs[key] = new LazyArbitrary(String(key));
+    }
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    return lazyArbs[key]!;
+  };
+  const strictArbs = builder(tie as any);
+
+  // Symbol to replace with a potentially circular reference later.
+  const placeholderSymbol = Symbol('placeholder');
+  const poolArbs: { [K in keyof T]: Arbitrary<unknown[]> } = safeObjectCreate(null);
+  for (const key in strictArbs) {
+    if (!safeHasOwnProperty(strictArbs, key)) {
+      // Prevents accidental iteration over properties inherited from an object’s prototype
+      continue;
+    }
+    const lazyAtKey: LazyArbitrary<unknown> | undefined = lazyArbs[key];
+    const lazyArb = lazyAtKey !== undefined ? lazyAtKey : new LazyArbitrary(key);
+    lazyArb.underlying = nat().map((index) => ({
+      [placeholderSymbol]: { key, index },
+    }));
+    lazyArbs[key] = lazyArb;
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    poolArbs[key] = array(strictArbs[key]!, { minLength: 1 });
+  }
+
+  for (const key in strictArbs) {
+    if (!safeHasOwnProperty(strictArbs, key)) {
+      // Prevents accidental iteration over properties inherited from an object’s prototype
+      continue;
+    }
+
+    const poolsArb = record(poolArbs as any) as Arbitrary<{ [K in keyof T]: unknown[] }>;
+    strictArbs[key] = poolsArb.map((pools) => {
+      const visited = new WeakSet();
+      function deref(value: unknown, source?: Record<PropertyKey, unknown>, sourceKey?: PropertyKey): unknown {
+        if (typeof value !== 'object' || value === null) {
+          return value;
+        }
+
+        if (visited.has(value)) {
+          return value;
+        }
+        visited.add(value);
+
+        if (safeHasOwnProperty(value, placeholderSymbol)) {
+          const { key, index } = (
+            value as {
+              [placeholderSymbol]: { key: keyof T; index: number };
+            }
+          )[placeholderSymbol];
+          const pool = pools[key];
+          const poolValue = pool[index % pool.length];
+          if (source !== undefined && sourceKey !== undefined) {
+            source[sourceKey] = poolValue;
+            return value;
+          } else {
+            return poolValue;
+          }
+        }
+
+        if (safeArrayIsArray(value)) {
+          for (let i = 0; i < value.length; i++) {
+            deref(value[i], value as unknown as Record<PropertyKey, unknown>, i);
+          }
+        } else {
+          for (const [key, item] of safeObjectEntries(value)) {
+            deref(item, value as Record<PropertyKey, unknown>, key);
+          }
+        }
+
+        return value;
+      }
+
+      // TODO: Do we need to clone here?
+      deref(pools);
+      return pools[key][0];
+    }) as (typeof strictArbs)[typeof key];
+  }
+
+  return strictArbs;
+}
+
+/**
+ * For mutually recursive types
+ *
+ * @example
+ * ```typescript
+ * type Leaf = number;
+ * type Node = [Tree, Tree];
+ * type Tree = Node | Leaf;
+ * const { tree } = fc.letrec<{ tree: Tree, node: Node, leaf: Leaf }>(tie => ({
+ *   tree: fc.oneof({depthSize: 'small'}, tie('leaf'), tie('node')),
+ *   node: fc.tuple(tie('tree'), tie('tree')),
+ *   leaf: fc.nat()
+ * }));
+ * // tree is 50% of node, 50% of leaf
+ * // the ratio goes in favor of leaves as we go deeper in the tree (thanks to depthSize)
+ * ```
+ *
+ * @param builder - Arbitraries builder based on themselves (through `tie`)
+ *
+ * @remarks Since 1.16.0
+ * @public
+ */
+export function letrec<T>(
+  builder: T extends Record<string, unknown> ? LetrecTypedBuilder<T> : never,
+  constraints?: LetrecConstraints,
+): LetrecValue<T>;
+/**
+ * For mutually recursive types
+ *
+ * @example
+ * ```typescript
+ * const { tree } = fc.letrec(tie => ({
+ *   tree: fc.oneof({depthSize: 'small'}, tie('leaf'), tie('node')),
+ *   node: fc.tuple(tie('tree'), tie('tree')),
+ *   leaf: fc.nat()
+ * }));
+ * // tree is 50% of node, 50% of leaf
+ * // the ratio goes in favor of leaves as we go deeper in the tree (thanks to depthSize)
+ * ```
+ *
+ * @param builder - Arbitraries builder based on themselves (through `tie`)
+ *
+ * @remarks Since 1.16.0
+ * @public
+ */
+export function letrec<T>(builder: LetrecLooselyTypedBuilder<T>, constraints?: LetrecConstraints): LetrecValue<T>;
+export function letrec<T>(
+  builder: LetrecLooselyTypedBuilder<T> | LetrecTypedBuilder<T>,
+  constraints: LetrecConstraints = {},
+): LetrecValue<T> {
+  return (constraints.circular ? circularLetrec : nonCircularLetrec)(builder);
 }
--- a/packages/fast-check/test/unit/arbitrary/letrec.spec.ts
+++ b/packages/fast-check/test/unit/arbitrary/letrec.spec.ts
@@ -1,5 +1,6 @@
 import { describe, it, expect } from 'vitest';
 import { letrec } from '../../../src/arbitrary/letrec';
+import { record } from '../../../src/arbitrary/record';
 import { LazyArbitrary } from '../../../src/arbitrary/_internals/LazyArbitrary';
 import { Value } from '../../../src/check/arbitrary/definition/Value';
 import { Stream } from '../../../src/stream/Stream';
@@ -345,5 +346,24 @@ describe('letrec (integration)', () => {
 
   it('should be able to shrink to the same values without initial context (if underlyings do)', () => {
     assertShrinkProducesSameValueWithoutInitialContext(letrecBuilder);
+  });
+});
+
+describe('letrec circular (integration)', () => {
+  const letrecBuilder = () => {
+    const { node } = letrec(
+      (tie) => ({
+        node: record({
+          value: new FakeIntegerArbitrary(),
+          next: tie('node'),
+        }),
+      }),
+      { circular: true },
+    );
+    return node;
+  };
+
+  it('should produce the same values given the same seed', () => {
+    assertProduceSameValueGivenSameSeed(letrecBuilder);
   });
 });
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -1,4 +1,12 @@
 packages:
-  - 'examples'
-  - 'packages/*'
-  - 'website'
+  - examples
+  - packages/*
+  - website
+
+onlyBuiltDependencies:
+  - '@swc/core'
+  - core-js
+  - core-js-pure
+  - es5-ext
+  - esbuild
+  - unrs-resolver


packages/fast-check/src/arbitrary/letrec.ts
```diff
@@ -50,6 +55,131 @@ export type LetrecLooselyTypedTie = (key: string) => Arbitrary<unknown>;
...
+      continue;
+    }
+    const lazyAtKey: LazyArbitrary<unknown> | undefined = lazyArbs[key];
+    const lazyArb = lazyAtKey !== undefined ? lazyAtKey : new LazyArbitrary(key);
+    lazyArb.underlying = nat().map((index) => ({
+      [placeholderSymbol]: { key, index },
+    }));
+    lazyArbs[key] = lazyArb;
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    poolArbs[key] = array(strictArbs[key]!, { minLength: 1 });
```
dubzzz: We are probably generating way more items in the pool than really required.

dubzzz: We probably don't need to preserve shrinking capabilities on the index thing.

packages/fast-check/src/arbitrary/letrec.ts
```diff
@@ -50,6 +55,131 @@ export type LetrecLooselyTypedTie = (key: string) => Arbitrary<unknown>;
...
+  }
+
+  for (const key in strictArbs) {
+    if (!safeHasOwnProperty(strictArbs, key)) {
+      // Prevents accidental iteration over properties inherited from an object’s prototype
+      continue;
+    }
+
+    const poolsArb = record(poolArbs as any) as Arbitrary<{ [K in keyof T]: unknown[] }>;
+    strictArbs[key] = poolsArb.map((pools) => {
```
dubzzz, line 199: Not providing an unmapper function will prevent capabilities to shrink values without any context on them.

packages/fast-check/src/arbitrary/letrec.ts
TomerAberbach: > We are probably generating way more items in the pool than really required.

Well, it kind of depends on how cyclical you want the resulting structure to be right? A smaller pool increases the change of cycles happening while a bigger pool makes it less likely?

Maybe the options should look roughly like this:
```ts
{
  /** @default false */
  // Btw, I like `withCycles` more than `cyclic` since it better matches other constraints
  // like `withBoxedValues`
  withCycles?: boolean | {
      // We would have to invert this input size since "higher frequency" needs to
      // be "smaller array", not "bigger array"
      frequency: SizeForArbitrary,
  }
}
```

And then we can also choose a reasonable default (for when passing a `boolean`)

> We probably don't need to preserve shrinking capabilities on the index thing.

Do you mean adding `noShrink` around `nat().map(...)`? Is that correct? Can you also help me understand when you should/should not preserve shrinking?

TomerAberbach:
@dubzzz would you also be able to take a look at my "open questions" in my PR description, and give your thoughts? https://github.com/dubzzz/fast-check/pull/6040#issue-3207416741

packages/fast-check/src/arbitrary/letrec.ts
TomerAberbach: I'm looking into implementing this and came across an interesting question.

Does `mapper` have to be bijective? Such that `unmapper(mapper(something)) === something`? Or is it okay for `unmapper(mapper(something))` to not always return `something`?

I think currently it cannot be bijective for this `poolsArb` thing. The problem is that each pool is not necessarily a unique array. So it's possible for one of the pools to have the same value twice, which means `unmapper` has two choices for what to use for the `index` in `placeholderSymbol`. Is this actually a problem, or is it fine?

TomerAberbach: I looked around and clearly it doesn't need to be bijective!

packages/fast-check/src/arbitrary/letrec.ts
TomerAberbach: Went ahead with this

dubzzz:
I'm not forgetting this PR. I just have a very very limited bandwidth at the moment. I promise I'll try to review and if needed provide you with feedback as soon as I find the time for it 🤗

dubzzz:
In the meantime, you can install it via:

```sh
npm i https://pkg.pr.new/fast-check@6040
```

packages/fast-check/test/unit/arbitrary/__test-helpers__/ArbitraryAssertions.ts
```diff
@@ -107,7 +107,7 @@ export function assertProduceCorrectValues<T, U = never>(
         },
       )
       .afterEach(poisoningAfterEach),
-    assertParameters,
+    { ...assertParameters, ...{ seed: -1659794355, path: '0', endOnFailure: true } },
```
dubzzz:
```suggestion
    assertParameters,
```

packages/fast-check/test/unit/arbitrary/letrec.spec.ts
```diff
@@ -347,3 +445,46 @@ describe('letrec (integration)', () => {
     assertShrinkProducesSameValueWithoutInitialContext(letrecBuilder);
   });
 });
+
+describe('letrec with cycles (integration)', () => {
```
dubzzz, line 449: Ideally it would be great to also support:

- should produce values seen as shrinkable without any context (if underlyings do)
- should be able to shrink to the same values without initial context (if underlyings do)

packages/fast-check/test/unit/arbitrary/letrec.spec.ts
```diff
@@ -30,6 +32,31 @@ describe('letrec', () => {
       expect(arb2).toBe(expectedArb2);
     });
 
+    it('should be able to construct independent arbitraries with cycles', () => {
```
dubzzz, line 35: Thanks for these new tests, they look great!

packages/fast-check/src/arbitrary/letrec.ts
```diff
@@ -50,6 +58,213 @@ export type LetrecLooselyTypedTie = (key: string) => Arbitrary<unknown>;
...
+  const lazyArbs: { [K in keyof T]?: LazyArbitrary<unknown> } = safeObjectCreate(null);
+  const tie = (key: keyof T): Arbitrary<any> => {
+    if (!safeHasOwnProperty(lazyArbs, key)) {
+      // Call to hasOwnProperty ensures that the property key will be defined
+      lazyArbs[key] = new LazyArbitrary(String(key));
+    }
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    return lazyArbs[key]!;
+  };
+  const strictArbs = builder(tie as any);
```
dubzzz, line 169: Personal note: Fully identical to the non-cycle one

packages/fast-check/src/arbitrary/letrec.ts
```diff
@@ -50,6 +58,213 @@ export type LetrecLooselyTypedTie = (key: string) => Arbitrary<unknown>;
...
+      // Prevents accidental iteration over properties inherited from an object’s prototype
+      continue;
+    }
+    const lazyAtKey: LazyArbitrary<unknown> | undefined = lazyArbs[key];
+    const lazyArb = lazyAtKey !== undefined ? lazyAtKey : new LazyArbitrary(key);
+    lazyArb.underlying = noShrink(nat().map((index) => ({ [placeholderSymbol]: { key, index } })));
+    lazyArbs[key] = lazyArb;
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    poolArbs[key] = array(strictArbs[key]!, poolConstraints);
+  }
```
dubzzz, line 190: Personal note: Close to the non-cycle one. The difference is that non-cycle one was doing `lazyArb.underlying = strictArbs[key]`. Plus the pool.

packages/fast-check/src/arbitrary/letrec.ts
```diff
@@ -50,6 +58,213 @@ export type LetrecLooselyTypedTie = (key: string) => Arbitrary<unknown>;
...
+    }
+    safeAdd(visited, value);
+
+    if (safeHasOwnProperty(value, placeholderSymbol)) {
+      // This is a while loop because it's possible for an arbitrary to be defined as just `arb: tie('otherArb')`, in
+      // which case what the `arb` generates is also a placeholder.
+      do {
+        const { key, index } = (value as { [placeholderSymbol]: { key: keyof T; index: number } })[placeholderSymbol];
+        const pool = pools[key];
+        value = pool[index % pool.length];
```
dubzzz: Could you create another variable? I prefer not changing the value of a parameter. So it would be better to have a `let myRandomNameValue = value` right outside of the `do..while`.

packages/fast-check/src/arbitrary/letrec.ts
```diff
@@ -50,6 +58,213 @@ export type LetrecLooselyTypedTie = (key: string) => Arbitrary<unknown>;
...
+    if (!safeHasOwnProperty(strictArbs, key)) {
+      // Prevents accidental iteration over properties inherited from an object’s prototype
+      continue;
+    }
+
+    const poolsArb = record(poolArbs as any) as Arbitrary<{ [K in keyof T]: unknown[] }>;
+    strictArbs[key] = poolsArb.map(
+      (pools) => {
+        derefPools(pools, placeholderSymbol);
+        return pools[key][0];
```
dubzzz: From my understanding of the deref trick, we could probably avoid deref on indices >0, except if they become useful. We clearly need to deref all keys (actually we could potentially only start to deref only the keys that really get requested by playing on a getter trick to only trigger the deref on "requested key" and "0" then unlocking the other ones needed by the deref process).

See that as an optimization but can be done later. I'd even say to postpone it to another review.

packages/fast-check/src/arbitrary/letrec.ts
```diff
@@ -50,6 +58,213 @@ export type LetrecLooselyTypedTie = (key: string) => Arbitrary<unknown>;
...
+    size: invertSize(resolveSize(constraints.frequency)),
+  };
+  for (const key in strictArbs) {
+    if (!safeHasOwnProperty(strictArbs, key)) {
+      // Prevents accidental iteration over properties inherited from an object’s prototype
+      continue;
+    }
+    const lazyAtKey: LazyArbitrary<unknown> | undefined = lazyArbs[key];
+    const lazyArb = lazyAtKey !== undefined ? lazyAtKey : new LazyArbitrary(key);
+    lazyArb.underlying = noShrink(nat().map((index) => ({ [placeholderSymbol]: { key, index } })));
```
dubzzz, line 186: The thing that blocks the code from being able to shrink without context (see the tests I asked for) is probably this one. It does not provide any unmapping for the `nat.map`.

Maybe we can see that later. Not critical for that PR and it should definitely be doable which is the only thing I was worried of.

packages/fast-check/src/arbitrary/letrec.ts
```diff
@@ -50,6 +58,213 @@ export type LetrecLooselyTypedTie = (key: string) => Arbitrary<unknown>;
...
+    lazyArbs[key] = lazyArb;
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    poolArbs[key] = array(strictArbs[key]!, poolConstraints);
```
dubzzz, line 189: Works well, in later versions of the feature I think we should try to find ways not to generate a whole array of values to finally only use some of them or none of them.

packages/fast-check/src/arbitrary/letrec.ts
```diff
@@ -50,6 +58,213 @@ export type LetrecLooselyTypedTie = (key: string) => Arbitrary<unknown>;
...
+ * Constraints to be applied on {@link LetrecConstraints.withCycles}
+ * @remarks Since 4.2.0
+ * @public
+ */
+export interface CycleConstraints {
+  /**
+   * Define how frequently cycles should occur in the generated values (at max)
+   * @remarks Since 4.2.0
+   */
+  frequency?: Exclude<SizeForArbitrary, 'max'>;
```
dubzzz: I'd probably stick to a naming with size in it. It would make it more explicit for the end users that the option is controlled by the size

packages/fast-check/src/arbitrary/letrec.ts
```diff
@@ -50,6 +58,213 @@ export type LetrecLooselyTypedTie = (key: string) => Arbitrary<unknown>;
...
+    const lazyAtKey: LazyArbitrary<unknown> | undefined = lazyArbs[key];
+    const lazyArb = lazyAtKey !== undefined ? lazyAtKey : new LazyArbitrary(key);
+    lazyArb.underlying = strictArbs[key];
+    lazyArbs[key] = lazyArb;
+  }
+  return strictArbs;
+}
+
+/** @internal */
+function derefPools<T>(pools: { [K in keyof T]: unknown[] }, placeholderSymbol: symbol) {
```
dubzzz: Worth being extracted into a dedicated file and tested. No need to write hundreds of tests but just one with several values and keys would already be great

packages/fast-check/src/arbitrary/letrec.ts
```diff
@@ -50,6 +58,213 @@ export type LetrecLooselyTypedTie = (key: string) => Arbitrary<unknown>;
...
+    const poolsArb = record(poolArbs as any) as Arbitrary<{ [K in keyof T]: unknown[] }>;
+    strictArbs[key] = poolsArb.map(
+      (pools) => {
+        derefPools(pools, placeholderSymbol);
+        return pools[key][0];
+      },
+      (value) =>
+        refPools(value, key, placeholderSymbol) as {
+          [K in keyof T]: unknown[];
+        },
```
dubzzz: Not sure of this part of the code, maybe for the first iteration we should drop it and start with something that cannot be reversed (anyway at this point it cannot given the nat thing does not provide an unmap function to the map).

Normally value is supposed to be something with the same type as pools[key][0] in the function above. I'm not sure to understand the way it works when reading refPools.

dubzzz (commented):
Awesome work! Thanks a lot for the time you spent on this PR. I'd have a few requests before merging it:

1. Drop any unmap code for now

_Currently you provided an unmap code via `poolsArb.map` but none for `nat().map`, so it won't be enough. Let do it in a second time, it would probably help focusing the PR on the core feature._

2. Move `derefPools` into a dedicated file and put a small test on it

3. Factorizing `letrecWithCycles` and `letrecWithoutCycles` would be nice

_I have the feeling that "without" one could stop at `return strictArbs;` while the "with" would have to add its pools._

packages/fast-check/test/unit/arbitrary/letrec.spec.ts
dubzzz: **POSTPONED** Let do it later when we will address the ability to unmap values or shrink them without more context on them.

packages/fast-check/src/arbitrary/letrec.ts
dubzzz: I fell that up to this loop we could have the two codes being the same. The "with cycles" one could just be a follow up code not returning immediately but doing the extra stuff with pools and cie.

packages/fast-check/src/arbitrary/letrec.ts
dubzzz: **POSTPONED** Let do it later, just an idea for a future PR. But let first focus on the key need.

packages/fast-check/src/arbitrary/letrec.ts
dubzzz: **POSTPONED** Let do it later, just an idea for a future PR. But let first focus on the key need.

packages/fast-check/src/arbitrary/letrec.ts
dubzzz: **POSTPONED** Let do it later, just an idea for a future PR. But let first focus on the key need.

packages/fast-check/src/arbitrary/letrec.ts
dubzzz: We used depthSize for fc anything

packages/fast-check/test/unit/arbitrary/__test-helpers__/ArbitraryAssertions.ts
TomerAberbach: Oops, thanks

packages/fast-check/src/arbitrary/letrec.ts
TomerAberbach: So `frequencySize`?

packages/fast-check/src/arbitrary/letrec.ts
TomerAberbach: Added tests, but left in the same file (since seems like it'll only be used here and we don't need to export outside of `fast-check`)

TomerAberbach:
@dubzzz I think I've addressed things now (sorry, was busy for a while)

The one thing we might want to think about is catching the scenario where we we accidentally don't replace some `placeholderSymbol`. I tried tracking them in a local set from `nat().map(...)` and then deleting them later when we replace, and asserting we have zero after, but this was causing problems, so I'm probably missing something. Do you have any ideas around how this should work?

TomerAberbach:
@gruhn since you seem to be around, curious if you have any thoughts on my [comment above](https://github.com/dubzzz/fast-check/pull/6040#issuecomment-3238775228)? I got a bit stuck on this point

packages/fast-check/src/arbitrary/letrec.ts
```diff
@@ -50,6 +57,154 @@ export type LetrecLooselyTypedTie = (key: string) => Arbitrary<unknown>;
  */
 export type LetrecLooselyTypedBuilder<T> = (tie: LetrecLooselyTypedTie) => LetrecValue<T>;
 
+/**
+ * Constraints to be applied on {@link letrec}
+ * @remarks Since 4.2.0
```
dubzzz:
```suggestion
 * @remarks Since 4.4.0
```

packages/fast-check/src/arbitrary/letrec.ts
```diff
@@ -50,6 +57,154 @@ export type LetrecLooselyTypedTie = (key: string) => Arbitrary<unknown>;
...
+ * @public
+ */
+export interface LetrecConstraints {
+  /**
+   * Generate objects with circular references
+   * @defaultValue false
+   * @remarks Since 4.2.0
```
dubzzz:
```suggestion
   * @remarks Since 4.4.0
```

packages/fast-check/src/arbitrary/letrec.ts
```diff
@@ -50,6 +57,154 @@ export type LetrecLooselyTypedTie = (key: string) => Arbitrary<unknown>;
...
+   */
+  withCycles?: boolean | CycleConstraints;
+}
+
+/**
+ * Constraints to be applied on {@link LetrecConstraints.withCycles}
+ * @remarks Since 4.2.0
```
dubzzz:
```suggestion
 * @remarks Since 4.4.0
```

packages/fast-check/src/arbitrary/letrec.ts
```diff
@@ -50,6 +57,154 @@ export type LetrecLooselyTypedTie = (key: string) => Arbitrary<unknown>;
...
+ * @public
+ */
+export interface CycleConstraints {
+  /**
+   * Define how frequently cycles should occur in the generated values (at max)
+   * @remarks Since 4.2.0
```
dubzzz:
```suggestion
   * @remarks Since 4.4.0
```

dubzzz (dismissed):
Perfect, let's start with this version. I'm gonna add in your PR the few changes related to version not being 4.2.0 but 4.4.0.

packages/fast-check/src/arbitrary/letrec.ts
```diff
@@ -50,6 +57,154 @@ export type LetrecLooselyTypedTie = (key: string) => Arbitrary<unknown>;
  */
 export type LetrecLooselyTypedBuilder<T> = (tie: LetrecLooselyTypedTie) => LetrecValue<T>;
 
+/**
+ * Constraints to be applied on {@link letrec}
+ * @remarks Since 4.4.0
+ * @public
+ */
+export interface LetrecConstraints {
```
dubzzz, line 65: Let's do that in a follow up PR so that we unlock this one:

We should import and export it from fast-check-default.ts so that end users can import it too.

It should be a type rather than an interface to fest with what we currently do for all others (to be confirmed, I have rechecked).

dubzzz (approved):

dubzzz:
A huge and massive thank you for this awesome and very valuable contribution ❤️ 

Sorry it took so long to validate it on my end.

TomerAberbach:
Thanks! Another thing that would be good to do is figure out if we can detect this case I mentioned:

https://github.com/dubzzz/fast-check/pull/6040#issuecomment-3238775228

Do you have thoughts?

dubzzz:
Cc @TomerAberbach - Just to give you a status

I delayed a little bit the next minor as I was not sure on some stuff around the new cyclic version of letrec. From my last tests it comes with some bugs whenever there is a need to map on a tie...

Here is a simple example:

```ts
import { expect, test } from 'vitest';
import fc from 'fast-check';

test('playing', () => {
  const [mappedValue] = fc.sample(
    fc.letrec(
      (tie) => ({
        recThing: fc.record({
          keys: tie('recThing').map((recThing: any) => Object.keys(recThing)),
        }),
      }),
      { withCycles: true }
    ).recThing,
    { numRuns: 1 }
  );
  expect(mappedValue).toEqual({ keys: ['keys'] });
});
```

_Accessible at: https://stackblitz.com/edit/vitest-dev-vitest-gv4kxibq?file=test%2Fbasic.test.ts_

I do have some ideas to make it work and to drop some unneeded computations (we compute big arrays while we may never use anything from it).

dubzzz:
Or I may release the flag with `experimentalWithCycles` instead.

dubzzz:
This one is also broken:

```ts
test('playing set', () => {
  const [setValue] = fc.sample(
    fc.letrec(
      (tie) => ({
        recThing: fc.set(tie('recThing'), { minLength: 1, maxLength: 1 }),
      }),
      { withCycles: true }
    ).recThing,
    { numRuns: 1 }
  );
  expect(setValue).toEqual(new Set([expect.any(Set)]));
});
```

Same my idea would probably fix it. But I need to spend time on it.

TomerAberbach:
Hmmm, surprised the `.map(...)` doesn't work, but the `fc.set(...)` one makes sense, cause we need to handle sets, maps, etc. in the logic

Not sure if you missed it, but I had a [question above](https://github.com/dubzzz/fast-check/pull/6040#issuecomment-3238775228) that, if we can answer, would help here. i.e. we should detect when we fail to replace a placeholder and maybe throw an error telling folks to file an issue

dubzzz:
Actually anything that needs to consume the value instantly breaks. Set and map were just examples.

Filter, chain, fc.func, fc.infiniteStream... also fail.

dubzzz:
I believe the whole issue is that we can hardly control what the user will do with the output of tie. I have the feeling that maybe letrec should not be cyclic at all but that we should come up with a new primitive a bit more constrained and able to cycle. I'm currently POC-ing one taking most of your idea but on a slightly changed API.

TomerAberbach:
I thought about this some more, and I think the inherent issue is:
1. Creating circular references fundamentally requires mutation
2. Creating circular references via mutation fundamentally requires knowing the "parent object" to mutate
3. Locating the parent objects of placeholder values requires traversing the generated object from the root
4. Traversing arbitrary objects is hard because you can have maps, sets, etc. or even things like functions, and you can't even traverse the "return values" of functions (since it's not like you wanna call them), and similarly for classes, which may have inaccessible private fields after construction

Thus, it would seem that the safest API should solve 2 in the list above, by requiring the end-user to essentially explain what the parent object is that needs to be mutated (and which key or entry or whatever)

This is unfortunate because it's hard to come up with a generalized API for that. e.g. For plain objects, you want the user to specify "this property can be circular", for arrays "this index", for maps "this key", for sets and functions how do you even specify?

Maybe you'e already thought of this and have a good API in mind, but I'll do some more thinking on my end about it

dubzzz:
I have an API in mind that should probably make us able to control more precisely the objects being used.

Still on playing with a poc and extending it to cover as many cases as possible to make it usable.

PR #6119, opened 2025-08-26
🏷️ Add union type overloads for nat() and bigInt()

### LINKED ISSUE #6073: Add union type overloads where applicable
## 🚀 Feature Request

Where the API provides overloads with different parameter types, provide variants that accept unions of those parameter types as well.

## Motivation

Let's say we are writing a `head` operation on Node.js readable streams. We want to generate a scenario with an arbitrary stream and number of items to include:

```typescript
interface HeadConstraints {
  upstream: TestStreamConstraints,  // defined elsewhere
  n: number | NatConstraints,
}
```

We want to render this into parameters to pass during our test:
```typescript
function headScenario(constraints: HeadConstraints): Arbitrary<[stream.Readable, number]> {
  return fc.tuple(
    testStream(constraints.upstream),
    nat(constraints.n),  // Error!
  );
}
```

fast-check gives the following error:
```typescript
    src/lib/stream.spec.ts:80:18 - error TS2769: No overload matches this call.
      Overload 1 of 3, '(max: number): Arbitrary<number>', gave the following error.
        Argument of type 'number | NatConstraints' is not assignable to parameter of type 'number'.
          Type 'NatConstraints' is not assignable to type 'number'.
      Overload 2 of 3, '(constraints: NatConstraints): Arbitrary<number>', gave the following error.
        Argument of type 'number | NatConstraints' is not assignable to parameter of type 'NatConstraints'.
          Type 'number' has no properties in common with type 'NatConstraints'.

    80     nat = fc.nat(constraints.n);
```

- `nat` can accept a `number`, but `number | NatConstraints` is not assignable to `number`.
- `nat` can accept a `NatConstraints`, but `number | NatConstraints` is not assignable to `NatConstraints`.

At that point, the compiler gives up.

Of course, `nat` works perfectly well with a argument of type `number | NatConstraints`, but TypeScript cannot infer this automatically (discussion [here](https://github.com/microsoft/TypeScript/issues/14107)). The current solution is to narrow the argument to one type or the other:

```typescript
let nat: Arbitrary<number>;
if (typeof constraints.n === 'number') {
  nat = fc.nat(constraints.n);
} else {
  nat = fc.nat(constraints.n);
}
```

A bit awkward.

The most convenent solution is to provide an overload for `fc.nat`:
```typescript
declare function nat(): Arbitrary<number>;
declare function nat(max: number): Arbitrary<number>;
declare function nat(constraints: NatConstraints): Arbitrary<number>;
declare function nat(constraints: Number | NatConstraints): Arbitrary<number>;  // fixed!
```

`bigInt` would be another candidate for this treatment, and possibly others.

dmurvihill:
**Description**

Adds the following function definitions:
```typescript
declare function nat(arg?: number | NatConstraints): Arbitrary<number>;
declare function bigInt(...args: [] | [bigint, bigint] | [BigIntConstraints]): Arbitrary<bigint>;
```

Previously, passing a `number | NatConstraints` to `nat` would fail:
```
    nat can accept a number, but number | NatConstraints is not assignable to number.
    nat can accept a NatConstraints, but number | NatConstraints is not assignable to NatConstraints.
```

And similar behavior for `bigInt`.

These are the only arbitraries that require appropriate union type overloads.

Fixes #6073.

**Advanced**

- [x] Category: 🏷️ Types change.
- [x] Impacts: Reflection on calls to `bigInt` or `nat` may be impacted (see the changes to `double.spec.ts` for an example). This shouldn't be guaranteed stable IMO, since it's very unlikely that production software would need rely on such a thing.

### DIFF
--- a/packages/fast-check/src/arbitrary/bigInt.ts
+++ b/packages/fast-check/src/arbitrary/bigInt.ts
@@ -81,6 +81,15 @@ function bigInt(min: bigint, max: bigint): Arbitrary<bigint>;
  * @public
  */
 function bigInt(constraints: BigIntConstraints): Arbitrary<bigint>;
+/**
+ * For bigint between min (included) and max (included)
+ *
+ * @param args - Either min/max bounds as an object or constraints to apply when building instances
+ *
+ * @remarks Since 2.6.0
+ * @public
+ */
+function bigInt(...args: [] | [bigint, bigint] | [BigIntConstraints]): Arbitrary<bigint>;
 function bigInt(...args: [] | [bigint, bigint] | [BigIntConstraints]): Arbitrary<bigint> {
   const constraints = buildCompleteBigIntConstraints(extractBigIntConstraints(args));
   if (constraints.min > constraints.max) {
--- a/.changeset/light-eggs-cut.md
+++ b/.changeset/light-eggs-cut.md
@@ -1,0 +1,5 @@
+---
+'fast-check': patch
+---
+
+Add union type overloads for `nat` and `bigInt`
--- a/packages/fast-check/src/arbitrary/nat.ts
+++ b/packages/fast-check/src/arbitrary/nat.ts
@@ -42,6 +42,15 @@ function nat(max: number): Arbitrary<number>;
  * @public
  */
 function nat(constraints: NatConstraints): Arbitrary<number>;
+/**
+ * For positive integers between 0 (included) and max (included)
+ *
+ * @param arg - Either a maximum number or constraints to apply when building instances
+ *
+ * @remarks Since 2.6.0
+ * @public
+ */
+function nat(arg?: number | NatConstraints): Arbitrary<number>;
 function nat(arg?: number | NatConstraints): Arbitrary<number> {
   const max = typeof arg === 'number' ? arg : arg && arg.max !== undefined ? arg.max : 0x7fffffff;
   if (max < 0) {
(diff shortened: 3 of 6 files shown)


packages/fast-check/src/arbitrary/bigInt.ts
```diff
@@ -81,6 +81,15 @@ function bigInt(min: bigint, max: bigint): Arbitrary<bigint>;
...
 function bigInt(constraints: BigIntConstraints): Arbitrary<bigint>;
+/**
+ * For bigint between min (included) and max (included)
+ *
+ * @param args - Either min/max bounds as an object or constraints to apply when building instances
+ *
+ * @remarks Since 2.6.0
+ * @public
+ */
+function bigInt(...args: [] | [bigint, bigint] | [BigIntConstraints]): Arbitrary<bigint>;
```
dubzzz, line 92: Should we replace the two overloads instead of being one extra one? I believe we could drop all overloads with this approach. That said not sure if we can have the same level of jsdoc with my suggestion.

Your suggestion has the benefit to be a first move. Let's do it that way for the first iteration.

dubzzz (approved):
Don't we have the same issue with integer? If so, maybe worth doing the same for them

dubzzz:
Just updated your branch given I should have fixed the CI

packages/fast-check/src/arbitrary/bigInt.ts
dmurvihill: That would make it difficult to document the parameter clearly, I think.

dmurvihill:
> Don't we have the same issue with integer? If so, maybe worth doing the same for them

The `(min, max)` overload was removed from `integer` in 3.0.0; so `function integer(constraints?: IntegerConstraints)` captures all valid inputs.

Issue #6216, opened 2025-10-22
Performance: replace multiply by Math.imul in perf-intensive contexts


dubzzz:

dubzzz:
Just released the fixed version on npm 🚀

PR #6278, opened 2025-11-06
⚡ Use Math.imul and shifts in perf-critical paths

emilianbold:
Replace integer multiplication with Math.imul and bit shift for better performance:

- FloatHelpers.ts: Use Math.imul for exponent multiplication (~7.5% faster)
- ReplayPath.ts: Use bit shift (<<) instead of multiply by 2

Both changes use integer-only operations in hot paths during property testing.

Related to #6216

Note: I had some extra credits so I unleashed Claude on looking for potential places where `imul` would make sense and this is the only one that made sense in the end. I did review it (too much, really). Since exponent should be 8 bits for a 32bit float, it does fit the 32bit imul int. It's really a tiny contribution but does seems to help performance.

**Description**

- [ ] My PR adds relevant tests and they would have failed without my PR (when applicable)

**Advanced**

- [ ] Category: ...
- [ ] Impacts: ...

### DIFF
--- a/packages/fast-check/src/arbitrary/_internals/helpers/FloatHelpers.ts
+++ b/packages/fast-check/src/arbitrary/_internals/helpers/FloatHelpers.ts
@@ -1,5 +1,6 @@
 const safeNegativeInfinity = Number.NEGATIVE_INFINITY;
 const safePositiveInfinity = Number.POSITIVE_INFINITY;
+const safeMathImul = Math.imul;
 
 /** @internal */
 export const MIN_VALUE_32: number = 2 ** -126 * 2 ** -23;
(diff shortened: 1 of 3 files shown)


packages/fast-check/src/arbitrary/_internals/helpers/FloatHelpers.ts
```diff
@@ -67,7 +67,7 @@ function indexInFloatFromDecomp(exponent: number, significand: number) {
   }
   // Offset due to exp = -126 + Offset of previous exp (excl. -126) + Offset in current exp
   // 2**24 + (exponent - (-126) -1) * 2**23 + (significand - 1) * 2**23
-  return (exponent + 127) * 0x800000 + (significand - 1) * 0x800000;
+  return Math.imul(exponent + 127, 0x800000) + (significand - 1) * 0x800000;
```
dubzzz: Could you replace imul by something as I did here: https://github.com/dubzzz/fast-check/blob/main/packages%2Ffast-check%2Fsrc%2Farbitrary%2Fnat.ts#L4.

dubzzz (commented):
Perfect. Just a minor request

packages/fast-check/src/arbitrary/_internals/helpers/FloatHelpers.ts
emilianbold: Is the new `const safeMathImul = Math.imul;` what you would want? It didn't occur to me you are being defensive against users replacing standard functions and constant. Interesting.

dubzzz (approved):

Issue #5562, opened 2024-12-26
Support `minLength` and `maxLength` for `stringMatching`


TomerAberbach:
## 🚀 Feature Request

`stringMatching` generates based on the regular expression, but doesn't provide a way to set a min or max length on the generated strings.

## Motivation

Sometimes you want to test specific lengths of a string matching a regex, but modifying a regular expression to have a specific min and/or max length can be pretty hard and confusing. It would be a lot easier to just be able to pass `minLength` and `maxLength` like the `string` arbitrary.

## Example

```
const nameRegex = /^[a-z]+$/

// Test different lengths without having to modify regex
fc.stringMatching(nameRegex, { minLength: 10 })
fc.stringMatching(nameRegex, { maxLength: 10 })
fc.stringMatching(nameRegex, { minLength: 5, maxLength: 10 })
```

dubzzz:
So far the only option I have for it is to play with `.filter` but unfortunately performance-wise it will not be ideal. I'd need to think about a way to make such constraint more optimal before offering it on the arbitrary.

gruhn:
Hmm, interesting problem. Regular expressions are closed under intersection. So one idea might be to "just" compute the intersection of `someRegex` and for example `/^.{5,10}$/`. That gives a new regular expression which is restricted to strings with min length 5 and max length 10. I think this can be done in ~`O(n^2)`. Not super cheap but it's an upfront cost and maybe negligible for small user defined regular expressions. 

One way to do it might be:

1. Convert both regular expressions to [NFAs](https://en.wikipedia.org/wiki/Nondeterministic_finite_automaton#NFA_with_%CE%B5-moves) (I think: `O(n)`)
2. [Compute NFA intersection](https://stackoverflow.com/questions/21662041/how-to-find-the-intersection-of-two-nfa) (I think: `O(n^2)`)
3. [Convert resulting NFA back to regular expression](https://courses.grainger.illinois.edu/cs374/fa2017/extra_notes/01_nfa_to_reg.pdf) (I think: `O(n)`)

Any route involving DFAs seems to come with the danger of exponential blow up. 

For the regular expressions from formal language theory, this is maybe not too complicated but I can imagine you need a ton of case distinctions for the the full JavaScript regular expressions. So not sure if this is the best approach.

TomerAberbach:
This certainly sounds quite tricky! Thanks for all the research here :)

[...]

PR #6285, opened 2025-11-07
✨ Add maxLength support to stringMatching

emilianbold:
**Description**

This is a weaker implementation as mentioned in #5562 but one which I think is superior to just adding a filter by hand.

Propagating max downstream means we do contain a bit better performance wise though we still need to filter the final result.

Also explored min but that seemed less valuable than containing max better.

  - [ ] My PR does not add any new dependency without an associated Issue or Discussion

**Advanced**

- [ ] Category: ...
- [ ] Impacts: ...

### DIFF
--- a/packages/fast-check/test/unit/arbitrary/_internals/helpers/SanitizeRegexAst.spec.ts
+++ b/packages/fast-check/test/unit/arbitrary/_internals/helpers/SanitizeRegexAst.spec.ts
@@ -25,4 +25,20 @@ describe('addMissingDotStar', () => {
     const twiceTransformedSourceAst = addMissingDotStar(transformedSourceAst);
     expect(twiceTransformedSourceAst).toEqual(transformedSourceAst);
   });
+
+  describe('with maxLength constraint', () => {
+    it.each`
+      source           | maxLength | target
+      ${/a/}           | ${10}     | ${/(?:^.{0,10}a.{0,10}$)/}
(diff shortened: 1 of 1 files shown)


emilianbold:
Note the patch has 100% coverage. Not sure why codecov is complaining.

packages/fast-check/src/arbitrary/stringMatching.ts
```diff
@@ -107,10 +113,23 @@ function toMatchingArbitrary(
           return string({ ...constraints, minLength: 0, maxLength: 1, unit: node });
         }
         case 'Range': {
+          // Cap quantifiers to user's maxLength constraint to prevent unbounded generation
+          const userMaxLength = constraints.maxLength;
```
dubzzz, line 117: I'd rather deal with everything linked to maxLength at the same place. So I'd either move these checks when we parse the regex (as you did for wildcard) or move them when reading from the parsed regex. I'd honestly rather advocate for the second option so that parsing stays totally agnostic of this notion of length.

dubzzz:
I'm taking back your initiative of supporting maxLength on stringMatching. Your current implementation proved to be a really good starting point for such feature. I'm gonna open a new PR with your commits kept to keep you as a contributor of the change.

dubzzz:
@emilianbold Normally you'll be marked as a co-author of the resulting commit. Thanks again for this contribution, it unlocked a long running request

PR #6318, opened 2025-11-24
♻️ Extract logic building lazy arbs

dubzzz:
**Description**

In the context of `letrec`, the `tie` function passed to the caller is responsible to build lazy instances of arbitraries. When first built they cannot generate anything given they have to be given more details on the way to generate the asked instances. They just now the name of the arbitrary they will be responsible to generate but not the how to build it.

In this PR we do several things:

- Re-use the `tie` function exposed to the user internally, so feel the lazy instances once the user returned us its instances
- Move from an Object to a Map for better performance

**Advanced**

- [x] Category: ...
- [x] Impacts: ...

### DIFF
--- a/.changeset/polite-crews-listen.md
+++ b/.changeset/polite-crews-listen.md
@@ -1,0 +1,5 @@
+---
+"fast-check": patch
+---
+
+♻️ Extract logic building lazy arbs


packages/fast-check/test/unit/arbitrary/letrec.spec.ts
```diff
@@ -216,6 +216,29 @@ describe('letrec', () => {
...
+    { key: 2147483648, stringKey: '2147483648' },
+    { key: -1, stringKey: '-1' },
+  ])('should be able to construct arbitraries referenced by the numeric key $key', ({ key, stringKey }) => {
+    // Arrange
+    const expectedArb = new FakeIntegerArbitrary(1, 4);
+
+    // Act
+    const { referenceToKey } = letrec((tie) => ({
+      referenceToKey: tie(stringKey),
+      [key]: expectedArb,
```
dubzzz:
```suggestion
      // built with a numeric key...
      [key]: expectedArb,
      // ...referenced by a string key (can't be referenced by the numeric key typing-wise)
      referenceToKey: tie(stringKey),
```

PR #6530, opened 2026-02-02
📝 Clearer guidelines for constraints of arbs in skill

dubzzz:


skills/javascript-testing-expert/SKILL.md
```diff
@@ -198,8 +198,11 @@ The risk being that you may end up rewriting the code being tested in the test
 
 **✅ Do** expect some aspects and characteristics of the returned value
 
-**❌ Don't** over constrain arbitraries when using `fast-check`  
-Eg.: if the algorithm has no restriction for the length of a string let fast-check generate any string, if the length causes performance issues to your algorithm or test consider using the `size` constraint available on string (and others) to ask for smaller strings  
+**❌ Don't** specify any `maxLength` on an arbitrary if it is a not a requirement of the algorithm  
+**👍 Prefer** specifying a `size: '-1'` if you feel that the algorithm will take very long on large inputs (by default fast-check generates up to 10 items, so only use `size` when clearly required)
```
dubzzz:
```suggestion
**👍 Prefer** specifying a `size: '-1'` if you feel that the algorithm will take very long on large inputs (by default fast-check generates up to 10 items, so only use `size` when clearly required)  
Eg.: No `fc.string({maxLength: 5})` or `fc.array(arb, {maxLength: 8})` except being a string requirement
```

skills/javascript-testing-expert/SKILL.md
```diff
@@ -198,8 +198,12 @@ The risk being that you may end up rewriting the code being tested in the test
 
 **✅ Do** expect some aspects and characteristics of the returned value
 
-**❌ Don't** over constrain arbitraries when using `fast-check`  
-Eg.: if the algorithm has no restriction for the length of a string let fast-check generate any string, if the length causes performance issues to your algorithm or test consider using the `size` constraint available on string (and others) to ask for smaller strings  
+**❌ Don't** specify any `maxLength` on an arbitrary if it is a not a requirement of the algorithm
```
dubzzz:
```suggestion
**❌ NEVER** specify any `maxLength` on an arbitrary if it is a not a requirement of the algorithm  
**❌ NEVER** specify any constraint on an arbitrary if it is not a requirement of the arbitrary, use defaults as much as possible  
```

skills/javascript-testing-expert/SKILL.md
```diff
@@ -198,8 +198,13 @@ The risk being that you may end up rewriting the code being tested in the test
...
-**❌ Don't** over constrain arbitraries when using `fast-check`  
-Eg.: if the algorithm has no restriction for the length of a string let fast-check generate any string, if the length causes performance issues to your algorithm or test consider using the `size` constraint available on string (and others) to ask for smaller strings  
+**❌ NEVER** specify any `maxLength` on an arbitrary if it is a not a requirement of the algorithm  
+**❌ NEVER** specify any constraint on an arbitrary if it is not a requirement of the arbitrary, use defaults as much as possible  
+**👍 Prefer** specifying a `size: '-1'` if you feel that the algorithm will take very long on large inputs (by default fast-check generates up to 10 items, so only use `size` when clearly required)  
+Eg.: No `fc.string({maxLength: 5})` or `fc.array(arb, {maxLength: 8})` except being a string requirement
+
+**👍 Prefer** using the defaults of an arbitrary  
+**👎 Avoid** passing constraints that are not part of the requirements of the algorithm  
 Eg.: if the algorithm should accept any integer just ask an integer without specifying any min and max
```
dubzzz, line 206:
```suggestion
**❌ NEVER** specify any `maxLength` on an arbitrary if it is a not a requirement of the algorithm  
**👍 Prefer** specifying a `size: '-1'` if you feel that the algorithm will take very long on large inputs (by default fast-check generates up to 10 items, so only use `size` when clearly required)  
Eg.: No `fc.string({maxLength: 5})` or `fc.array(arb, {maxLength: 8})` except being a string requirement

**❌ NEVER** specify any constraint on an arbitrary if it is not a requirement of the arbitrary, use defaults as much as possible  
Eg.: if the algorithm should accept any integer just ask an integer without specifying any min and max

```

Issue #3942, opened 2023-05-31
Support for beforeEach/afterEach in fast-check/jest


hsommerlandreason:
## 🚀 Feature Request

I am having issues using sinon to mock things when running fast-check with jest, one cannot re-mock  something with sinon that has already been mocked.
So after each test the mocks have to be restored, this can be done via either the `afterEach` hook for a specific property test or with the global hooks when not using fast-check/jest.

But when using fast-check/jest these hooks are not available (at least not what i could find) and the global hooks don't seem to be applied. The beforeEach/afterEach hooks provided by jest does not work as they are not executed on every single check which they need to be.

This also applies for the `cleanup` method from `@testing-library/react` which will also need to be executed after every test.

## Motivation

This would be very appreciated as the omission of this feature from fast-check/jest makes testing several things a lot more complicated and annoying requiring either manually calling the cleanup stuff or constructing workarounds.

## Example
Here is a bit of a contrived example.
 ```typescript
 const whatever = {
  bob: (n: number) => n - 10
}

test.prop([fc.nat()])('failing', (n) => {
  sinon.replace(whatever, 'bob', sinon.fake.returns(n))
  expect(whatever.bob(n)).toEqual(n)
  // sinon.restore()  /* if this is called here the tests will work.
})

test('works', () => {
  fc.assert(fc.property(fc.nat(), (n) => {
    sinon.replace(whatever, 'bob', sinon.fake.returns(n))
    expect(whatever.bob(n)).toEqual(n)
    }).afterEach(() => sinon.restore())
  )
})
 ```
 
 This will fail with:
 ```
  Property failed after 2 tests
    { seed: 1576879931, path: "1:0", endOnFailure: true }
    Counterexample: [0]
    Shrunk 1 time(s)
    Got TypeError: Attempted to replace bob which is already replaced
 ```

dubzzz:
Definitely something worth adding. It would require te have a way to either hook into the beforeEach defined for Jest or expose one via fast-check and fill the one of Jest with it.

dmurvihill:
+1 for this, I'm trying to use with `jest-mock-axios` and having trouble resetting the mock between tests.

dubzzz:
For now, you can use https://fast-check.dev/docs/configuration/global-settings/#shared-settings with beforeEach or beforeEachAsync

ej-shafran:
Hey - any updates on this? If needed, I can try implementing this and open a pull request if I manage to...

[...]

PR #6586, opened 2026-02-10
🪝 Support for `beforeEach`/`afterEach` in `@fast-check/jest`

rushelex:
**Description**

This PR closes issue https://github.com/dubzzz/fast-check/issues/3942.

Currently, when using `it.prop` / `test.prop` in `@fast-check/jest`, `beforeEach` and `afterEach` hooks are only called once per test — not once per property run. This means that if a property executes 100 runs inside a single `it.prop`, hooks like `beforeEach` (commonly used for setup/teardown, mocking, DB transactions, etc.) only fire once instead of 100 times.

This PR makes `beforeEach`/`afterEach` fire for **each property run**, not just once per test. For example, with 1 regular `it()` + 1 regular `it()` + 1 `it.prop()` with `numRuns: 10`, `beforeEach`/`afterEach` will now fire 12 times total instead of 3.

**Strategy:**
- The test framework calls its own `beforeEach` before the test (covers 1st run)
- Between consecutive property runs: call framework's `afterEach` → `beforeEach` via `propertyInstance.beforeEach()` hook
- The test framework calls its own `afterEach` after the test (covers last run)
- Result: exactly N `beforeEach` + N `afterEach` for N property runs

**`@fast-check/jest`:** Accesses jest-circus internal state via `Object.getOwnPropertySymbols(globalThis)` matching `Symbol('JEST_STATE_SYMBOL')` string representation. Only works with jest-circus runner (default since Jest 27); graceful degradation for jest-jasmine2 (hooks fire once as before). Does not work in worker mode (property runs in a separate process without access to jest-circus state).

**Advanced**

- [x] Impacts: Generated values: no impact. Shrink values: no impact. Performance: minimal overhead from hook collection and invocation between property runs. Typings: no user-facing type changes — `import type` from `jest-circus` used internally only.


packages/jest/src/internals/TestWithPropRunnerBuilder.ts
```diff
@@ -61,6 +154,45 @@ export function buildTestWithPropRunner<Ts extends [any] | any[], TsParameters e
...
+          propertyInstance.beforeEach(async (previousHook: () => Promise<void>) => {
+            await previousHook();
+
+            if (isFirstRun) {
+              isFirstRun = false;
+              return;
+            }
+
+            // Between runs: close previous iteration, then open next
+            for (const hook of afterEachHooks) {
```
dubzzz: Why not hooking yourself in afterEach?

packages/jest/src/internals/TestWithPropRunnerBuilder.ts
```diff
@@ -61,6 +154,45 @@ export function buildTestWithPropRunner<Ts extends [any] | any[], TsParameters e
...
+        const beforeEachHooks = collectJestBeforeEachHooks(describeBlock);
+        const afterEachHooks = collectJestAfterEachHooks(describeBlock);
+
+        if (beforeEachHooks.length > 0 || afterEachHooks.length > 0) {
+          let isFirstRun = true;
+
+          propertyInstance.beforeEach(async (previousHook: () => Promise<void>) => {
+            await previousHook();
+
+            if (isFirstRun) {
```
dubzzz: The same logic goes for last run? Unfortunately we won't be able to know if it's the last or not

packages/jest/src/internals/TestWithPropRunnerBuilder.ts
```diff
@@ -61,6 +154,45 @@ export function buildTestWithPropRunner<Ts extends [any] | any[], TsParameters e
...
+      // Hook into fc's property lifecycle to call jest's beforeEach/afterEach
+      // between consecutive runs within the single test.
+      //
+      // Strategy (jest-circus only):
+      //   - jest calls its own beforeEach before the test (covers 1st run)
+      //   - fc property beforeEach between runs: jest afterEach + jest beforeEach
+      //   - jest calls its own afterEach after the test (covers last run)
+      //
+      // Result: exactly N beforeEach + N afterEach for N property runs.
+      const jestState = getJestCircusState();
```
dubzzz, line 166: Only working with Jest circus in CJS mode, ESM execution of Jest might not work

packages/vitest/src/internals/TestWithPropRunnerBuilder.ts
```diff
@@ -39,6 +85,44 @@ export function buildTestWithPropRunner<Ts extends [any] | any[], TsParameters e
...
+      //
+      // Strategy:
+      //   - vitest calls its own beforeEach before the test (covers 1st run)
+      //   - fc property beforeEach between runs: vitest afterEach + vitest beforeEach
+      //   - vitest calls its own afterEach after the test (covers last run)
+      //
+      // Result: exactly N beforeEach + N afterEach for N property runs.
+      const test = getCurrentTest();
+      const suite = test?.suite || test?.file;
+      if (suite) {
```
dubzzz: add explicit check: maybe !==undefined or !==null

packages/vitest/src/internals/TestWithPropRunnerBuilder.ts
dubzzz, line 1: Move vitest change in another PR so that each PR stays focused on one key change

packages/vitest/src/internals/TestWithPropRunnerBuilder.ts
```diff
@@ -1,12 +1,58 @@
...
 }
 
+/**
+ * Collect the suite chain from the given suite up to the file-level suite.
+ * Returns the chain with the innermost suite first.
+ */
+function getSuiteChain(suite: RunnerTestSuite): RunnerTestSuite[] {
+  const chain: RunnerTestSuite[] = [];
+  let current = suite;
+  while (current) {
```
dubzzz: prefer an explicit check, such as while current!==undefined/null

packages/vitest/src/internals/TestWithPropRunnerBuilder.ts
```diff
@@ -1,12 +1,58 @@
...
+    chain.push(current);
+    if ('filepath' in current) break;
+    current = current.suite || current.file;
```
dubzzz: prefer ??

packages/vitest/src/internals/TestWithPropRunnerBuilder.ts
```diff
@@ -1,12 +1,58 @@
...
+/**
+ * Collect beforeEach hooks in top-down order (parent suites first),
+ * matching vitest's callSuiteHook behavior for "beforeEach".
+ */
+function collectBeforeEachHooks(suite: RunnerTestSuite): ReturnType<typeof getHooks>['beforeEach'] {
+  const chain = getSuiteChain(suite);
+  const hooks: ReturnType<typeof getHooks>['beforeEach'] = [];
+  for (let i = chain.length - 1; i >= 0; i--) {
+    const h = getHooks(chain[i]);
+    if (h?.beforeEach) hooks.push(...h.beforeEach);
```
dubzzz: prefer explicit !== undefined check

packages/vitest/src/internals/TestWithPropRunnerBuilder.ts
```diff
@@ -1,12 +1,58 @@
...
+ * Collect afterEach hooks in bottom-up order (current suite first),
+ * with hooks within each suite in reverse registration order (stack behavior),
+ * matching vitest's callSuiteHook behavior for "afterEach".
+ */
+function collectAfterEachHooks(suite: RunnerTestSuite): ReturnType<typeof getHooks>['afterEach'] {
+  const chain = getSuiteChain(suite);
+  const hooks: ReturnType<typeof getHooks>['afterEach'] = [];
+  for (let i = 0; i < chain.length; i++) {
+    const h = getHooks(chain[i]);
+    if (h?.afterEach) hooks.push(...[...h.afterEach].reverse());
```
dubzzz: same !==undefined check

dubzzz: rather than building a cloned array that you reverse just do a simple loop doing the push in the reversed order

packages/vitest/src/internals/TestWithPropRunnerBuilder.ts
dubzzz: What would it mean to have no suite?

We could probably drop one level of if nesting by doing:
const beforeEachHooks = suite!==undefined?collectBeforeEachHooks(suite):[];
// same for after

packages/jest/src/internals/TestWithPropRunnerBuilder.ts
dubzzz: By doing so these after don't run with others... but actually I get why you did it

rushelex: What do you mean? Call `propertyInstance.afterEach` after jest `afterEach` (and vitest `afterEach`)?

packages/jest/src/internals/TestWithPropRunnerBuilder.ts
rushelex: Why do you think it won't work in ESM? `jest-circus` is included by default in Jest 27+ and runs independently of the CJS/ESM environment. `globalThis` and `Symbol('JEST_STATE_SYMBOL')` will also be available

packages/jest/src/internals/TestWithPropRunnerBuilder.ts
rushelex: > but actually I get why you did it

So, what should we do in this place? What do you think? 🙂

packages/jest/src/internals/TestWithPropRunnerBuilder.ts
rushelex: Yes, here's the lifecycle of fc and jest hook calls:

Current pass for N runs:
```
jeft.beforeEach         ← covers the 1st beforeEach
    fc.beforeEach: skip ← 1st run, do nothing
    run(v1)
    fc.afterEach: (noop)

    fc.beforeEach: afterEach + beforeEach ← closed the 1st run, opened the 2nd
    run(v2)
    fc.afterEach: (noop)

    fc.beforeEach: afterEach + beforeEach ← closed the 2nd run, opened the 3rd
    run(v3)
    fc.afterEach: (noop)
jest.afterEach          ← covers the last afterEach
```

Total: N beforeEach + N afterEach — the order is correct.

If a different approach were used, with a naive call to jest.beforeEach and jest.afterEach in fc.beforeEach and fc.afterEach, respectively, the final run would call afterEach twice:

Pass for N=3:
```
jest.beforeEach            ← bE #1 ✅
    fc.beforeEach: skip
    run(v1)
    fc.afterEach: jest.aE  ← aE #1 ✅

    fc.beforeEach: jest.bE ← bE #2 ✅
    run(v2)
    fc.afterEach: jest.aE  ← aE #2 ✅

    fc.beforeEach: jest.bE ← bE #3 ✅
    run(v3)
    fc.afterEach: jest.aE  ← aE #3 ✅
jest.afterEach             ← aE #4 ❌
```

Result: N beforeEach (ok), but N+1 afterEach (error) — jest.afterEach can't be suppressed, and it adds an extra call.

Maybe this also partially answers your question you asked here: https://github.com/dubzzz/fast-check/pull/6586#discussion_r2787637269

packages/vitest/src/internals/TestWithPropRunnerBuilder.ts
rushelex: `test` is undefined if called outside a test callback (for example, in a worker thread where the `_test` vitest state is not set).

`suite` is undefined if the test has neither a parent test suite nor a file reference (a top-level test or a detached test).

I'll add a comment to the code that describes this behavior.

rushelex: Also, we can't do `suite !== undefined ? collectBeforeEachHooks(suite) : []`, because both `test` and `suite` are required to call the hooks. We must ensure that `test` and `suite` are defined before running the hooks to ensure type safety.

packages/vitest/src/internals/TestWithPropRunnerBuilder.ts
rushelex: Okay, I'll do it

packages/vitest/src/internals/TestWithPropRunnerBuilder.ts
rushelex: Answered here: https://github.com/dubzzz/fast-check/pull/6586#discussion_r2788125265

packages/vitest/src/internals/TestWithPropRunnerBuilder.ts
rushelex: I replaced it with `while (true) { ... }` because current is always either `RunnerTestSuite` or `File` (`current.suite ?? current.file`), so it makes more sense to use an infinite loop with an explicit exit when we reach the file and the suites chain ends

packages/vitest/src/internals/TestWithPropRunnerBuilder.ts
rushelex: I've moved all Vitest related code into a separate PR: https://github.com/dubzzz/fast-check/pull/6587

dubzzz:
@rushelex now that we merged vitest one, we can come back to this one. I believe we should stick with the same philosophy as the pr we merged for vitest. So probably worth checking and updating code and test in the same spirit

rushelex:
@dubzzz Hi. Unfortunately, I don't have much time to work on this PR right now. I'd appreciate it if you could complete it, or we can wait until I have more time.

PR #6587, opened 2026-02-10
✨ Support for `beforeEach`/`afterEach` in `@fast-check/vitest`

rushelex:
**Description**

This PR closes issue https://github.com/dubzzz/fast-check/issues/3942.

This PR related to Jest PR https://github.com/dubzzz/fast-check/pull/6586.

Currently, when using `it.prop` / `test.prop` in `@fast-check/vitest`, `beforeEach` and `afterEach` hooks are only called once per test — not once per property run. This means that if a property executes 100 runs inside a single `it.prop`, hooks like `beforeEach` (commonly used for setup/teardown, mocking, DB transactions, etc.) only fire once instead of 100 times.

This PR makes `beforeEach`/`afterEach` fire for **each property run**, not just once per test. For example, with 1 regular `it()` + 1 regular `it()` + 1 `it.prop()` with `numRuns: 10`, `beforeEach`/`afterEach` will now fire 12 times total instead of 3.

**Strategy:**
- The test framework calls its own `beforeEach` before the test (covers 1st run)
- Between consecutive property runs: call framework's `afterEach` → `beforeEach` via `propertyInstance.beforeEach()` hook
- The test framework calls its own `afterEach` after the test (covers last run)
- Result: exactly N `beforeEach` + N `afterEach` for N property runs

**`@fast-check/vitest`:** Uses vitest internals (`getCurrentSuite()`, `getHooks()`) to collect hooks from the suite chain. Collects cleanup functions returned by `beforeEach` hooks and executes them before `afterEach` hooks between runs, matching vitest's cleanup-before-afterEach execution order.

**Advanced**

- [x] Impacts: Generated values: no impact. Shrink values: no impact. Performance: minimal overhead from hook collection and invocation between property runs. Typings: no user-facing type changes.

### DIFF
--- a/packages/vitest/src/internals/TestWithPropRunnerBuilder.ts
+++ b/packages/vitest/src/internals/TestWithPropRunnerBuilder.ts
@@ -38,15 +38,10 @@ function getSuiteChain(suite: RunnerTestSuite): RunnerTestSuite[] {
 function collectBeforeEachHooks(suite: RunnerTestSuite): ReturnType<typeof getHooks>['beforeEach'] {
   const chain = getSuiteChain(suite);
   const hooks: ReturnType<typeof getHooks>['beforeEach'] = [];
-
   for (let i = chain.length - 1; i >= 0; i--) {
     const h = getHooks(chain[i]);
-
-    if (h?.beforeEach !== undefined) {
-      hooks.push(...h.beforeEach);
-    }
+    hooks.push(...h.beforeEach);
   }
-
   return hooks;
 }
 


packages/vitest/test/vitest-fast-check.spec.ts
```diff
@@ -152,6 +155,225 @@ describe.each<DescribeOptions>([
...
+          afterAllVi(() => {
+            if (errors.length > 0) {
+              throw new Error('Hook order violations: ' + errors.join('; '));
+            }
+            if (beforeEachCount !== afterEachCount) {
+              throw new Error('bE/aE mismatch: bE=' + beforeEachCount + ', aE=' + afterEachCount);
+            }
+            if (beforeEachCount < 2) {
+              throw new Error('Expected at least 2 runs (including shrinks), got: ' + beforeEachCount);
+            }
```
dubzzz: My not be true... Actually if first run is with int=10, we will stop without shrinking

packages/vitest/src/internals/TestWithPropRunnerBuilder.ts
```diff
@@ -39,7 +107,73 @@ export function buildTestWithPropRunner<Ts extends [any] | any[], TsParameters e
...
+      // Result: exactly N beforeEach + N afterEach for N property runs.
+      const test = getCurrentTest();
+      const suite = test?.suite || test?.file;
+
+      // `test` is undefined when called outside a test callback (e.g. in a worker thread
+      // where vitest's _test state is not set).
+      // `suite` is undefined when the test has neither a parent suite nor a file
+      // reference (top-level or detached test).
+      if (test === undefined || suite === undefined) {
+        return;
```
dubzzz: Probably worth throwing, with a message explaining what you put as comment

packages/vitest/src/internals/TestWithPropRunnerBuilder.ts
```diff
@@ -39,7 +107,73 @@ export function buildTestWithPropRunner<Ts extends [any] | any[], TsParameters e
...
+
+          // Between runs: close previous iteration, then open next.
+          // Vitest calls beforeEach cleanup functions before afterEach hooks,
+          // so we replicate that order here.
+          for (let i = pendingCleanups.length - 1; i >= 0; i--) {
+            await pendingCleanups[i]();
+          }
+          // Clear the array in-place so the same reference stays valid
+          // inside the closure across iterations.
+          pendingCleanups.length = 0;
```
dubzzz: I'd drop a few comments to ease future reading

```suggestion
          // Vitest calls beforeEach cleanup functions before afterEach hooks, so we replicate that order here.
          for (let i = pendingCleanups.length - 1; i >= 0; i--) {
            await pendingCleanups[i]();
          }
          pendingCleanups.length = 0;
```

packages/vitest/src/internals/TestWithPropRunnerBuilder.ts
```diff
@@ -39,7 +107,73 @@ export function buildTestWithPropRunner<Ts extends [any] | any[], TsParameters e
...
       await fc.assert(propertyInstance, customParams);
+
+      // Run remaining cleanup functions from the last manual beforeEach call.
+      // Vitest supports returning a cleanup function from beforeEach; the first
+      // iteration's cleanup is handled by vitest's own afterEach lifecycle, but
+      // cleanups from iterations 2..N are captured and invoked by us.
+      for (let i = pendingCleanups.length - 1; i >= 0; i--) {
+        await pendingCleanups[i]();
+      }
     },
```
dubzzz: So if I get it well, clean-ups of the first beforeEach will be executed at the end by vitest. This manual clean-up is cleaning the last run within fast-check (if there was one not being the first).

dubzzz: Said differently we might have:
- beforeEach#1 done by vitest
- beforeEach#2 done by beforeEach of fast-check
- cleanUp of beforeEach#2 done by beforeEach of fast-check
- beforeEach#3 done by beforeEach of fast-check
- cleanUp of beforeEach#3 done by the manual clean-up loop after assert
- cleanUp of beforeEach#1 done by vitest

packages/vitest/test/vitest-fast-check.spec.ts
rushelex: I reduced the threshold to 1 since it is impossible to guarantee the presence of shrinks with an arbitrary initial value.

packages/vitest/src/internals/TestWithPropRunnerBuilder.ts
rushelex: Yes, you're right. Cleanup#1 is held by vitest in a local variable of runTest and runs after vitest's own afterEach — we cannot intercept it.

packages/vitest/src/internals/TestWithPropRunnerBuilder.ts
```diff
@@ -1,12 +1,80 @@
...
+ * matching vitest's callSuiteHook behavior for "beforeEach".
+ */
+function collectBeforeEachHooks(suite: RunnerTestSuite): ReturnType<typeof getHooks>['beforeEach'] {
+  const chain = getSuiteChain(suite);
+  const hooks: ReturnType<typeof getHooks>['beforeEach'] = [];
+
+  for (let i = chain.length - 1; i >= 0; i--) {
+    const h = getHooks(chain[i]);
+
+    if (h?.beforeEach !== undefined) {
```
dubzzz: From a typing point of view, neither h, nor beforeEach could be undefined. Why this check?

packages/vitest/src/internals/TestWithPropRunnerBuilder.ts
```diff
@@ -1,12 +1,80 @@
...
+ * matching vitest's callSuiteHook behavior for "afterEach".
+ */
+function collectAfterEachHooks(suite: RunnerTestSuite): ReturnType<typeof getHooks>['afterEach'] {
+  const chain = getSuiteChain(suite);
+
+  const hooks: ReturnType<typeof getHooks>['afterEach'] = [];
+
+  for (let i = 0; i < chain.length; i++) {
+    const h = getHooks(chain[i]);
+    if (h?.afterEach !== undefined) {
```
dubzzz: From a typing point of view neither h nor afterEach can be undefined, why this check?

packages/vitest/src/internals/TestWithPropRunnerBuilder.ts
```diff
@@ -39,7 +107,72 @@ export function buildTestWithPropRunner<Ts extends [any] | any[], TsParameters e
...
+      if (test === undefined) {
+        throw new Error(
+          'Could not find the running test context. Make sure your property-based test is defined inside a vitest test() or it() block. Running outside a standard vitest test callback (e.g. in a worker thread) is not supported.',
+        );
+      }
+      if (suite === undefined) {
+        throw new Error(
+          'Could not find a parent suite or file for the current test. Make sure your test is defined inside a describe() block or a test file, not as a standalone detached test.',
+        );
+      }
```
dubzzz, line 114:
```suggestion
      const test = getCurrentTest();
      if (test === undefined) {
        throw new Error(
          'Could not find the running test context. Make sure your property-based test is defined inside a vitest test() or it() block. Running outside a standard vitest test callback (e.g. in a worker thread) is not supported.',
        );
      }
      
      const suite = test.suite ?? test.file;
      if (suite === undefined) {
        throw new Error(
          'Could not find a parent suite or file for the current test. Make sure your test is defined inside a describe() block or a test file, not as a standalone detached test.',
        );
      }
```

packages/vitest/test/vitest-fast-check.spec.ts
```diff
@@ -152,6 +155,225 @@ describe.each<DescribeOptions>([
...
+
+      // Act
+      const out = await runSpec(specDirectory);
+
+      // Assert
+      expectPass(out);
+    });
+
+    it.concurrent(
+      `should never call beforeEach/afterEach twice per run in ${runnerName}.prop`,
```
dubzzz, line 209: Actually the test does not really does that. This one mostly checks that we don't have a too large diff between number of beforeEach runs vs number of afterEach runs

packages/vitest/test/vitest-fast-check.spec.ts
```diff
@@ -152,6 +155,225 @@ describe.each<DescribeOptions>([
...
+
+        // Act
+        const out = await runSpec(specDirectory);
+
+        // Assert
+        expectPass(out);
+      },
+    );
+
+    it.concurrent(
```
dubzzz, line 258: IMO no real need to check for shrinking vs no-shrinking. Having the no-shrinking cover is good enough for this feature. So worth dropping this test case

packages/vitest/test/vitest-fast-check.spec.ts
```diff
@@ -152,6 +155,225 @@ describe.each<DescribeOptions>([
...
```
dubzzz, line 305: I'd rather group the two last tests into one doing something like:

Have an array being empty.
In beforeEach, push "beforeEach('runId')" in the array.
In run, push "predicate('runId')" in the array.
In clean-up of beforeEach, push "clean-up('runId')" in the array.
In afterEach, push "afterEach('runId')" in the array.

Maybe limit yourself to 3 runs and in afterAll expect an hardcoded array. Such test would have the benefit to capture all remaining things including relative ordering of calls and also the thing around clean-up of 1st run not being called exactly when it should be.

packages/vitest/src/internals/TestWithPropRunnerBuilder.ts
```diff
@@ -1,12 +1,80 @@
...
+    current = current.suite ?? current.file;
+  }
+
+  return chain;
+}
+
+/**
+ * Collect beforeEach hooks in top-down order (parent suites first),
+ * matching vitest's callSuiteHook behavior for "beforeEach".
+ */
```
dubzzz:
```suggestion
/** Collect beforeEach hooks in top-down order (parent suites first) as vitest does */
```

packages/vitest/src/internals/TestWithPropRunnerBuilder.ts
```diff
@@ -1,12 +1,80 @@
...
+  }
+
+  return hooks;
+}
+
+/**
+ * Collect afterEach hooks in bottom-up order (current suite first),
+ * with hooks within each suite in reverse registration order (stack behavior),
+ * matching vitest's callSuiteHook behavior for "afterEach".
+ */
```
dubzzz:
```suggestion
/** Collect afterEach hooks in bottom-up order (current suite first) as vitest does */
```

packages/vitest/src/internals/TestWithPropRunnerBuilder.ts
```diff
@@ -1,12 +1,80 @@
...
+
+  for (let i = chain.length - 1; i >= 0; i--) {
+    const h = getHooks(chain[i]);
+
+    if (h?.beforeEach !== undefined) {
+      hooks.push(...h.beforeEach);
+    }
+  }
+
+  return hooks;
```
dubzzz, line 45:
```suggestion
  const chain = getSuiteChain(suite);
  const hooks: ReturnType<typeof getHooks>['beforeEach'] = [];
  for (let i = chain.length - 1; i >= 0; i--) {
    const h = getHooks(chain[i]);
    hooks.push(...h.beforeEach);
  }
  return hooks;
```

packages/vitest/src/internals/TestWithPropRunnerBuilder.ts
```diff
@@ -1,12 +1,80 @@
...
+  for (let i = 0; i < chain.length; i++) {
+    const h = getHooks(chain[i]);
+    if (h?.afterEach !== undefined) {
+      for (let j = h.afterEach.length - 1; j >= 0; j--) {
+        hooks.push(h.afterEach[j]);
+      }
+    }
+  }
+
+  return hooks;
```
dubzzz, line 58:
```suggestion
  const chain = getSuiteChain(suite);
  const hooks: ReturnType<typeof getHooks>['afterEach'] = [];
  for (let i = 0; i < chain.length; i++) {
    const h = getHooks(chain[i]);
    for (let j = h.afterEach.length - 1; j >= 0; j--) {
      hooks.push(h.afterEach[j]);
    }
  }
  return hooks;
```

packages/vitest/src/internals/TestWithPropRunnerBuilder.ts
```diff
@@ -39,7 +107,72 @@ export function buildTestWithPropRunner<Ts extends [any] | any[], TsParameters e
...
+        });
+      }
+
       await fc.assert(propertyInstance, customParams);
+
+      // Cleanup from the last iteration (N). Cleanups 2..N-1 run inside
+      // fc's beforeEach hook above. Cleanup#1 is held by vitest in a local
+      // variable of runTest and runs after vitest's own afterEach — we cannot
+      // intercept it.
+      for (let i = pendingCleanups.length - 1; i >= 0; i--) {
```
dubzzz, line 156: With this trick, I have the feeling that we might not support correctly things such as:

```ts
beforeEach(() => {
  console.log("before");
});

test("abc", () => {
  console.log("test");
  throw new Error("a");
});

afterEach(() => {
  console.log("after");
});
```

In such context vitest prints the three logs. In our case if fc.assert fails, I feel that we might not execute clean-ups properly...

dubzzz: Not blocking at all for the first iteration on this feature

packages/vitest/test/vitest-fast-check.spec.ts
dubzzz: Regarding my comment https://github.com/dubzzz/fast-check/pull/6587/changes#r2819152307, if we address it (maybe in another PR) it would be great to have a test with the same idea as the one described above but in the context of a failure. This time no way to hardcode the output but we could in the afterAll test its shape.

packages/vitest/src/internals/TestWithPropRunnerBuilder.ts
dubzzz: Solved by https://github.com/dubzzz/fast-check/pull/6587/changes/e8ee29ac95cdeacbc4cf2e68b346fe8e8e397da4

packages/vitest/src/internals/TestWithPropRunnerBuilder.ts
dubzzz: Solved by https://github.com/dubzzz/fast-check/pull/6587/changes/f710bbc94b9dc35e355b86aaf4dc8a741670db0d

packages/vitest/src/internals/TestWithPropRunnerBuilder.ts
dubzzz: Solved by https://github.com/dubzzz/fast-check/pull/6587/changes/e8ee29ac95cdeacbc4cf2e68b346fe8e8e397da4

packages/vitest/src/internals/TestWithPropRunnerBuilder.ts
dubzzz: Solved by https://github.com/dubzzz/fast-check/pull/6587/changes/c1059208f243188fb7538e0b77312ee91a7a0f5d

packages/vitest/src/internals/TestWithPropRunnerBuilder.ts
dubzzz: Solved by https://github.com/dubzzz/fast-check/pull/6587/changes/7e308e897322793bcc5b298bf501645f5c3dc1f5

packages/vitest/src/internals/TestWithPropRunnerBuilder.ts
dubzzz: Solved

dubzzz:
Let's go for it! Thanks for the contribution.

I'm gonna address on separate PRs the remaining points.

packages/vitest/test/vitest-fast-check.spec.ts
dubzzz: @claude could you try to take that one? one a PR doing the fix I'm asking

packages/vitest/src/internals/TestWithPropRunnerBuilder.ts
dubzzz: @claude could you try to take that one? one a PR doing the fix I'm asking

dubzzz:
@rushelex I had to recreate the PR to deal with last few blockers and totally unlock it. The new PR is accessible at https://github.com/dubzzz/fast-check/pull/6695. No worry you'll still be marked as author on the finally merged PR and still be added as a contributor to the repository.

My plan is to address the few last TODOs: tests and bugs in term of errors before merging it.

dubzzz:
Closing as already merged

PR #6679, opened 2026-03-04
⚡️ Bump pure-rand to v8

dubzzz:
## Description

Bump our dependency to head for the last generation of this dependency.

## Checklist

### DIFF
--- a/.changeset/old-cobras-teach.md
+++ b/.changeset/old-cobras-teach.md
@@ -1,0 +1,5 @@
+---
+"fast-check": patch
+---
+
+⚡️ Bump pure-rand to v8


packages/fast-check/src/random/generator/Random.ts
dubzzz, line 1: For another PR:

- deprecate next(n)
- deprecate nextInt()
- use uniformFloat64 from pure-rand

dubzzz: Plus:

- move worker to 8 too

dubzzz:
@claude mark Random::next(n) and Random::nextInt() [the no arg only] as deprecated

dubzzz:
@claude move worker package to pure-rand 8 as we just did in this PR for fast-check

PR #6695, opened 2026-03-07
✨ Support for `beforeEach`/`afterEach` in `@fast-check/vitest`

### LINKED ISSUE #3942: Support for beforeEach/afterEach in fast-check/jest

dubzzz:
## Description

Original PR with full description available at #6587.

Fixes #3942

## Checklist

### DIFF
--- a/packages/vitest/src/internals/TestWithPropRunnerBuilder.ts
+++ b/packages/vitest/src/internals/TestWithPropRunnerBuilder.ts
@@ -146,14 +146,16 @@ export function buildTestWithPropRunner<Ts extends [any] | any[], TsParameters e
         });
       }
 
-      await fc.assert(propertyInstance, customParams);
-
-      // Cleanup from the last iteration (N). Cleanups 2..N-1 run inside
-      // fc's beforeEach hook above. Cleanup#1 is held by vitest in a local
-      // variable of runTest and runs after vitest's own afterEach — we cannot
-      // intercept it.
-      for (let i = pendingCleanups.length - 1; i >= 0; i--) {
-        await pendingCleanups[i]();
+      try {
+        await fc.assert(propertyInstance, customParams);
+      } finally {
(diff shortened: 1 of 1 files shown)


dubzzz:
@claude please add a note on before each and after each support in fast-check vitest. On its readme. Open a PR to add such details. Also talk about limitations

PR #6697, opened 2026-03-07
📝(vitest) Document beforeEach/afterEach hook support and limitations

dubzzz:
Adds documentation for `beforeEach`/`afterEach` hook support in `@fast-check/vitest` README, including usage examples and known limitations.

Related to #6695

Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -525,6 +525,13 @@
       "avatar_url": "https://avatars.githubusercontent.com/u/1519098?v=4",
       "profile": "https://github.com/josephjunker",
       "contributions": ["doc"]
+    },
+    {
+      "login": "rushelex",
+      "name": "Aleksey Shelementev",
+      "avatar_url": "https://avatars.githubusercontent.com/u/41122157?v=4",
+      "profile": "https://github.com/rushelex",
+      "contributions": ["code"]
     }
   ],
   "contributorsPerLine": 7
--- a/.changeset/wise-news-drum.md
+++ b/.changeset/wise-news-drum.md
@@ -1,0 +1,5 @@
+---
+"fast-check": patch
+---
+
+👥 Add rushelex as code contributor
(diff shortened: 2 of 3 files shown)


packages/vitest/README.md
```diff
@@ -90,6 +90,42 @@ If you want to forward custom parameters to fast-check, `test.prop` accepts an o
 
 `@fast-check/vitest` also comes with `.only`, `.skip`, `.todo` and `.concurrent` from vitest. It also accepts more complex ones such as `.concurrent.skip`.
 
+### Hooks: `beforeEach` / `afterEach`
```
dubzzz: @claude move this section before the advanced one. Ho straight to the goal: before each and after each from vitest gets automatically wired into predicates. As such they get called respectively before (and after) each execution of predicate. Keep a simple to understand warning stating that before each clean ups may occur later than expected (if used) but that before and after will always be called before and after each predicate. The clean up though of the before for the first exec of predicate will be delayed to the end (it's the only warning). Anyway if predicate runs n times, there will be n call to before each before it, n to after after it, and n clean up with the one from first run being late

packages/vitest/README.md
```diff
@@ -84,6 +84,32 @@ test.prop({ a: fc.string(), b: fc.string(), c: fc.string() })('should detect the
 
 Please note that the properties accepted by `@fast-check/vitest` as input can either be synchronous or asynchronous (even just `PromiseLike` instances).
 
+### Hooks: `beforeEach` / `afterEach`
+
+Vitest's `beforeEach` and `afterEach` hooks are automatically wired into predicates. They are called respectively before and after each execution of the predicate. If a predicate runs _n_ times, `beforeEach` will be called _n_ times before it and `afterEach` _n_ times after it.
+
+```javascript
```
dubzzz: @claude no need for the code example, at worst just tell it with words or extend your sentences before to capture that if unclear

packages/vitest/README.md
```diff
@@ -84,6 +84,10 @@ test.prop({ a: fc.string(), b: fc.string(), c: fc.string() })('should detect the
 
 Please note that the properties accepted by `@fast-check/vitest` as input can either be synchronous or asynchronous (even just `PromiseLike` instances).
 
+### Hooks: `beforeEach` / `afterEach`
```
dubzzz: @claude rename, maybe tell about native or seemless integration with vitest's before and aftereach

PR #6699, opened 2026-03-07
fix(jsonValue): ensure generated doubles survive JSON round-trip

### LINKED ISSUE #6095: Possible error in values built from `jsonValue`
Details at: https://github.com/dubzzz/fast-check/actions/runs/16734594221/job/47370573425?pr=6088

```txt
 FAIL   fast-check  test/unit/arbitrary/jsonValue.spec.ts > jsonValue (integration) > should only produce correct values
Error: Property failed after 19 tests
{ seed: 773276696, path: "18", endOnFailure: true }
Counterexample: [-713157037,1566263274,Stream(9,9,6,8,13,13,10,16,17,1,5,20,1,10,18,6,15,3,17,8,9,16,11,19,7,3,18,12,2,12,13,19,19…),undefined]
Shrunk 0 time(s)

Hint: Enable verbose mode in order to have the list of all failing values encountered during the run
 ❯ buildError lib/check/runner/utils/RunDetailsFormatter.js:156:19
    154|     }
    155|     const ErrorWithCause = Error;

    156|     const error = new ErrorWithCause(errorMessage, { cause: out.errorI…
       |                   ^
    157|     if (!('cause' in error)) {
    158|         safeObjectAssign(error, { cause: out.errorInstance });
 ❯ throwIfFailed lib/check/runner/utils/RunDetailsFormatter.js:165:11
 ❯ reportRunDetails lib/check/runner/utils/RunDetailsFormatter.js:178:16
 ❯ Module.assert lib/check/runner/Runner.js:61:25
 ❯ assertProduceCorrectValues test/unit/arbitrary/__test-helpers__/ArbitraryAssertions.ts:88:6
 ❯ test/unit/arbitrary/jsonValue.spec.ts:64:5

Caused by: Error: Expect: {"A":"","TA \\>":3.340473894173502e-256} to be a correct value

Got error: AssertionError: expected { A: '', …(1) } to deeply equal { A: '', …(1) }
```

dubzzz:
The `jsonValue` arbitrary generates doubles that should be valid JSON values. However, subtle precision mismatches between `String(v)` (used by `fc.stringify`) and `JSON.parse` can cause some doubles to not perfectly round-trip.

This fix maps generated doubles through `JSON.parse(String(v))` to ensure they are exactly the values that `JSON.parse` would produce.

Fixes #6095

Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/packages/fast-check/src/arbitrary/_internals/helpers/JsonConstraintsBuilder.ts
+++ b/packages/fast-check/src/arbitrary/_internals/helpers/JsonConstraintsBuilder.ts
@@ -1,5 +1,4 @@
 import type { Arbitrary } from '../../../check/arbitrary/definition/Arbitrary.js';
-import { String } from '../../../utils/globals.js';
 import { boolean } from '../../boolean.js';
 import { constant } from '../../constant.js';
 import { double } from '../../double.js';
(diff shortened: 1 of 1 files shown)


dubzzz:
The fix is not a fix but is breaking the correct behaviour instead of fixing the right place

PR #6866, opened 2026-04-15
✨ Reversible `json` arbitrary

dubzzz:
## Description

The `json()` arbitrary now supports `canShrinkWithoutContext`. This means users can shrink any JSON string — even ones not originally generated by fast-check — making it usable in scenarios like replaying or combining arbitraries where the original generation context is unavailable.

Previously, `json()` could only shrink values it had generated itself. With this change, you can pass any valid JSON string to `json().canShrinkWithoutContext(value)` and get `true`, enabling fast-check to shrink it toward simpler JSON values.

Under the hood, this is achieved by providing `JSON.parse` as the unmapper for the `.map()` call, making the arbitrary reversible.

Note: `assertShrinkProducesSameValueWithoutInitialContext` is not applicable because `JSON.stringify(-0)` produces `"0"` but `JSON.parse("0")` gives `0` (not `-0`), making the roundtrip lossy for `-0`.

## Checklist

### DIFF
--- a/.changeset/moody-days-count.md
+++ b/.changeset/moody-days-count.md
@@ -1,0 +1,5 @@
+---
+"fast-check": minor
+---
+
+Make json() arbitrary reversible by providing JSON.parse as unmapper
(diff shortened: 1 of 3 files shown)


dubzzz:
@claude use pr template, propose something matching the template

dubzzz:
@claude in your PR description, focus first on what will be the change for the final user. You mostly talk of implementation details. We want reader to understand what it will bring to them

PR #6872, opened 2026-04-16
🔨 Rework hooks for Claude Code

dubzzz:
## Description

Rework Claude Code hook commands to use absolute paths via `$CLAUDE_PROJECT_DIR` instead of relative paths with explicit `bash` invocation. This ensures hooks resolve to the correct script path regardless of the current working directory.

Also makes `.claude/npm-to-pnpm.sh` executable so it can be invoked directly without an explicit `bash` prefix.

## Checklist


dubzzz:
@claude is it correct. maybe you can confirm by reading https://code.claude.com/docs/en/hooks#hooks-reference

dubzzz:
@claude edit pr description following pr template

PR #6878, opened 2026-04-17
👷 Scope permissions of clean-caches

dubzzz:
## Description

Move the `actions: write` permission from the workflow top level into the single `clean_caches` job. This fixes the OpenSSF Scorecard Token-Permissions warning about an elevated top-level permission without changing behavior.

## Checklist


dubzzz:
@claude use pr template

PR #6877, opened 2026-04-17
👷 Lowercase discussion_category_name to "announcements"

dubzzz:
## Description

Lowercase `discussion_category_name` from `Announcements` to `announcements` across all per-package release jobs in `build-status.yml`.

This fixes the `softprops/action-gh-release` error:
> HttpError: Discussion could not be created. Make sure you passed a valid category name.

The category name matching appears to be case-sensitive, and the actual GitHub Discussions category is lowercase.

## Checklist

— _Don't delete this checklist and make sure you do the following before opening the PR_

- [x] I have a full understanding of every line in this PR — whether the code was hand-written, AI-generated, copied from external sources or produced by any other tool
- [x] I flagged the impact of my change (minor / patch / major) either by running `pnpm run bump` or by following the instructions from the changeset bot
- [x] I kept this PR focused on a single concern and did not bundle unrelated changes
- [x] I followed the [gitmoji](https://gitmoji.dev/) specification for the name of the PR, including the package scope (e.g. `🐛(vitest) Something...`) when the change targets a package other than `fast-check`
- [x] I added relevant tests and they would have failed without my PR (when applicable)

PR #6932, opened 2026-05-03
👷 Grant `discussions: write` to release jobs

dubzzz:
The release publish jobs use softprops/action-gh-release with
discussion_category_name to link an Announcements discussion to each
release, but the GITHUB_TOKEN was scoped to contents/id-token/attestations
write only. With discussions permission missing, GitHub's release update
endpoint surfaces the misleading error "Discussion could not be created.
Make sure you passed a valid category name." (and the action retries
exhausted), even though the category exists and the name is valid.

This is also why no past release ever had a linked discussion regardless
of the "Announcements" / "announcements" casing tried in #6877.

Add discussions: write to every per-package publish job so the linked
discussion can actually be created when finalizing the release.

### DIFF
--- a/.github/workflows/build-status.yml
+++ b/.github/workflows/build-status.yml
@@ -595,6 +595,7 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       contents: write
+      discussions: write
       id-token: write
       attestations: write
     steps:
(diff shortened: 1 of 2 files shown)


dubzzz:
@claude fix zizmor job

PR #6934, opened 2026-05-03
✅ Tolerate `\p{...}` value drift in docs tests

dubzzz:
When SKIP_DOCUMENTATION is set, rather than skipping the whole docs test
suite, only reuse the original generated values for `stringMatching` snippets
that rely on `\p{...}`. Their output depends on the runtime's Unicode version,
so the comparison is relaxed there while every other snippet is still
re-executed and compared as usual.

https://claude.ai/code/session_01StoxgCx8Nnwjk421jxrcwL


packages/fast-check/test/e2e/documentation/Docs.md.spec.ts
```diff
@@ -42,14 +42,47 @@ describe('Docs.md', () => {
...
+  const originalChunks = originalContent.split(separator);
+  const refreshedChunks = refreshedContent.split(separator);
+  if (originalChunks.length !== refreshedChunks.length) {
+    return false;
+  }
+  return originalChunks.some(
+    (originalChunk, index) =>
+      index !== 0 &&
+      originalChunk !== refreshedChunks[index] &&
+      /fc\.stringMatching\([^)]*\\p\{/.test(refreshedChunks[index - 1]),
```
dubzzz:
```suggestion
      /fc\.stringMatching\([^)]*\\[pP]\{/.test(refreshedChunks[index - 1]),
```

PR #6959, opened 2026-05-08
👷 Announce releases on Bluesky

dubzzz:
## Description

Adds a new GitHub Actions workflow that automatically publishes a short
Bluesky post each time a release is published, linking back to the
GitHub Release discussion that `softprops/action-gh-release` already
creates under the `announcements` category. The post text follows the
shape `🚀 New release: <label>\n\nRead the announcement and changelog: <url>`,
with `<label>` derived from the git tag (`v3.5.0` → `fast-check v3.5.0`,
`vitest/v1.2.0` → `@fast-check/vitest v1.2.0`), so it works uniformly
for `fast-check` and the six sibling packages. The workflow skips when
the release has no linked discussion. Two new repository secrets are
required: `BLUESKY_IDENTIFIER` (the bot's Bluesky handle) and
`BLUESKY_APP_PASSWORD` (an App Password generated in Bluesky settings);
without them the job fails loudly so the issue surfaces in CI rather
than silently dropping announcements.

The workflow is a standalone file triggered on `release: published`
rather than an extra step appended to each of the seven
`publish_package_*` jobs in `build-status.yml` — this keeps the social
announcement decoupled from the publishing pipeline, fires exactly once
per release event, and reuses `release.discussion_url` straight from
the webhook payload (no extra GitHub API call). The post script lives
in `.github/workflows/scripts/post-bluesky.mjs` (ESM with top-level
await, matching the existing `.cjs` helpers in that folder) and uses
the official `@atproto/api` client (`CredentialSession` + `Agent` +
`RichText`) — `RichText.detectFacets` handles URL facet construction,
which is otherwise tricky because Bluesky requires UTF-8 byte offsets
rather than character offsets. `@atproto/api` is added to root
`devDependencies` so it's locked in `pnpm-lock.yaml`, and the workflow
runs `pnpm install --frozen-lockfile` before invoking the script,
matching the project's existing supply-chain hardening (frozen
lockfile, pinned action SHAs). Impact level: patch — CI-only addition,
no behavioral change to any published package. Scope is single-concern
(workflow + its supporting script + the new dev dependency it pulls
in). No automated tests are added: the surface is a thin glue script
against the Bluesky API and GitHub release events, neither of which is
practical to test in this repo's vitest suite; the script was instead
smoke-tested locally to confirm env-var validation, tag-to-label
formatting for all seven packages, ESM loading of `@atproto/api`, and
that the credential session reaches the AT Protocol endpoint.

## Checklist


.github/workflows/scripts/post-bluesky.mjs
```diff
@@ -0,0 +1,42 @@
+// @ts-check
+import { Agent, CredentialSession, RichText } from '@atproto/api';
+
+const identifier = requireEnv('BLUESKY_IDENTIFIER');
+const password = requireEnv('BLUESKY_APP_PASSWORD');
+const discussionUrl = requireEnv('DISCUSSION_URL');
+const releaseTag = requireEnv('RELEASE_TAG');
+
+const session = new CredentialSession(new URL('https://bsky.social'));
```
dubzzz:
```suggestion
const session = new CredentialSession(new URL('https://eurosky.social'));
```

.github/workflows/scripts/post-bluesky.mjs
```diff
@@ -0,0 +1,42 @@
+// @ts-check
+import { Agent, CredentialSession, RichText } from '@atproto/api';
+
+const identifier = requireEnv('BLUESKY_IDENTIFIER');
```
dubzzz, line 5:
```suggestion
const url = requireEnv('BLUESKY_URL');
const identifier = requireEnv('BLUESKY_IDENTIFIER');
```

.github/workflows/scripts/post-bluesky.mjs
dubzzz:
```suggestion
const session = new CredentialSession(new URL(url));
```

PR #7010, opened 2026-05-20
🐛 Fix latent state-sharing bug in entityGraph

dubzzz:
## Description

> AI-agent disclosure: this PR was authored by an automated agent (Claude Code, claude-opus-4-7) and has not been line-by-line reviewed by a human before submission.

For consumers of `entityGraph` with inverse relationships, this fixes a latent correctness issue in `OnTheFlyLinksForEntityGraphArbitrary`: back-references could be pushed onto an `index` array still shared by reference with an earlier production state, mutating data the copy-on-write contract considers immutable. In the current `generate` loop the previous state is dropped right after `commit`, so the leak is not observable through generated values today, but it leaves the door open as soon as anything (future shrinking, replay, debugging) holds on to an earlier state.

`draftNextProductionState` already clones at two levels — the per-type array (`getOrCreateProducedLinksFor`) and the per-entity links instance (`getOrCreateLinksFor`) — but the relation entry living inside an entity's links was still shared by reference with the previous state. When `appendBackReference` pushed to `.index`, it was pushing into that shared array.

This PR extends the CoW chain with a third level: `getOrCreateRelationFor(type, indexInType, property)` returns the entity's relation entry, cloning it the first time it is touched in this draft and shallow-cloning `.index` when it is an array (checked via `typeof ... === 'object'`, since `index` is `number[] | number | undefined`). `appendBackReference` now goes through that helper, so the push only ever lands in the current state's own array.

Impact: patch — internal refactor scoped to the entity-graph arbitrary, no public API change, single concern. Existing unit tests in `test/unit/arbitrary/_internals/OnTheFlyLinksForEntityGraphArbitrary.spec.ts` and `test/unit/arbitrary/entityGraph.spec.ts` continue to pass; no dedicated regression test was added because the leak isn't observable from the generated value alone in the current loop — happy to add one if you'd like a state-retention scenario.

## Checklist

---
_Generated by [Claude Code](https://claude.ai/code/session_01R8uyr2RcxHjweK5xsyEzcC)_

### DIFF
--- a/packages/fast-check/src/arbitrary/_internals/OnTheFlyLinksForEntityGraphArbitrary.ts
+++ b/packages/fast-check/src/arbitrary/_internals/OnTheFlyLinksForEntityGraphArbitrary.ts
@@ -179,13 +179,13 @@ function draftNextProductionState<TEntityFields, TEntityRelations extends Entity
     // `links` is the brand-new instance from `createEmptyLinksInstanceFor` and nothing is shared with a previous state.
     const originalEntity = producedLinks[type][indexInType];
     if (originalEntity !== undefined && links[property] === originalEntity[property]) {
-      const clonedRelation = { ...links[property] };
-      // When `index` is an array it is still shared with the previous state and must be cloned.
-      // In every other case (either `number` or `undefined`) it already got safely "cloned" by the spread.
-      if (typeof clonedRelation.index === 'object') {
-        clonedRelation.index = safeSlice(clonedRelation.index);
-      }
-      links[property] = clonedRelation;
+      const originalRelation = links[property];
+      // `index` is the only field that can carry a shared mutable reference: when it is an array we must
+      // shallow-clone it; in every other case (`number` or `undefined`) the primitive can be reused as-is.
+      links[property] = {
+        type: originalRelation.type,
+        index: typeof originalRelation.index === 'object' ? safeSlice(originalRelation.index) : originalRelation.index,
+      };
     }
     return links[property];
   }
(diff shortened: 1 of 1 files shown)


packages/fast-check/src/arbitrary/_internals/OnTheFlyLinksForEntityGraphArbitrary.ts
```diff
@@ -168,6 +168,23 @@ function draftNextProductionState<TEntityFields, TEntityRelations extends Entity
...
+  function getOrCreateRelationFor(type: keyof TEntityFields, indexInType: number, property: string) {
+    const links = getOrCreateLinksFor(type, indexInType) as Record<
+      string,
+      EntityLinks<TEntityFields, TEntityRelations>[keyof EntityLinks<TEntityFields, TEntityRelations>]
+    >;
+    const originalEntity = producedLinks[type][indexInType];
+    if (originalEntity !== undefined && links[property] === originalEntity[property]) {
+      const clonedRelation = safeObjectAssign({}, links[property]);
+      // When `index` is an array it is still shared with the previous state and must be cloned.
+      // In every other case it is either a primitive (`number` or `undefined`) — so already safely copied by the `Object.assign` above — or nothing left to do.
```
dubzzz:
```suggestion
      // In every other case (either `number` or `undefined`) it already got safely "cloned" by the `Object.assign`.
```

packages/fast-check/src/arbitrary/_internals/OnTheFlyLinksForEntityGraphArbitrary.ts
```diff
@@ -168,6 +168,24 @@ function draftNextProductionState<TEntityFields, TEntityRelations extends Entity
...
     return producedLinksForType[indexInType];
   }
+  function getOrCreateRelationFor(
+    type: keyof TEntityFields,
+    indexInType: number,
+    property: keyof TEntityRelations[keyof TEntityFields],
+  ) {
+    const links = getOrCreateLinksFor(type, indexInType);
+    const originalEntity = producedLinks[type][indexInType];
+    if (originalEntity !== undefined && links[property] === originalEntity[property]) {
```
dubzzz, line 179:
```suggestion
    if (links[property] === originalEntity[property]) {
```

PR #7020, opened 2026-05-23
⚡ Speed up `ConstantArbitrary` generate hot path

dubzzz:
## Description

`fc.constant(...)` and `fc.constantFrom(...)` are the cheapest arbitraries in fast-check — they're meant to be free. Two small tweaks bring them closer to that ideal without changing any observable behaviour.

### What changes

- **Pre-compute `hasAnyCloneable` once at construction.** Before this PR, every call to `generate` did `hasCloneMethod(value)` on the picked value, which probes for the `cloneMethod` symbol property. In a typical `constantFrom('a', 'b', 'c')` (or, especially, `constant(undefined)`), the values array can hold a mix of primitives, plain objects and cloneables — so the property-access pattern is megamorphic, defeating the inline cache. Moving the probe to construction time and storing one boolean lets `generate` collapse to the trivial path for the (overwhelmingly common) case where no value is cloneable.
- **Drop the `Set`-backed lookup for small arrays in `canShrinkWithoutContext`.** The `FastConstantValuesLookup` was lazily built on first call; for arrays of 8 or fewer values, a flat `Object.is` linear scan beats both the lazy `Set` construction and the subsequent lookup, because the `Set` has to special-case ±0. The cut-off is the typical size of `constantFrom` usage in user code.

### Background reading

- V8 inline caches & polymorphism budget: <https://mathiasbynens.be/notes/shapes-ics>.
- Why `for (i…) === Object.is(values[i], …)` beats a small `Set`: <https://v8.dev/blog/elements-kinds#smis-vs-doubles> + the `Set` slow-path discussion in <https://v8.dev/blog/hash-code>.
- "Hidden classes and inline caches in V8" (Marvin Hagemeister, _Speeding up the JS ecosystem_): the megamorphic-probe failure mode this PR side-steps: <https://marvinh.dev/blog/speeding-up-javascript-ecosystem-part-2/>.

### Ideas considered but dropped

A larger split into a dedicated `SingleConstantArbitrary` for `fc.constant(x)` was tried (≥+90% on tuples-of-constants, +35–150% on the constant suites) but is **not** included here — `constant.spec.ts` spies on `ConstantArbitrary` and asserts it is the class actually instantiated, so the split would break tests we don't want to change.

### Observable behaviour

No public API change. Same seed → same generated value. `canShrinkWithoutContext` returns the same booleans (the small-array branch handles ±0 identically to the `Set` branch). All three relevant suites pass unchanged:

- `test/unit/arbitrary/constant.spec.ts`
- `test/unit/arbitrary/constantFrom.spec.ts`
- `test/unit/arbitrary/_internals/ConstantArbitrary.spec.ts`

Total: 35/35 tests; broader `test/unit/arbitrary/` sweep: 1690/1690.

### Numbers

Median of 5 runs × 2 s, paired against `main`:

| Suite | Δ |
| --- | ---: |
| `constant('hello')` | **+48%** |
| `constant(undefined)` | **+50%** |
| `constant(null)` | **+47%** |
| `constant({…})` | **+52%** |
| `constantFrom('a'..'e')` | **+25%** |
| `constantFrom(true, false)` | **+42%** |
| `constantFrom(1)` | **+79%** |
| `constantFrom(... strings)` | **+33%** |
| `canShrinkWithoutContext(present)` | **+30%** |
| `tuple(constantFrom, constant)` | **+33%** |

Re-runnable benchmark:

```js
// bench-constant.mjs
import { Random, constant, constantFrom, tuple } from 'fast-check';

  let n = 0;
    n += 256;
  }
}

const c1 = constant('hello');
const cu = constant(undefined);
const cf2 = constantFrom(true, false);
const cf5 = constantFrom('a', 'b', 'c', 'd', 'e');

bench('constant("hello")          ', () => c1.generate(rng, undefined));
bench('constant(undefined)        ', () => cu.generate(rng, undefined));
bench('constantFrom(true, false)  ', () => cf2.generate(rng, undefined));
bench('constantFrom("a".."e")     ', () => cf5.generate(rng, undefined));
bench('canShrinkWithoutContext    ', () => cf5.canShrinkWithoutContext('c'));
```

### Why this is patch-level

No API change. No behaviour change for a given seed. Only `_internals/ConstantArbitrary.ts` is touched. Existing tests pin both contracts (generation + `canShrinkWithoutContext` semantics).

## Checklist

---


packages/fast-check/src/arbitrary/_internals/ConstantArbitrary.ts
```diff
@@ -43,24 +43,52 @@ class FastConstantValuesLookup<T> {
...
     super();
+    // Pre-scan once at construction. For the common case (no cloneable inside),
+    // generate() can skip the per-call hasCloneMethod() probe entirely.
+    let hasAnyCloneable = false;
+    for (let idx = 0; idx !== values.length; ++idx) {
+      if (hasCloneMethod(values[idx])) {
+        hasAnyCloneable = true;
+        break;
+      }
+    }
```
dubzzz, line 58: Computation could be extracted into an internal function

packages/fast-check/src/arbitrary/_internals/ConstantArbitrary.ts
```diff
@@ -43,24 +43,52 @@ class FastConstantValuesLookup<T> {
 /** @internal */
 export class ConstantArbitrary<T> extends Arbitrary<T> {
   private fastValues: FastConstantValuesLookup<T> | undefined;
+  private readonly hasAnyCloneable: boolean;
 
   constructor(readonly values: T[]) {
```
dubzzz, line 48: Actually we could even envision two distinct versions of the class. One for non-cloneable (so not even an if) and another for possibly cloneable (rare case). The question of single versus multi values can also be raised the same way.

PR #7022, opened 2026-05-23
⚡ Speed up `IntegerArbitrary` generate and shrink

dubzzz:
## Description

`IntegerArbitrary` is the lowest-level numeric arbitrary in fast-check. `fc.nat`, `fc.boolean` (`integer({min:0,max:1}).map(...)`), the length of every `fc.array(…)` / `fc.string(…)`, the index draws inside `fc.constantFrom` / `fc.oneof` / `fc.tuple`, and many user-built composites all go through `IntegerArbitrary.generate(…)`. Anything we shave here is paid back many times over by everything downstream.

### What changes

- **Fast path in `generate` when `biasFactor === undefined`.** Before this PR, every call funnelled through `computeGenerateRange(...)`, which allocated a fresh `{ min, max }` object on the hot path. The fast branch now calls `mrng.nextInt(this.min, this.max)` directly; the biased branch is inlined back into `generate` and `computeGenerateRange` is removed altogether.
- **Cache `defaultTargetValue` once at construction.** `min <= 0 && max >= 0 ? 0 : min < 0 ? max : min` depends only on `min`/`max`; it's now a private readonly field. `shrink(value, undefined)` reads the field directly instead of going through `defaultTarget()`.
- **Cache `biasNumericRange(min, max, integerLogLike)` lazily on first biased call.** That helper allocates an array of `{ min, max }` objects and runs `Math.log` to pick break points. Its output depends only on `min`/`max`, so we compute it once and reuse it for every biased generate.
- **Replace `function*` generators in `shrinkInteger` with a hand-rolled `ShrinkIntegerIterator` class.** The previous implementation used two `function*` helpers, each of which allocates a generator frame + a closure capturing `current`/`target` per call to `.next()`. The class-based iterator allocates one object total and produces the same shrink sequence step-for-step (verified by the integer + helper tests). For shrink-heavy property-based tests (or property failures that exercise the full shrink tree), this is the biggest single win in the PR — ~4–15× depending on how far from the target the failing value sits.

### Background reading

- V8 element kinds & why `{ min, max }` allocation in a hot path costs measurable time: <https://v8.dev/blog/elements-kinds>.
- "Faster JS calls" / hidden-class & inline cache discipline: <https://v8.dev/blog/fast-properties>.
- Generator function cost (per-call frame + closure allocation): <https://v8.dev/blog/preparser> mentions parse-time savings, but the runtime overhead is well-documented in <https://www.zhenghao.io/posts/javascript-generators> and observed directly in our dexnode profile (`[deoptimizing]` / generator-frame allocations dominated the shrink path).
- Marvin Hagemeister, _Speeding up the JS ecosystem_: the recurring lesson that **caching values that depend only on construction-time constants** ranks above almost every other micro-fix — <https://marvinh.dev/blog/speeding-up-javascript-ecosystem-part-2/>.

### Observable behaviour

No public API change. No behaviour change for any seed:

- `generate` produces the same values (the only behavioural-adjacent change is reordering the `mrng.nextInt(1, biasFactor)` roll inside the biased branch, which is preserved bit-for-bit).
- `shrink` returns the same sequence of values in the same order (the `ShrinkIntegerIterator` implements the same `realGap`-halving algorithm as the original `function*`).

Tests pass unchanged:
- `test/unit/arbitrary/integer.spec.ts`
- `test/unit/arbitrary/_internals/IntegerArbitrary.spec.ts`
- `test/unit/arbitrary/_internals/helpers/ShrinkInteger.spec.ts`
- `test/unit/arbitrary/nat.spec.ts`

Total: 46/46 (plus the broader `test/unit/arbitrary/` sweep — 1690/1690 — still passes since this is a foundational change).

### Numbers

Two views: generate (single-step latency) and shrink (the iterator wins are largest here).

**Generate** (50M iters per config, TurboFan-warmed medians):

| Suite | Baseline | After | Δ |
| --- | ---: | ---: | ---: |
| `integer()` default i32 range | 26.7 Mop/s | 50.2 | **+88%** |
| `integer({ min: -100, max: 100 })` | 27.4 | 45.6 | **+66%** |
| `integer({ min: 0, max: 255 })` byte | 27.9 | 51.6 | **+85%** |
| `integer({ min: -1<<30, max: 1<<30 })` wide | 19.9 | 19.5 | ≈ flat (`uniformInt` dominates) |
| `nat()` default | 25.7 | 50.3 | **+96%** |

**Shrink** (paired compare, very stable):

| Suite | Baseline | After | Δ |
| --- | ---: | ---: | ---: |
| `shrink` default mid value | 393 kop/s | 6.0 Mop/s | **~15×** |
| `shrink` small positive | 507 kop/s | 4.1 Mop/s | **~8×** |
| `shrink` large negative | 395 kop/s | 1.7 Mop/s | **~4.3×** |
| `shrink` near `min` | 474 kop/s | 2.2 Mop/s | **~4.7×** |

The big shrink numbers are the generator-replacement effect — every step of the original shrink tree paid one generator-frame allocation + one closure allocation; the iterator pays one allocation total for the whole stream.

Re-runnable benchmark:

```js
// bench-integer.mjs
import { Random, integer, nat } from 'fast-check';

  let n = 0;
    n += 256;
  }
}

const i32   = integer();
const small = integer({ min: 0, max: 100 });
const byte  = integer({ min: 0, max: 255 });
const aNat  = nat();

bench('integer() default  ', () => i32.generate(rng, undefined));
bench('integer(0..100)    ', () => small.generate(rng, undefined));
bench('integer(0..255)    ', () => byte.generate(rng, undefined));
bench('nat()              ', () => aNat.generate(rng, undefined));

// shrink: drain the first 16 entries
const arb = integer();
bench('shrink mid (16 steps)', () => {
  let k = 0;
  for (const _ of arb.shrink(123456, undefined)) if (++k === 16) break;
});
```

### Why this is patch-level

No API change. No behaviour change for a given seed (verified by the integer/nat/shrink suites). Pure micro-optimisation of one of the busiest arbitraries in the library. Two files touched: `_internals/IntegerArbitrary.ts` and `_internals/helpers/ShrinkInteger.ts`.

## Checklist

---


packages/fast-check/src/arbitrary/_internals/IntegerArbitrary.ts
```diff
@@ -11,15 +11,38 @@ const safeObjectIs = Object.is;
...
+    this.defaultTargetValue = min <= 0 && max >= 0 ? 0 : min < 0 ? max : min;
   }
 
   generate(mrng: Random, biasFactor: number | undefined): Value<number> {
-    const range = this.computeGenerateRange(mrng, biasFactor);
+    if (biasFactor === undefined || mrng.nextInt(1, biasFactor) !== 1) {
+      return new Value(mrng.nextInt(this.min, this.max), undefined);
+    }
+    // Biased branch: pick one of the precomputed sub-ranges.
+    let ranges = this.cachedBiasedRanges;
```
dubzzz, line 33: The main performance gain is probably linked to the caching of biased ranges

dubzzz: Except inlining the code the rest of the code is unchanged from a generate point of view

dubzzz:
Looks promising, worth refining the PR and splitting it into sub-optims:

- +15% for generate
- x20 for shrink

PR #7025, opened 2026-05-23
⚡ Speed up `ChainArbitrary` generate and shrink

dubzzz:
## Description

`Arbitrary.prototype.chain(chainer)` (backed by `ChainArbitrary`) is the monadic composition primitive in fast-check — used both directly in user code (e.g. `fc.nat().chain(c => fc.array(fc.nat(c)))`) and internally for dependent arbitraries. The class has three hot frames worth attention: `generate`, `valueChainer`, and `shrink`.

### What changes

- **Inline `valueChainer` into `generate(...)`.** `generate` was effectively a thin wrapper that called `arb.generate(...)` then handed off to `valueChainer`. Inlining removes one method-call frame from the hottest path. `valueChainer` is preserved unchanged for the shrink re-entry (which still needs the "given a fresh inner value, produce a wrapped `Value<U>` with the right context" path).
- **Stable hidden class for `ChainArbitraryContext`.** `shrink(...)` used to build the context via `{ ...context, chainedContext, stoppedForOriginal: true }`, while `valueChainer` built it as an explicit literal. The two literals produced different field orderings, costing inline-cache slots on subsequent context-field reads. `shrink(...)` now uses an explicit literal whose key order matches `valueChainer`, so both call sites produce contexts with one shared V8 hidden class.
- **Skip `Stream.nil().join(chainedShrunk)` on the recursive branch.** When `stoppedForOriginal === true`, the joined-with-`Stream.nil()` form is observationally identical to just `chainedShrunk`. Returning `chainedShrunk` directly saves one Stream + one generator allocation per shrink invocation on this branch.

### Background reading

- V8 hidden classes & inline caches — why a literal's key order matters for IC stability: <https://mathiasbynens.be/notes/shapes-ics>.
- Marvin Hagemeister, _Speeding up the JS ecosystem_: same pattern (initialise all fields in a stable order, never `delete`) shows up across many "why is V8 slow here" investigations — <https://marvinh.dev/blog/speeding-up-javascript-ecosystem-part-2/>.
- Generator-frame allocation cost (`Stream.nil` and `Stream.join` are generator-backed): <https://www.zhenghao.io/posts/javascript-generators>; this is the same cost class that motivated the `IntegerArbitrary` shrink iterator in #7022 and the `FilterArbitrary` shrink iterator in #7024.

### Ideas considered but dropped

- **Lazy `mrng.clone()`.** Can't be deferred: `mrng` is mutated immediately by `arb.generate(...)`, so we need the snapshot at the original state.
- **Storing the bare pure-rand state instead of a wrapped `Random`.** Saves a tiny bit on generate but costs a wrap on every shrink iteration; net negative.
- **Reusing `dst` (the inner `Value`) with a swapped context.** `Value.context` is `readonly` and `Value.ts` is off-limits in this PR.
- **Class-based `ChainArbitraryContext`.** V8 already gives stable hidden classes to literals with explicit field ordering; adding a class adds prototype overhead with no benefit.

### Observable behaviour

No public API change. Same seed → same generated value, same shrink ordering. The chainer is invoked with the same `mrng` state as before; the pre-arb `mrng.clone()` happens at the same point.

Tests pass unchanged:
- `test/unit/check/arbitrary/definition/Arbitrary.utest.spec.ts`
- `test/unit/check/arbitrary/definition/Arbitrary.itest.spec.ts`
- `test/unit/check/arbitrary/definition/Value.utest.spec.ts`
- `test/unit/arbitrary/chainUntil.spec.ts`

Total: 42/42 (broader `test/unit/check/` sweep: 268/268).

### Numbers

Median of 7 runs × 3 s, paired against `main`:

| Suite | mean ratio | range |
| --- | ---: | ---: |
| `chain` generate, trivial constant chainer | **1.063** | 0.974–1.150 |
| `chain` generate, dependent array | **1.097** | 1.006–1.173 |
| `chain` generate, ignored chainer | **1.006** | 0.844–1.076 |
| `chain` generate, biased dependent array | **1.079** | 1.007–1.141 |
| `chain` shrink dependent array first 10 | **1.082** | 1.050–1.104 |

Roughly **+6 to +10%** on generate and **~+8%** on shrink. The "ignored chainer" case (where the chainer returns `fc.integer()` regardless of the input) is dominated by inner `fc.integer()` allocation done inside the user chainer, so the ChainArbitrary changes barely move it. Overall mean across all non-chain benchmarks: -2.0% to +1.6% across multiple compare runs — pure run-to-run noise. No deopts visible in the dexnode profile.

Re-runnable benchmark:

```js
// bench-chain.mjs
import { Random, integer, constant, array, nat } from 'fast-check';

  let n = 0;
    n += 256;
  }
}

const cTrivial = integer({ min: 0, max: 10 }).chain((n) => constant(n));
const cDep     = nat({ max: 10 }).chain((n) => array(integer(), { minLength: n, maxLength: n }));
const cIgnored = integer({ min: 0, max: 100 }).chain((_) => integer());

bench('chain trivial constant   ', () => cTrivial.generate(rng, undefined));
bench('chain dependent array    ', () => cDep.generate(rng, undefined));
bench('chain ignored chainer    ', () => cIgnored.generate(rng, undefined));
bench('chain shrink first 10    ', () => {
  let k = 0;
  for (const _ of cDep.shrink([1, 2, 3], undefined)) if (++k === 10) break;
});
```

### Why this is patch-level

No API change. No behaviour change for a given seed (verified by the Arbitrary + chainUntil suites). One file touched (`Arbitrary.ts`, only the `ChainArbitrary` class section).

## Checklist

---


dubzzz:
Not sure there would be a huge gain. The main benefit of this change is that the code in shrink might be simpler to read. The optimization consisting into in-lining the code looks not-ideal in terms of maintenance I'd prefer keeping the reusable method. That said if the performance gains are confirmed then there is a matter to discuss.

dubzzz:
The generate gain is rather great. The shrink one if more limited:

```txt

   fast-check  current-0 - test/bench/arbitraries.bench.ts > integer().chain(.) > generate
    1.01x faster than current-2
    1.01x faster than current-1
    1.86x faster than main-1
    1.86x faster than main-2
    1.87x faster than main-0

   fast-check  main-0 - test/bench/arbitraries.bench.ts > integer().chain(.) > shrink
    1.05x faster than main-1
    1.08x faster than main-2
    1.12x faster than current-2
    1.13x faster than current-0
    1.14x faster than current-1
```

Worth pushing that PR further.

PR #7045, opened 2026-06-04
⚡️ Faster `fc.entityGraph` on `generate`

dubzzz:
## Description

This is a **performance-only** change to `fc.entityGraph`. Generated values (graph topology and entity data) are unchanged — only the speed of `generate` improves. No public API change, no behavior change for end users. `entityGraph` is by far the most expensive arbitrary in the suite, so this is a welcome reduction in per-`generate` cost.

Measured improvement (interleaved separate-process A/B, median of medians):

- `entityGraph({ node: { id: integer() } }, { node: { linkTo: { arity: 'many', type: 'node' } } })`: **~−8% ns/op** (≈75.3 → ≈69.1 µs/op).
- The docstring `employee`/`team` schema (`0-1` manager, `1` team): **~−13% ns/op** (≈47.5 → ≈41.3 µs/op).

### Why

`buildEntityStepArbitrary` called `draftNextProductionState(lastState)` once **purely to read** the current entity (`getCurrentEntity()`) and the per-type existing counts (`getExistingEntityCount(targetType)`), and then called it a second time inside `.map` to perform the actual mutation. That first call is wasted work: it allocates a full clone of the produced-links state (`Object.assign(Object.create(null), …)`) plus several closures, all immediately discarded. `chainUntil` invokes this once per produced entity (~100× per construction), so it sits squarely on the hot path.

The two values it read are trivially derivable from `lastState` without any allocation:
- current entity = `lastState.toBeProducedEntities[lastState.nextIndex]`;
- existing count = `lastState.producedLinks[targetType].length` — the freshly-drafted state has not enqueued anything yet, so the lengths are identical (verified, see below).

So the preparation phase now reads those directly, and the mutable draft is built lazily only inside the `.map` where the real mutation happens. While there, this also removes a dead `countsInTargetType` object that was populated but never read, and the two now-unused getter closures on the draft object (fewer closures allocated per remaining step).

### Scope, correctness and impact

- **Impact: patch.** A changeset (`fast-check: patch`, `⚡️ Faster fc.entityGraph on generate`) is included.
- **Single concern:** one file (`OnTheFlyLinksForEntityGraphArbitrary.ts`) plus the changeset.
- **Tests:** the `entityGraph` integration spec and the internal `_internals` specs (`OnTheFlyLinks`, `UnlinkedToLinkedEntities`, `InitialPool`, `UnlinkedEntities`) all pass, and the full suite is green. Output was additionally checked to be structurally identical to `main` (800 generations, 0 fingerprint mismatches), so the eliminated call provably only read values equal to the direct reads — no new test is needed because the distribution is unchanged. The removed getters had no other callers (grep-confirmed, including tests).

## Checklist

---
_Generated by [Claude Code](https://claude.ai/code/session_01SRx4yxfDXbBWd5JDy7nrtz)_

### DIFF
--- a/packages/fast-check/src/arbitrary/_internals/OnTheFlyLinksForEntityGraphArbitrary.ts
+++ b/packages/fast-check/src/arbitrary/_internals/OnTheFlyLinksForEntityGraphArbitrary.ts
@@ -189,10 +189,6 @@ function draftNextProductionState<TEntityFields, TEntityRelations extends Entity
 
   const toBeProduced = toBeProducedEntities[nextIndex];
   return {
-    // The entity being produced in this step. Exposed as a value (not a setter) since it cannot be mutated from here.
-    getCurrentEntity: (): Readonly<ToBeProducedEntity<TEntityFields>> => toBeProduced,
-    // Number of entities of the given type already produced so far.
-    getExistingEntityCount: (targetType: keyof TEntityFields) => newProducedLinks[targetType].length,
     // Edit functions
     setOutboundLink: (
       name: keyof TEntityRelations[keyof TEntityFields],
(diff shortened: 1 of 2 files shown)


packages/fast-check/src/arbitrary/_internals/OnTheFlyLinksForEntityGraphArbitrary.ts
```diff
@@ -257,12 +253,15 @@ function buildEntityStepArbitrary<TEntityFields, TEntityRelations extends Entity
   inversedRelations: ReturnType<typeof buildInversedRelationsMapping<TEntityFields>>,
   lastState: ProductionState<TEntityFields, TEntityRelations>,
 ): Arbitrary<ProductionState<TEntityFields, TEntityRelations>> {
-  const state = draftNextProductionState(lastState);
-  const currentEntity = state.getCurrentEntity();
+  // Cheap read-only access to the current entity and per-type counts.
+  // We intentionally avoid `draftNextProductionState` here: building it allocates a full clone of the
+  // produced-links state plus several closures, all of which would be thrown away as this preparation
+  // phase only needs read-only data. The mutable draft is created lazily inside the `.map` below.
```
dubzzz:
```suggestion
```

PR #7046, opened 2026-06-04
⚡️ Faster fc.record on generate

dubzzz:
## Description

This speeds up the `generate` path of `fc.record(...)`. Records are extremely common building blocks (directly, and inside `letrec`/`memo`/`entityGraph`/`anything`), so making their construction cheaper benefits a wide range of property-based tests. There is **no observable behavior change**: generated records have the exact same keys, values, property attributes (configurable/enumerable/writable) and prototype as before — this is a pure performance change with patch impact.

### What changed

The mapper that turns generated values into the final record object (`valuesAndSeparateKeysToObjectMapper`) used `Object.defineProperty` for every present key:

```js
safeObjectDefineProperty(obj, keys[idx], { value, configurable: true, enumerable: true, writable: true });
```

`Object.defineProperty` with all three flags set to `true` produces **exactly the same own data property** as a plain assignment `obj[key] = value`, but in V8 it is markedly slower and tends to push the freshly created object onto a slower path. The hot loop now assigns directly, which keeps the object on the fast path and removes a per-property builtin call.

The single exception is the string key `"__proto__"`: a plain assignment would trigger the prototype setter on a regular object instead of defining an own property, so that one key keeps using `defineProperty`. Symbol keys and every other string key (including `constructor`, `toString`, …) are plain data properties on `Object.prototype` and are safely created by assignment, so they take the fast path.

### Why this design

- It is the minimal, surgical change: one mapper, one `__proto__` guard, identical observable output.
- The `__proto__`-only guard is provably sufficient — `__proto__` is the only accessor on `Object.prototype`; every other "dangerous" key (`constructor`, `toString`, …) is a plain data property that assignment safely shadows. The existing dangerous-key unit tests cover exactly this and still pass.

### Measurements

Isolated mapper microbenchmark (lower is better):

| keys | `Object.defineProperty` | direct assignment |
| --- | --- | --- |
| 1 | ~356 ns/op | **~15 ns/op** |
| 3 | ~1049 ns/op | **~78 ns/op** |

End-to-end `record(...).generate(...)` improves by roughly **20–34%** on multi-key records (the RNG dominates single-key records, where the mapper work is negligible). No regression observed on single-key records.

### Tests

No new tests were required: the behavior is unchanged and the existing suites already pin it down. `test/unit/arbitrary/_internals/mappers/ValuesAndSeparateKeysToObject.spec.ts` (incl. the `"__proto__"`-as-key cases), `record.spec.ts`, and `PartialRecordArbitraryBuilder.spec.ts` all pass (47 tests), and `tsc` is clean. A changeset (`patch`) is included.

## Checklist

---
_Generated by [Claude Code](https://claude.ai/code/session_0172JCqbtKRTQA1r4TN9Tg6p)_

### DIFF
--- a/packages/fast-check/src/arbitrary/_internals/mappers/ValuesAndSeparateKeysToObject.ts
+++ b/packages/fast-check/src/arbitrary/_internals/mappers/ValuesAndSeparateKeysToObject.ts
@@ -16,15 +16,25 @@ export function buildValuesAndSeparateKeysToObjectMapper<T, TNoKey>(keys: Enumer
     definition: ObjectDefinition<T, TNoKey>,
   ): Partial<T> & Pick<T, EnumerableKeyOf<T>> {
     const obj: Partial<Record<EnumerableKeyOf<T>, T[keyof T]>> = definition[1] ? safeObjectCreate(null) : {};
+    const values = definition[0];
     for (let idx = 0; idx !== keys.length; ++idx) {
-      const valueWrapper = definition[0][idx];
+      const valueWrapper = values[idx];
       if (valueWrapper !== noKeyValue) {
-        safeObjectDefineProperty(obj, keys[idx], {
-          value: valueWrapper,
-          configurable: true,
-          enumerable: true,
-          writable: true,
-        });
+        const key = keys[idx];
+        // A plain assignment creates the same own configurable/enumerable/writable data property as
+        // `Object.defineProperty` with all flags set, but is significantly cheaper in V8. The only key
+        // needing the slower path is the string "__proto__": a plain assignment would trigger the
+        // prototype setter (on a regular object) instead of defining an own property.
+        if (key === '__proto__') {
+          safeObjectDefineProperty(obj, key, {
+            value: valueWrapper,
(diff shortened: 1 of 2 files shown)


packages/fast-check/src/arbitrary/_internals/mappers/ValuesAndSeparateKeysToObject.ts
```diff
@@ -16,15 +16,25 @@ export function buildValuesAndSeparateKeysToObjectMapper<T, TNoKey>(keys: Enumer
...
-          value: valueWrapper,
-          configurable: true,
-          enumerable: true,
-          writable: true,
-        });
+        const key = keys[idx];
+        // A plain assignment creates the same own configurable/enumerable/writable data property as
+        // `Object.defineProperty` with all flags set, but is significantly cheaper in V8. The only key
+        // needing the slower path is the string "__proto__": a plain assignment would trigger the
+        // prototype setter (on a regular object) instead of defining an own property.
```
dubzzz:
```suggestion
```

PR #7047, opened 2026-06-04
⚡️ Faster fc.dictionary on generate

dubzzz:
## Description

This speeds up the `generate` path of `fc.dictionary(...)`. There is **no observable behavior change**: generated dictionaries have the exact same keys, values, property attributes (configurable/enumerable/writable) and prototype (object vs null) as before — a pure performance change with patch impact.

### What changed

The mapper that turns generated `[key, value]` pairs into the final object (`keyValuePairsToObjectMapper`) used `Object.defineProperty` for every pair:

```js
safeObjectDefineProperty(obj, keyValue[0], { enumerable: true, configurable: true, writable: true, value: keyValue[1] });
```

`Object.defineProperty` with all three flags set to `true` produces **exactly the same own data property** as a plain assignment `obj[key] = value`, but in V8 it is markedly slower and tends to push the freshly created object onto a slower path. The hot loop now assigns directly.

As with the matching `fc.record` change, the single exception is the string key `"__proto__"`: a plain assignment would trigger the prototype setter on a regular object instead of defining an own property, so that one key keeps using `defineProperty`. Symbol/number keys and every other string key (including `constructor`, `toString`, …) are plain data properties and are safely created by assignment.

### Why this design

- Minimal, surgical change isolated to the one mapper, with an identical observable result.
- The `__proto__`-only guard is provably sufficient — `__proto__` is the only accessor on `Object.prototype`; every other "dangerous" key is a plain data property that assignment safely shadows. The existing dangerous-key unit tests (`['__proto__', 'constructor', 'toString']`, both object- and null-prototype) cover exactly this and still pass.

### Measurements

Isolated mapper microbenchmark (lower is better):

| keys | `Object.defineProperty` | direct assignment |
| --- | --- | --- |
| 3 | ~799 ns/op | **~59 ns/op** |
| 10 | ~2518 ns/op | **~188 ns/op** |

`fc.dictionary().generate()` is dominated by string-key generation, so the end-to-end gain is most visible on larger dictionaries; the mapper itself is ~13× faster.

### Tests

No new tests required — behavior is unchanged and the existing suites pin it down. `test/unit/arbitrary/_internals/mappers/KeyValuePairsToObject.spec.ts` (incl. the dangerous-key and null-prototype cases) and `dictionary.spec.ts` all pass (24 tests); `tsc` is clean. A changeset (`patch`) is included.

This is a sibling of the `fc.record` mapper change (kept as a separate, single-concern PR).

## Checklist

---
_Generated by [Claude Code](https://claude.ai/code/session_0172JCqbtKRTQA1r4TN9Tg6p)_

### DIFF
--- a/packages/fast-check/src/arbitrary/_internals/mappers/KeyValuePairsToObject.ts
+++ b/packages/fast-check/src/arbitrary/_internals/mappers/KeyValuePairsToObject.ts
@@ -15,13 +15,23 @@ export function keyValuePairsToObjectMapper<K extends PropertyKey, V>(
   definition: ObjectDefinition<K, V>,
 ): Record<K, V> {
   const obj: Record<K, V> = definition[1] ? safeObjectCreate(null) : {};
-  for (const keyValue of definition[0]) {
-    safeObjectDefineProperty(obj, keyValue[0], {
-      enumerable: true,
-      configurable: true,
-      writable: true,
-      value: keyValue[1],
-    });
+  const keyValues = definition[0];
+  for (let idx = 0; idx !== keyValues.length; ++idx) {
+    const key = keyValues[idx][0];
+    // A plain assignment creates the same own configurable/enumerable/writable data property as
+    // `Object.defineProperty` with all flags set, but is significantly cheaper in V8. The only key
+    // needing the slower path is the string "__proto__": a plain assignment would trigger the
+    // prototype setter (on a regular object) instead of defining an own property.
+    if (key === '__proto__') {
+      safeObjectDefineProperty(obj, key, {
+        enumerable: true,
+        configurable: true,
+        writable: true,
+        value: keyValues[idx][1],
+      });
+    } else {
+      obj[key] = keyValues[idx][1];
+    }
   }
   return obj;
 }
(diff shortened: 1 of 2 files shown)


packages/fast-check/src/arbitrary/_internals/mappers/KeyValuePairsToObject.ts
```diff
@@ -15,13 +15,23 @@ export function keyValuePairsToObjectMapper<K extends PropertyKey, V>(
...
-      writable: true,
-      value: keyValue[1],
-    });
+  const keyValues = definition[0];
+  for (let idx = 0; idx !== keyValues.length; ++idx) {
+    const key = keyValues[idx][0];
+    // A plain assignment creates the same own configurable/enumerable/writable data property as
+    // `Object.defineProperty` with all flags set, but is significantly cheaper in V8. The only key
+    // needing the slower path is the string "__proto__": a plain assignment would trigger the
+    // prototype setter (on a regular object) instead of defining an own property.
```
dubzzz:
```suggestion
```

PR #7050, opened 2026-06-04
⚡️ Faster `fc.stringMatching` for `\W` `\D` `\S` `.`

dubzzz:
## Description

This is a **performance-only** change to `fc.stringMatching`. Generated values are byte-for-byte unchanged — only the speed of `generate` improves for regexes using the negated meta-character classes `\W`, `\D`, `\S` and `.`. No public API change, no behavior change for end users.

Measured improvement (interleaved separate-process A/B, median of medians):

- `stringMatching(/\W{8}/)` and similar `\W`-heavy patterns: **~−5–8% ns/op** (measured −5.5% in a clean run, ~−8% in the discovering agent's runs; always beats the matched noise floor).
- Regexes that don't use these negated classes (e.g. the common `/^[a-zA-Z0-9]+$/`) are unaffected — the control sample moved −1.1%, inside the noise band.
- `\D`/`\S`/`.` benefit too in principle, but their forbidden-character arrays are tiny (10/6/2–4 entries) so the gain there stays within measurement noise.

### Why

The negated meta-character filters were implemented as a linear scan over a forbidden-character array, run **per generated character** on the rejection-sampling path:

```js
case '\\W': return defaultChar().filter((c) => safeIndexOf(wordChars, c) === -1);
```

`\W`'s `wordChars` array has 63 entries and `\W` accepts only ~34% of `grapheme-ascii` candidates, so each accepted character costs several full-length `O(n)` scans. This precomputes the membership sets once at module scope and tests with the poisoning-safe `safeHas` (`O(1)`):

```js
const wordCharsSet = new Set(wordChars); // + digit/space/terminator/newLineAndTerminator sets
case '\\W': return defaultChar().filter((c) => !safeHas(wordCharsSet, c));
```

Because the accept/reject decision for every candidate is identical, the RNG draw sequence — and therefore the generated (and shrunk) values — are exactly preserved.

### Scope, correctness and impact

- **Impact: patch.** A changeset is included.
- **Single concern:** one file (`stringMatching.ts`) plus the changeset. Uses the existing poisoning-safe `Set`/`safeHas` globals, consistent with the rest of the codebase.
- **Tests:** `stringMatching`, `SanitizeRegexAst`, `TokenizeRegex`, `ClampRegexAst` all pass (229 tests). Output was verified byte-identical to `main` across many regexes × seeds × bias settings, so the distribution is provably unchanged and no new test is needed.
- **Note for reviewers:** this is independent of #7044 (which touches the `Alternative` branch); the only textual overlap is the `globals.js` import line, a trivial rebase for whichever merges second.

## Checklist

---
_Generated by [Claude Code](https://claude.ai/code/session_01SRx4yxfDXbBWd5JDy7nrtz)_

### DIFF
--- a/packages/fast-check/src/arbitrary/stringMatching.ts
+++ b/packages/fast-check/src/arbitrary/stringMatching.ts
@@ -1,14 +1,5 @@
 import type { Arbitrary } from '../check/arbitrary/definition/Arbitrary.js';
-import {
-  safeCharCodeAt,
-  safeEvery,
-  safeJoin,
-  safeSubstring,
-  Error,
-  safeMap,
-  Set,
-  safeHas,
-} from '../utils/globals.js';
+import { safeCharCodeAt, safeEvery, safeJoin, safeSubstring, Error, safeMap, Set, safeHas } from '../utils/globals.js';
 import { stringify } from '../utils/stringify.js';
 import { clampRegexAst } from './_internals/helpers/ClampRegexAst.js';
 import type { SizeForArbitrary } from './_internals/helpers/MaxLengthFromMinLength.js';
--- a/packages/fast-check/src/arbitrary/stringMatching.ts
+++ b/packages/fast-check/src/arbitrary/stringMatching.ts
@@ -1,5 +1,14 @@
 import type { Arbitrary } from '../check/arbitrary/definition/Arbitrary.js';
-import { safeCharCodeAt, safeEvery, safeJoin, safeSubstring, Error, safeIndexOf, safeMap } from '../utils/globals.js';
+import {
+  safeCharCodeAt,
+  safeEvery,
+  safeJoin,
+  safeSubstring,
+  Error,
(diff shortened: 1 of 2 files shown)


packages/fast-check/src/arbitrary/stringMatching.ts
```diff
@@ -44,6 +53,15 @@ const newLineChars = [...'\r\n'];
 const terminatorChars = [...'\x1E\x15'];
 const newLineAndTerminatorChars = [...newLineChars, ...terminatorChars];
 
+// Precomputed membership sets for the negated meta-character filters (\W \D \S and .).
+// Using a Set gives O(1) membership instead of the O(n) linear scan performed by
+// safeIndexOf, without changing which characters are accepted nor the RNG draw sequence.
```
dubzzz:
```suggestion
// Precomputed membership sets for faster lookups
```

PR #7063, opened 2026-06-07
🐛 Equiprobable alternatives in `stringMatching`

dubzzz:
## Description

**TL;DR** — Fixes a distribution bug in `fc.stringMatching`: alternatives of a
regex alternation (`a|b|c|…`) were **not equiprobable** (later branches were far
more likely). As a side benefit, flattening the alternation also makes `generate`
**faster** on alternation-heavy patterns. Values still match the same regex and
same-seed determinism is preserved per pattern; the only observable change is the
distribution across alternatives. Impact level: **patch** (a changeset is
included).

### The bug — skewed alternatives

A regex alternation `a|b|c|d` is parsed as a *nested* binary disjunction tree —
`Disjunction(Disjunction(Disjunction(a, b), c), d)` — and was translated
recursively into `oneof(oneof(oneof(a, b), c), d)`. That nesting makes the
distribution **non-uniform**: in `a|b|c|d`, `d` is picked ~1/2 of the time while
`a` and `b` get only ~1/8 each. Users expect each alternative to be equally
likely.

### The fix

The disjunction tree is flattened into a **single `oneof` over all
alternatives**, so every alternative is now equiprobable. As a bonus, choosing an
alternative is a single selection regardless of arity instead of walking an N−1
deep stack of `FrequencyArbitrary` (`oneof`) wrappers — which also speeds up
`generate`.

### `generate` gains

A disjunction-dominated pattern shows the effect cleanly (`vitest bench`, median
of 5 runs):

| pattern | main | this PR | speedup |
| --- | ---: | ---: | ---: |
| `/^(?:a\|b\|c\|d\|e\|f\|g\|h\|i\|j\|k\|l\|m\|n\|o\|p)+$/` | 403,956 hz | 520,888 hz | **1.29×** |

### Effect on the composite benchmark regexes (vs current `main`)

Measured against up-to-date `main` (median of 7 `vitest bench` runs):

| benchmark regex | speedup |
| --- | ---: |
| email `/^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,}$/` | ≈1.0× (no alternation) |
| URL (the #7059 bench regex) | ≈1.0× (a single shallow `(https?\|ftp)`) |

These composites contain little or no nested alternation, so the flattening shows
no end-to-end change there; the gain appears on alternation-heavy patterns (table
above), while the equiprobable-distribution fix applies regardless of arity.

### Scope & tests

Single concern: only the `Disjunction` case is changed. Correctness is covered by
the existing integration tests (`should only produce correct values`, `should
produce the same values given the same seed`), which keep passing; documentation
samples that depend on the generated stream were refreshed to reflect the new
(uniform) distribution.

## Checklist

### DIFF
--- a/packages/fast-check/src/arbitrary/stringMatching.ts
+++ b/packages/fast-check/src/arbitrary/stringMatching.ts
@@ -178,7 +178,7 @@ function toMatchingArbitrary(
     case 'Disjunction': {
       const stack = [astNode.left, astNode.right];
       const branches: Arbitrary<string>[] = [];
-      for (let i = 0 ; i !== stack.length ; ++i) {
+      for (let i = 0; i !== stack.length; ++i) {
         const node = stack[i];
         if (node === null) {
           safePush(branches, constant(''));
--- a/packages/fast-check/src/arbitrary/stringMatching.ts
+++ b/packages/fast-check/src/arbitrary/stringMatching.ts
@@ -178,8 +178,8 @@ function toMatchingArbitrary(
     case 'Disjunction': {
       const stack = [astNode.left, astNode.right];
       const branches: Arbitrary<string>[] = [];
-      while (branches.length !== stack.length) {
-        const node = stack[branches.length];
+      for (let i = 0 ; i !== stack.length ; ++i) {
+        const node = stack[i];
         if (node === null) {
           safePush(branches, constant(''));
         } else if (node.type === 'Disjunction') {
--- a/packages/fast-check/src/arbitrary/stringMatching.ts
+++ b/packages/fast-check/src/arbitrary/stringMatching.ts
@@ -188,7 +188,7 @@ function toMatchingArbitrary(
         } else {
           safePush(branches, toMatchingArbitrary(node, constraints, flags));
         }
-      };
+      }
       return oneof(...branches);
     }
     case 'Assertion': {
--- a/packages/fast-check/src/arbitrary/stringMatching.ts
+++ b/packages/fast-check/src/arbitrary/stringMatching.ts
@@ -176,23 +176,19 @@ function toMatchingArbitrary(
       return toMatchingArbitrary(astNode.expression, constraints, flags);
     }
     case 'Disjunction': {
-      // A regex alternation `a|b|c|d` parses as a nested binary disjunction tree:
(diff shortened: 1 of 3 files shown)


packages/fast-check/src/arbitrary/stringMatching.ts
```diff
@@ -166,9 +176,24 @@ function toMatchingArbitrary(
...
+          safePush(branches, constant(''));
+        } else if (node.type === 'Disjunction') {
+          collectBranches(node.left);
+          collectBranches(node.right);
+        } else {
+          safePush(branches, toMatchingArbitrary(node, constraints, flags));
+        }
+      };
+      collectBranches(astNode.left);
+      collectBranches(astNode.right);
```
dubzzz:
```suggestion
      const stack = [astNode.left, astNode.right];
      const branches: Arbitrary<string>[] = [];
      while (branches.length !== stack.length) {
        const node = stack[branches.length];
        if (node === null) {
          safePush(branches, constant(''));
        } else if (node.type === 'Disjunction') {
          safePush(stack, node.left);
          safePush(stack, node.right);
        } else {
          safePush(branches, toMatchingArbitrary(node, constraints, flags));
        }
      };
```

packages/fast-check/src/arbitrary/stringMatching.ts
```diff
@@ -166,9 +176,20 @@ function toMatchingArbitrary(
...
+        if (node === null) {
+          safePush(branches, constant(''));
+        } else if (node.type === 'Disjunction') {
+          safePush(stack, node.left);
+          safePush(stack, node.right);
+        } else {
+          safePush(branches, toMatchingArbitrary(node, constraints, flags));
+        }
+      }
+      return oneof(...branches);
```
dubzzz, line 217:
```suggestion
      const stack = [astNode.left, astNode.right];
      const branches: Arbitrary<string>[] = [];
      for (let i = 0 ; i !== stack.length ; ++i) {
        const node = stack[i];
        if (node === null) {
          safePush(branches, constant(''));
        } else if (node.type === 'Disjunction') {
          safePush(stack, node.left);
          safePush(stack, node.right);
        } else {
          safePush(branches, toMatchingArbitrary(node, constraints, flags));
        }
      }
      return oneof(...branches);
```

PR #7136, opened 2026-07-11
🐛 Make float/double/typed-array NaN generation non-canonical

### LINKED ISSUE #6532: `float()`, `double()`, `float32Array()`, and `float64Array()` never produce non-canonical NaNs
## 🐛 Bug Report

ECMAScript declares that all NaN values are indistinguishable from each other, but that the underlying bit pattern, which may be non-canonical, may be observable by writing to an `ArrayBuffer` depending on the engine.

https://tc39.es/ecma262/multipage/ecmascript-data-types-and-values.html#sec-ecmascript-language-types-number-type

I first noticed that `float32Array()` and `float64Array()` actually never seem to be able to produce non-canonical NaN bit patterns because they are derived from lists of floats/doubles and those are also always generated as canonical. 

I think it would be better if at least `float32Array()` and `float64Array()` could produce any sequence of underlying bytes (e.g. maybe by being derived from an arbitrary `ArrayBuffer`). This would help with testing things that are supposed to roundtrip data perfectly.

I'm less opinionated about `float()` and `double()` producing non-canonical NaNs as well, because those differences might be really hard to observe in practice (only when writing to an `ArrayBuffer` before/without doing operations that happen to canonicalize the NaN), but it might still be worth it for catching those edge cases.

## To Reproduce

Steps to reproduce:

```ts
const canonicalNaNBytes = new Uint8Array(new Float32Array([Number.NaN]).buffer)

fc.sample(
  fc
    .float32Array()
    // Filter for Float32Arrays with non-canonical NaNs.
    .filter(array => {
      const nanIndex = array.findIndex(value => Number.isNaN(value))
      if (nanIndex === -1) {
        return false
      }

      const nanBytes = new Uint8Array(
        array.slice(nanIndex, nanIndex + 1).buffer,
      )
      for (let i = 0; i < nanBytes.length; i++) {
        if (nanBytes[i] !== canonicalNaNBytes[i]) {
          // Found a non-canonical NaN!
          return true
        }
      }

      return false
    }),
  10,
)
```

## Expected behavior

The sampling should terminate by finding something, but it never finds a matching value.

## Your environment

| Packages / Softwares | Version(s) |
| -------------------- | ---------- |
| fast-check           |   4.5.3         |
| node                 |  v24.8.0          |
| TypeScript\*         |    5.9.3        |

\*Only for TypeScript's users

vishnujayvel:
## Description

Fixes #6532

`fc.float()`, `fc.double()`, `fc.float32Array()` and `fc.float64Array()` always hardcoded `Number.NaN` as their "extra" NaN value. Since all NaN values are indistinguishable from each other at the JS language level (`Number.isNaN`, `===`, `Object.is`...), these arbitraries could never surface the non-canonical NaN bit patterns (signaling NaNs, alternate payloads, negative NaNs...) that only become observable once a value is copied into a `Float32Array`/`Float64Array` — exactly the scenario @cschanhniem described in the issue.

Root cause and fix approach per @cschanhniem's analysis on the issue: reuse the single index that `anyFloat`/`anyDouble` already dedicate to NaN, but map it to a fixed non-canonical bit pattern instead of `Number.NaN`, rather than drawing extra randomness for a NaN "variant" index. The latter was tried first and reliably destabilized `float32Array`/`float64Array`'s "shrink without initial context" contract (`ArrayArbitrary` drops per-element context, and `IntegerArbitrary`'s context-free shrink target depends on the exact index range), so it was discarded in favor of this zero-extra-entropy design.

Two already-quiet, non-canonical 32/64-bit patterns are used (one for ranges extending past +0, one for ranges below it), avoiding signaling patterns since some engines quiet them when written into a TypedArray.

Impact: patch — no public API changes, only the underlying bit pattern of an already-produced NaN sentinel value changes (which is unobservable except via raw byte inspection, as described above).

## Checklist

## Test plan

Ran locally on a fresh fork/clone against current `main`:

```
pnpm install
pnpm exec vitest run --project fast-check \
  packages/fast-check/test/unit/arbitrary/float.spec.ts \
  packages/fast-check/test/unit/arbitrary/double.spec.ts \
  packages/fast-check/test/unit/arbitrary/float32Array.spec.ts \
  packages/fast-check/test/unit/arbitrary/float64Array.spec.ts \
  packages/fast-check/test/unit/arbitrary/_internals/helpers/FloatHelpers.spec.ts \
  packages/fast-check/test/unit/arbitrary/_internals/helpers/FloatOnlyHelpers.spec.ts \
  packages/fast-check/test/unit/arbitrary/_internals/helpers/DoubleOnlyHelpers.spec.ts \
  packages/fast-check/test/unit/arbitrary/_internals/helpers/DoubleHelpers.spec.ts \
  packages/fast-check/test/e2e/arbitraries/FloatArbitrary.spec.ts \
  packages/fast-check/test/e2e/arbitraries/DoubleArbitrary.spec.ts \
  packages/fast-check/test/e2e/NoRegression.spec.ts
# → Test Files 11 passed (11), Tests 229 passed (229)

cd packages/fast-check && pnpm run typecheck   # → passes, no errors
pnpm run lint:check                            # → passes (pre-existing warnings only, none touch changed files)
pnpm run format:check                          # → passes, no diffs
```

I also confirmed the 4 new regression tests actually catch the bug: reverting only `src/arbitrary/double.ts`, `src/arbitrary/float.ts`, and their `_internals/helpers/{Double,Float}Helpers.ts` back to the pre-fix `Number.NaN` hardcoding (while keeping the new tests) makes exactly those 4 tests fail, confirming they would have failed without this PR:
- `float32Array.spec.ts > should be able to produce non-canonical NaN bit patterns (see #6532)`
- `float64Array.spec.ts > should be able to produce non-canonical NaN bit patterns (see #6532)`
- `float.spec.ts > should properly convert the extra value to NaN`
- `double.spec.ts > should properly convert the extra value to NaN`

The `NoRegression` e2e snapshot suite passes unchanged, confirming the index topology (and thus shrink behavior) is unaffected — only the sentinel's underlying bit pattern changed.

---

### AI-assistance disclosure

In keeping with this project's contributing guidance: this contribution was prepared with AI assistance and was **reviewed, run, and approved by me** before submission — I understand the change and take responsibility for it.

**What it does, in my words:** the float/double/typed-array generators only ever emitted the *canonical* NaN bit pattern, so property tests could never exercise non-canonical NaNs (#6532). This maps the existing single NaN sentinel index to a fixed non-canonical bit pattern with **zero additional entropy**, so the value-index topology — and therefore all shrinking behaviour and the `NoRegression` snapshots — stays identical; only the sentinel's underlying bits change.

**What I verified myself:** the full test-plan above (229/229 targeted + NoRegression, typecheck/lint/format clean), plus the revert-confirms-fail check on the 4 new tests. Root-cause credit to @cschanhniem.

Happy to adjust anything to fit your preferences.

### DIFF
--- a/.changeset/noncanonical-nan-generation.md
+++ b/.changeset/noncanonical-nan-generation.md
@@ -1,0 +1,5 @@
+---
+"fast-check": patch
+---
+
+🐛 Make float/double/typed-array NaN generation non-canonical
--- a/packages/fast-check/src/arbitrary/_internals/helpers/DoubleHelpers.ts
+++ b/packages/fast-check/src/arbitrary/_internals/helpers/DoubleHelpers.ts
@@ -20,6 +20,42 @@ function bitCastDoubleToUInt64(f: number): [number, number] {
   f64[0] = f;
   return [u32[1], u32[0]];
 }
+/** @internal */
+function bitCastUInt64ToDouble(hi: number, lo: number): number {
+  u32[1] = hi;
+  u32[0] = lo;
+  return f64[0];
+}
+
+/**
+ * 64-bit floating point NaN bit patterns that fast-check is able to produce, expressed as [hi, lo] pairs of
+ * 32-bit unsigned integers (hi holding the sign, exponent and top mantissa bits, lo the low mantissa bits).
+ *
+ * All the values below are indistinguishable from each other -and from any other NaN- as soon as they hit the
+ * language level: `Number.isNaN`, `===`, `Object.is`... never let any difference show up. The only way to
+ * observe the difference is to look at their raw bits, for instance by copying the value into a `Float64Array`
+ * and reading it back as a `Uint32Array` (see #6532).
+ *
+ * NAN_64_BIT_PATTERNS[0] is the canonical one: it is the pattern used whenever JavaScript itself has to
+ * produce a NaN (for instance as the result of `0/0`, or as `Number.NaN`).
+ *
+ * @internal
+ */
+export const NAN_64_BIT_PATTERNS: readonly [number, number][] = [
+  [0x7ff80000, 0x00000000], // Canonical quiet NaN (positive)
+  [0x7ff00000, 0x00000001], // Signaling NaN (positive, smallest payload)
+  [0x7fffffff, 0xffffffff], // Quiet NaN with all mantissa bits set
+  [0xfff80000, 0x00000000], // Quiet NaN (negative)
+  [0xfff00000, 0x00000001], // Signaling NaN (negative, smallest payload)
+];
+
+/**
+ * 64-bit floating point NaN values that fast-check is able to produce, derived from {@link NAN_64_BIT_PATTERNS}.
+ * NAN_64_VALUES[0] is the canonical one.
+ *
+ * @internal
+ */
+export const NAN_64_VALUES: readonly number[] = NAN_64_BIT_PATTERNS.map(([hi, lo]) => bitCastUInt64ToDouble(hi, lo));
 
 /**
  * Decompose a 64-bit floating point number into its interpreted parts:
--- a/packages/fast-check/src/arbitrary/_internals/helpers/FloatHelpers.ts
+++ b/packages/fast-check/src/arbitrary/_internals/helpers/FloatHelpers.ts
@@ -21,6 +21,40 @@ function bitCastFloatToUInt32(f: number): number {
   f32[0] = f;
   return u32[0];
 }
+/** @internal */
+function bitCastUInt32ToFloat(n: number): number {
+  u32[0] = n;
+  return f32[0];
+}
+
+/**
+ * 32-bit floating point NaN bit patterns that fast-check is able to produce.
+ *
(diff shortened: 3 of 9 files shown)


dubzzz:
Actually so far we don't want to address that issue. Thanks for the suggestion but we might close it until we make it something we really want

vishnujayvel:
Thanks for taking a look, and totally understand — no worries if it's not something you want to take on right now. Happy to close it, or leave it open as a reference for later, whatever's easiest for you. Appreciate the time either way. 🙏

dubzzz:
I'm gonna look to it right before opening the 5.x version. It might be a good fit for 5.x as it may break some users with the new behaviour. That said the new behaviour is the one most of people should expect so a minor would be fine too.

vishnujayvel:
Thanks for following up on it. Please keep me posted if you want me to rebase and publish it for new release in the future!

vishnujayvel:
Circling back on this one — happy to rebase whenever you start planning the 5.x cut, since you mentioned this might fit better there given the behavior change. Let me know when that's on the horizon and I'll get it current.

PR #7140, opened 2026-07-17
💥 Drop global capture of functions

dubzzz:
## Description

In the past, fast-check decided to capture all the globals and methods it relies on. The motivation behind that logic has been to be safe in case someone during the execution of a test breaks them.

Since that point in time, things moved, we moved. We now consider that in the context of a test there is no real reason to break the globals except doing it on purpose or manipulating risky code potentially subject to prototype pollution. In such cases we prefer our users to use our library called `@fast-check/poisoning`. This package is tailored to protect you and your tests against such prototype pollutions.

Another motivation to drop that is taht it increases the congnitive complexity of our code and thus its maintainance. It also adds a barrier for external contributors. Last point is that it may bring a performance penalty.

Next part of this cleaning: drop capture on methods.

## Checklist


dubzzz:
## CI failure review

**TL;DR:** A find‑and‑replace slip in `double.ts` broke the double `unmapper`: `safeNumberIsNaN(value)` was replaced with `Number.isInteger(value)` instead of `Number.isNaN(value)`. This corrupts the value→index mapping used to rebuild the shrink tree, so the *"should be able to shrink to the same values without initial context"* property fails for `double` (and transitively `float64Array`, which is built on `double`). One‑character‑category fix: change `isInteger` back to `isNaN`.

### Where it fails
Failing job: **Test (24.x, macos-latest, 3, 3)** — 2 failing tests (the other shards were `cancelled` by fail‑fast, so this is **not** macOS‑specific):
- `test/unit/arbitrary/double.spec.ts › double (integration) › should be able to shrink to the same values without initial context`
- `test/unit/arbitrary/float64Array.spec.ts › float64Array (integration) › should be able to shrink to the same values without initial context`

```
Caused by: AssertionError: expected 6.3453711467974095e+193 to strictly equal 1.7415426550182127e+308
```

### Root cause
In `packages/fast-check/src/arbitrary/double.ts`, the `anyDouble` unmapper:

```ts
(value) => {
  if (typeof value !== 'number') throw new Error('Unsupported type');
  if (Number.isInteger(value)) return maxIndex !== maxIndexWithNaN ? maxIndexWithNaN : minIndexWithNaN; // ⛔ should be Number.isNaN
  return doubleToIndex(value);
},
```

The diff shows the intent was purely to inline the captured globals:

```diff
-const safeNumberIsNaN = Number.isNaN;
...
-      if (safeNumberIsNaN(value)) return maxIndex !== maxIndexWithNaN ? maxIndexWithNaN : minIndexWithNaN;
+      if (Number.isInteger(value)) return maxIndex !== maxIndexWithNaN ? maxIndexWithNaN : minIndexWithNaN;
```

`safeNumberIsNaN` was substituted with `Number.isInteger` rather than `Number.isNaN`. As a result the unmapper now returns the **NaN sentinel index** for *every integer‑valued double* instead of computing `doubleToIndex(value)`. When fast-check rebuilds the shrinker from a value without its original context, integer values (and shrink paths that pass through them) get a wrong index, so the reconstructed shrink sequence diverges from the original — exactly what this property asserts.

It surfaces intermittently / on a single shard only because it's a seeded property test: it needs a generated (or shrunk) value that hits the integer branch, which not every seed does.

Note the sibling `float.ts` unmapper was migrated correctly (`Number.isNaN(value)`), so `double.ts` is the only affected spot.

### Suggested fix
```ts
if (Number.isNaN(value)) return maxIndex !== maxIndexWithNaN ? maxIndexWithNaN : minIndexWithNaN;
```

Happy to push this one‑line change to the branch if you'd like.

<sub>🤖 Automated review by Claude Code</sub>

---

packages/fast-check/src/arbitrary/double.ts
```diff
@@ -127,12 +119,12 @@ function anyDouble(constraints: Omit<DoubleConstraints, 'noInteger'>): Arbitrary
...
   return bigInt({ min: minIndexWithNaN, max: maxIndexWithNaN }).map(
     (index) => {
-      if (maxIndex < index || index < minIndex) return safeNaN;
+      if (maxIndex < index || index < minIndex) return Number.NaN;
       else return indexToDouble(index);
     },
     (value) => {
       if (typeof value !== 'number') throw new Error('Unsupported type');
-      if (safeNumberIsNaN(value)) return maxIndex !== maxIndexWithNaN ? maxIndexWithNaN : minIndexWithNaN;
+      if (Number.isInteger(value)) return maxIndex !== maxIndexWithNaN ? maxIndexWithNaN : minIndexWithNaN;
```
dubzzz:
```suggestion
      if (Number.isNaN(value)) return maxIndex !== maxIndexWithNaN ? maxIndexWithNaN : minIndexWithNaN;
```

dubzzz:
### Follow-up: full sweep for similar slips

I reconstructed every changed file by inlining each dropped capture back into the base and diffing against this PR. Only **two** hunks are genuine behaviour changes; all the rest are prettier re-wraps or intentional re-captures (`safeDateNow`/`safeSetTimeout`/`safeClearTimeout` in `DecorateProperty`, `safeDateNow` in `QualifiedParameters`).

**1. `double.ts` — the CI failure** (already noted above): `Number.isNaN` → `Number.isInteger`.

**2. `utils/stringify.ts` — latent regression, not caught by CI** (Node always has `Buffer`):

```diff
-const safeBufferIsBuffer = typeof Buffer !== 'undefined' ? Buffer.isBuffer : undefined;
 ...
-      if (typeof safeBufferIsBuffer === 'function' && safeBufferIsBuffer(value)) {
+      if (Buffer.isBuffer(value)) {
```

The old capture wasn't only about poisoning — the `typeof Buffer !== 'undefined'` ternary also guarded the **existence** of `Buffer`. Inlining it as a bare `Buffer.isBuffer(value)` drops that guard, so in any environment without a global `Buffer` (browsers without a polyfill) this line throws `ReferenceError: Buffer is not defined`. It sits in the typed-array `case` block, so it fires for **every** `Uint8Array`/`Float64Array`/… passed to `fc.stringify` — i.e. counterexample reporting would crash in the browser. Node CI never hits it because `Buffer` is always defined there.

Suggested guard-preserving form:
```ts
if (typeof Buffer !== 'undefined' && Buffer.isBuffer(value)) {
```

Everything else in the PR checks out as a clean mechanical inline.

<sub>🤖 Automated review by Claude Code</sub>

---

PR #7146, opened 2026-07-17
💥 Rename interfaces for properties

dubzzz:
## Description

Drop any idea of asynchronous or synchronous from both naming and parameters they accept.

- Drop `IRawProperty`
- Rename `IAsyncProperty` into `Property`
- Rename `IAsyncPropertyWithHooks` into `PropertyWithHooks`
- Rename `AsyncPropertyHookFunction` into `PropertyHookFunction`

Next step will be to rename `asyncProperty` into `property`, then to adapt all remaining asynchronous specific helpers.

## Checklist


dubzzz:
**TL;DR:** LGTM overall ✅ — a clean, mostly mechanical rename (`IRawProperty`/`IAsyncProperty*` → `Property`/`PropertyWithHooks`) plus a sensible restructure into `types/`, `plugins/` and `_internals/`. Dropping the `IsAsync` type parameter is a real simplification (the `as any` casts in `SkipAfterProperty` and the `fromCached` overload tower in `IgnoreEqualValuesProperty` are gone). Verified locally on the PR head: build, typecheck and the 166 property/runner unit tests all pass. Main nits: a few dead `isAsync` leftovers, and the changeset undersells two small runtime behavior changes (`decorateProperty` timeout gating, `SkipAfterProperty` now always returning a `Promise` on time-limit).

<details>
<summary>Full review</summary>

### What the PR does

- Deletes `IRawProperty` and replaces it with a single async-flavored `Property<Ts>` interface (no more `IsAsync` generic, no more `isAsync()` member).
- Renames `IAsyncProperty` → `Property`, `IAsyncPropertyWithHooks` → `PropertyWithHooks`, `AsyncPropertyHookFunction` → `PropertyHookFunction`; splits them into dedicated files under `types/`.
- Moves decorators (`SkipAfterProperty`, `TimeoutProperty`, `UnbiasedProperty`, `IgnoreEqualValuesProperty`) under `plugins/`, the implementation class (`AsyncProperty` → `PropertyImplem`) and `runIdToFrequency` under `_internals/`.
- Adapts `@fast-check/worker`, tests, type tests, and the runners docs page. Major changeset included.

### Findings

**Dead `isAsync` leftovers** (no callers remain anywhere in the workspace):
- `packages/fast-check/src/check/property/_internals/PropertyImplem.ts:66` — `isAsync(): true` is still on the class even though `Property` no longer declares it. If it's kept on purpose for the transition, a short comment would help; otherwise it can go.
- `packages/worker/src/internals/worker-property/NoopWorkerProperty.ts:16` — same leftover (`isAsync(): true`), while `WorkerPropertyFromWorker` had its removed.
- `packages/fast-check/test/unit/check/runner/Tosser.spec.ts:20` — `isAsync = () => false;` in the `wrap` helper class.

**Behavior changes hidden behind the "rename" changeset** (both look intended given the all-async direction, but worth a word in the changeset/migration notes):
- `decorateProperty` now applies `TimeoutProperty` whenever `timeout` is set — previously gated on `rawProperty.isAsync()`. For a genuinely synchronous predicate the timeout still can't fire mid-run (event loop), so it's mostly harmless, but the run output becomes a `Promise` where it previously stayed sync and the test "Should not enable TimeoutProperty on synchronous property" was deleted rather than replaced.
- `SkipAfterProperty.run` now always returns `Promise.resolve(preconditionFailure)` once the time limit is hit, and always uses `Promise.race` on the interrupt path (tests were correctly updated to reflect this).

**Minor / observations:**
- `Sampler.toProperty` duck-typing switched from `hasOwnProperty(generator, 'isAsync')` to `hasOwnProperty(generator, 'runBeforeEach')`. Semantics are unchanged (both are prototype methods, so class-based property instances — including `PropertyImplem` itself — fail the check and get re-wrapped as if they were arbitraries, with `runIdToFrequency` applied twice). That's a pre-existing quirk, but since this line is being touched, `'runBeforeEach' in generator` would make the detection actually match class instances.
- `@remarks` versioning is inconsistent: `Property`'s members are all restamped `Since 5.0.0` (though `generate`/`shrink`/`run` date back to 0.0.7/3.0.0), while `PropertyWithHooks.beforeEach/afterEach` keep `Since 1.6.0`. Also, the new interfaces say 5.0.0 but the base branch is `next-v4_9_0` — just double-checking the target release is the one you expect.
- `PropertyHookFunction` still references `GlobalAsyncPropertyHookFunction` — consistent with the PR description saying the remaining async-specific helpers come next. 👍
- Stale name in a comment: `PropertyHelpers.ts` says "inheriting from IProperty".
- Nice touches: the `has()`+`get()` pair in `IgnoreEqualValuesProperty.run` collapsed into a single `get()` (safe since `run` never returns `undefined`), and `fakeProperty` mocks are now properly typed via `MockInstance` with sane defaults.

### Verification performed

On `move-interfaces-property` (3669e42): `pnpm run build`, typecheck, and `vitest run test/unit/check/property test/unit/check/runner` → 13 files, 166 tests, all green. Grep confirms no non-historical references to the old names remain.

</details>

🤖 Review generated with [Claude Code](https://claude.com/claude-code)

---

dubzzz:
**Conclusion:** direction and implementation are sound, but the PR is not mergeable as-is — CI is red because of stale e2e snapshots, and two documentation snippets contradict the API this PR itself changes.

- 🔴 **CI failure**: `test/e2e/__snapshots__/NoRegressionStack.spec.ts.snap` still expects `AsyncProperty.run (…/AsyncProperty.generic.ts)` and `asyncRunIt`; runtime now emits `PropertyImplem.run (…/_internals/PropertyImplem.ts)` and `runIt`. Regenerate the snapshots.
- 🔴 **Docs stale**: `website/docs/core-blocks/runners.md` — the “Rewrite `assert` with `check`” tip still treats `fc.check` as synchronous and suggests `property.isAsync()`, contradicting the `Promise<RunDetails>` signature updated a few lines above in this very PR. Needs `await` + drop the `isAsync()` mention.
- 🟡 **Docs stale**: `website/docs/configuration/custom-reports.md:68` — snippet calls `property.isAsync()` (removed from the public interface). It sits in a historical “in the past” box, so either keep and label it as pre-v5, or drop it.
- 🟡 **Leftover `isAsync` remnants**: `PropertyImplem.ts:66`, worker’s `NoopWorkerProperty.ts:16`, and `Tosser.spec.ts:20` still define it although nothing calls it anymore.
- 🟡 **Changeset scope**: only `fast-check: major` is declared, yet `@fast-check/worker`’s exported `WorkerProperty` type changes shape (renamed base types, `isAsync` gone). Consider a worker changeset so the breaking surface is tracked.
- ✅ Blog posts checked: none reference `IAsyncProperty`/`IRawProperty`/`AsyncPropertyHookFunction`; the only remaining doc references are in historical migration guides, which is fine.

<details>
<summary>Details on the CI failure</summary>

Job `Test (24.x, macos-latest, 2, 3)` fails on 3 snapshots in `NoRegressionStack.spec.ts` (the other test jobs were cancelled as a consequence). Representative diff:

```diff
-     at AsyncProperty.predicate (packages/fast-check/src/check/property/AsyncProperty.ts:?:?)
+     at PropertyImplem.predicate (packages/fast-check/src/check/property/AsyncProperty.ts:?:?)
-     at AsyncProperty.run (packages/fast-check/src/check/property/AsyncProperty.generic.ts:?:?)
+     at PropertyImplem.run (packages/fast-check/src/check/property/_internals/PropertyImplem.ts:?:?)
-     at asyncRunIt (packages/fast-check/src/check/runner/Runner.ts:?:?)
+     at runIt (packages/fast-check/src/check/runner/Runner.ts:?:?)
```

Side note: these snapshots also show that user-facing failure reports now surface the internal name `PropertyImplem` and the `_internals/` path in stack traces — acceptable, but worth being deliberate about since it leaks into every reported counterexample stack.

</details>

<details>
<summary>Details on the runners.md snippet</summary>

Current tip (lines ~47–56):

```js
function assert(property, params) {
  // In this example we only support synchronous properties.
  // To support both of them, you could use `property.isAsync()` and `asyncDefaultReportMessage`.
  const out = fc.check(property, params);
  if (out.failed) {
    throw new Error(fc.defaultReportMessage(out), { cause: out.errorInstance });
  }
}
```

With this PR, `check` always returns `Promise<RunDetails>` (as documented in the updated signature on the same page), so `out.failed` reads a property off a Promise and the snippet no longer works. A minimal fix is `async function assert` + `await fc.check(...)`, removing both comment lines.

</details>

<details>
<summary>Other observations (non-blocking)</summary>

- `Sampler.toProperty` now duck-types on `hasOwnProperty(generator, 'runBeforeEach')`. Since `runBeforeEach` is a prototype method on `PropertyImplem` (as `isAsync` was before), class-based property instances still fail the check and get re-wrapped as arbitraries — identical to the pre-existing behavior, so no regression, but `'runBeforeEach' in generator` would make the detection actually match class instances if you want to fix it while touching the line.
- `IgnoreEqualValuesProperty`: collapsing `has()`+`get()` into a single `get()` with an `undefined` guard is safe — `run` can never return `undefined` per its type.
- `SkipAfterProperty`/`decorateProperty` behavior changes (always-`Promise` on time-limit, timeout no longer gated on `isAsync`) match the base branch where properties are already always async-capable; the updated unit tests reflect this correctly.
- The new `Property` interface members are stamped `@remarks Since 5.0.0` while the base branch is `next-v4_9_0` — double-check the intended release number matches.

</details>

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

PR #7148, opened 2026-07-18
💥 Drop global capture of methods [Batch 1]

dubzzz:
## Description

In the past, fast-check decided to capture all the globals and methods it relies on. The motivation behind that logic has been to be safe in case someone during the execution of a test breaks them.

Since that point in time, things moved, we moved. We now consider that in the context of a test there is no real reason to break the globals except doing it on purpose or manipulating risky code potentially subject to prototype pollution. In such cases we prefer our users to use our library called `@fast-check/poisoning`. This package is tailored to protect you and your tests against such prototype pollutions.

Another motivation to drop that is taht it increases the congnitive complexity of our code and thus its maintainance. It also adds a barrier for external contributors. Last point is that it may bring a performance penalty.

This part focues on the cleaning: drop capture on methods.

## Checklist


packages/fast-check/src/arbitrary/bigInt64Array.ts
dubzzz:
```suggestion
    BigInt64Array,
```

PR #7163, opened 2026-07-18
💥 Drop string-based PRNG in runner parameters

dubzzz:
## Description

`randomType` no longer accepts one of the hardcoded strings (`'mersenne'`, `'congruential'`, `'congruential32'`, `'xorshift128plus'`, `'xoroshiro128plus'`): it now only accepts a factory function building a `RandomGenerator` from a seed. The `RandomType` type is no longer exported. **Breaking change**: users relying on a string value should pass the corresponding factory from pure-rand instead — flagged as major via a changeset.

Keeping a single way to customize the generator removes the string-to-factory switch and three pure-rand imports from the runner configuration. Tests were updated to cover the factory path (factories are now passed through as-is since all pure-rand generators are jumpable) and the throw on non-function values.

## Checklist

### DIFF
--- a/packages/fast-check/src/check/runner/configuration/QualifiedParameters.ts
+++ b/packages/fast-check/src/check/runner/configuration/QualifiedParameters.ts
@@ -1,10 +1,7 @@
 import type { Parameters } from './Parameters.js';
 import { VerbosityLevel } from './VerbosityLevel.js';
 import type { RunDetails } from '../reporter/RunDetails.js';
-import { congruential32 } from 'pure-rand/generator/congruential32';
-import { mersenne } from 'pure-rand/generator/mersenne';
 import { xorshift128plus } from 'pure-rand/generator/xorshift128plus';
-import { xoroshiro128plus } from 'pure-rand/generator/xoroshiro128plus';
 import { adaptRandomGenerator } from '../../../random/generator/RandomGenerator.js';
 
 import type { RandomGenerator, RandomGeneratorInternal } from '../../../random/generator/RandomGenerator.js';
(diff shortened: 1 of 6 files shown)


packages/fast-check/test/unit/check/runner/configuration/QualifiedParameters.spec.ts
```diff
@@ -66,30 +59,27 @@ describe('QualifiedParameters', () => {
...
-            if (resolvedRandomType === undefined) {
-              expect(qparams.randomType).toBe(defaultRandomType);
-            } else if (resolvedRandomType === 'congruential32' || resolvedRandomType === 'mersenne') {
-              expect(qparams.randomType).not.toBe(defaultRandomType);
-              expect(qparams.randomType).not.toBe(prand[resolvedRandomType]); // re-wrapped so not fully the same
+            if (randomTypeName === undefined) {
+              expect(qparams.randomType).toBe(xorshift128plus);
             } else {
-              expect(qparams.randomType).toBe(prand[resolvedRandomType]);
+              // all pure-rand generators are jumpable, so the factory is passed through as-is
```
dubzzz:
```suggestion
```

packages/fast-check/test/unit/check/runner/configuration/QualifiedParameters.spec.ts
```diff
@@ -66,30 +59,27 @@ describe('QualifiedParameters', () => {
...
+              expect(qparams.randomType).toBe(prand[randomTypeName]);
             }
           },
         ),
       ));
-    it('Should throw on invalid randomType', async () =>
+    it('Should throw on non-function randomType', async () =>
```
dubzzz: The test does not make sense anymore

packages/fast-check/src/check/runner/configuration/QualifiedParameters.ts
```diff
@@ -125,20 +122,8 @@ function readSeed<T>(p: Parameters<T>): number {
...
-      case 'congruential32':
-        return createQualifiedRandomGenerator(congruential32);
-      case 'xorshift128plus':
-        return xorshift128plus as (seed: number) => QualifiedRandomGenerator;
-      case 'xoroshiro128plus':
-        return xoroshiro128plus as (seed: number) => QualifiedRandomGenerator;
-      default:
-        throw new Error(`Invalid random specified: '${p.randomType}'`);
-    }
+  if (typeof p.randomType !== 'function') {
```
dubzzz: Does not make sense

PR #7177, opened 2026-07-24
🐛 Handle infinite freq in option

xianjianlf2:
## Summary
- Return the wrapped arbitrary directly when `fc.option` gets `freq: Number.POSITIVE_INFINITY`
- Add unit and integration coverage for the infinite-frequency case

## Tests
- `pnpm exec vitest run packages/fast-check/test/unit/arbitrary/option.spec.ts`
- `pnpm --filter fast-check run typecheck`

## Validation update (2026-09-22)
Merged upstream main, adapting to the relocated arbitrary tests and upstream removal of safeHasOwnProperty. Option tests: 9 passed. fast-check typecheck and changed-file oxfmt/oxlint passed. Head 61452604.

AI assistance: OpenAI Codex.

### DIFF
--- a/.changeset/option-infinite-frequency.md
+++ b/.changeset/option-infinite-frequency.md
@@ -1,0 +1,5 @@
+---
+"fast-check": patch
+---
+
+Handle infinite frequency in option arbitrary.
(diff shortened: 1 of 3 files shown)


dubzzz:
@xianjianlf2 Please follow our PR template. Current proposal does not use it while it should.

PR #7187, opened 2026-07-30
✅ Fix mock blowup in FrequencyArbitrary specs

dubzzz:
## Description

From the end-user point of view this PR changes nothing: it only touches the unit tests of the main package. For maintainers it fixes the root cause that forced the `FrequencyArbitrary` specs to stay split into `part1`/`part2`, makes these two files run about 5× faster and makes their CI failures replayable again via `DEFAULT_SEED`.

**Root cause of the historical split.** In `part2`, `should reject calls having a total weight of zero` allocated a fresh `fakeArbitrary()` (6 `vi.fn()` each) for every one of its up-to-1000 entries, on every run of the predicate — around 300k mocks registered in Vitest's per-file mock registry by a single test, and never released for the lifetime of the file. Since `vi.restoreAllMocks()` runs before every test (and before every execution of a predicate), every test executed after that one paid the cost of walking the bloated registry. Reproduced at the commit that introduced the split (Jest→Vitest migration, vitest 1.4, where `restoreAllMocks` still processed `vi.fn` mocks): tests taking 120–300ms standalone took ~2s each once merged after the heavy test — 64s for one merged file vs ~14s per split file — which is why the merged file failed on CI back then. The fix is to reuse a single mocked arbitrary for all entries (the arbitraries are never even read: `from` throws on the weights first). The test drops from ~5s to ~20ms and the two files become safe to merge back together — verified by running a merged version under vitest 4.1 (26 tests in ~4.8s) and by applying the equivalent fix at the split-era commit under vitest 1.4 (64s → 21s, all tests green). Merging the files back is intentionally left out of this PR.

**Secondary fix.** Both files called `fc.configureGlobal({ beforeEach: beforeEachHook })`, which replaces the whole global configuration and therefore silently dropped the `seed` set from `DEFAULT_SEED` by `vitest.setup.mjs`: failures of these two files on CI could not be replayed from the reported seed. They now spread `fc.readConfigureGlobal()` first, the same pattern as `SpyCleaner.ts`. The stale comment `// This test does not pass even alone` (a leftover of the historical failure) is removed, and the outdated comment above the hooks is rewritten to describe what the hooks actually do.

No changeset added: the change is test-only and has no impact on the published packages. The PR is focused on this single concern. No new tests: the changed code is the tests themselves, and both files were re-run (including with fixed `DEFAULT_SEED` values), plus `tsc --noEmit` and Prettier.

## Checklist

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01AcXuY4bSKv97JyuYB5wwax

---
_Generated by [Claude Code](https://claude.ai/code/session_01AcXuY4bSKv97JyuYB5wwax)_

### DIFF
--- a/packages/fast-check/test/unit/arbitrary/_internals/FrequencyArbitrary.part1.spec.ts
+++ b/packages/fast-check/test/unit/arbitrary/_internals/FrequencyArbitrary.part1.spec.ts
@@ -17,11 +17,12 @@ import { sizeArb } from '../__test-helpers__/SizeHelpers.js';
 function beforeEachHook() {
   vi.restoreAllMocks();
 }
-// No need to restore hook between each execution of fast-check
-// the hooks will be resetted within the tests themselves if needed
-//  >  fc.configureGlobal({ beforeEach: beforeEachHook });
+// Restore spies between tests and between each execution of a predicate: call counts on
+// vi.spyOn-based spies must not leak from one run to the next one.
+// Spreading the already configured global settings preserves the seed possibly set from
+// DEFAULT_SEED by vitest.setup.mjs (configureGlobal replaces the whole configuration).
 beforeEach(beforeEachHook);
-fc.configureGlobal({ beforeEach: beforeEachHook });
+fc.configureGlobal({ ...fc.readConfigureGlobal(), beforeEach: beforeEachHook });
 
 const frequencyValidInputsArb = fc
   .tuple(
--- a/packages/fast-check/test/unit/arbitrary/_internals/FrequencyArbitrary.part1.spec.ts
+++ b/packages/fast-check/test/unit/arbitrary/_internals/FrequencyArbitrary.part1.spec.ts
@@ -17,10 +17,6 @@ import { sizeArb } from '../__test-helpers__/SizeHelpers.js';
 function beforeEachHook() {
   vi.restoreAllMocks();
 }
-// Restore spies between tests and between each execution of a predicate: call counts on
-// vi.spyOn-based spies must not leak from one run to the next one.
-// Spreading the already configured global settings preserves the seed possibly set from
-// DEFAULT_SEED by vitest.setup.mjs (configureGlobal replaces the whole configuration).
 beforeEach(beforeEachHook);
 fc.configureGlobal({ ...fc.readConfigureGlobal(), beforeEach: beforeEachHook });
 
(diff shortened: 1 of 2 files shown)


packages/fast-check/test/unit/arbitrary/_internals/FrequencyArbitrary.part1.spec.ts
```diff
@@ -17,11 +17,12 @@ import { sizeArb } from '../__test-helpers__/SizeHelpers.js';
 function beforeEachHook() {
   vi.restoreAllMocks();
 }
-// No need to restore hook between each execution of fast-check
-// the hooks will be resetted within the tests themselves if needed
-//  >  fc.configureGlobal({ beforeEach: beforeEachHook });
+// Restore spies between tests and between each execution of a predicate: call counts on
+// vi.spyOn-based spies must not leak from one run to the next one.
+// Spreading the already configured global settings preserves the seed possibly set from
+// DEFAULT_SEED by vitest.setup.mjs (configureGlobal replaces the whole configuration).
```
dubzzz:
```suggestion
```

packages/fast-check/test/unit/arbitrary/_internals/FrequencyArbitrary.part2.spec.ts
```diff
@@ -10,11 +10,12 @@ import { sizeArb } from '../__test-helpers__/SizeHelpers.js';
 function beforeEachHook() {
   vi.restoreAllMocks();
 }
-// No need to restore hook between each execution of fast-check
-// the hooks will be resetted within the tests themselves if needed
-//  >  fc.configureGlobal({ beforeEach: beforeEachHook });
+// Restore spies between tests and between each execution of a predicate: call counts on
+// vi.spyOn-based spies must not leak from one run to the next one.
+// Spreading the already configured global settings preserves the seed possibly set from
+// DEFAULT_SEED by vitest.setup.mjs (configureGlobal replaces the whole configuration).
```
dubzzz:
```suggestion
```

packages/fast-check/test/unit/arbitrary/_internals/FrequencyArbitrary.part2.spec.ts
```diff
@@ -141,14 +142,18 @@ describe('FrequencyArbitrary', () => {
...
 
-    // This test does not pass even alone
     it('should reject calls having a total weight of zero', async () =>
       await fc.assert(
         fc.asyncProperty(fc.nat({ max: 1000 }), (numEntries) => {
           // Arrange
+          // Reuse a single mocked arbitrary for all the entries: allocating one per entry means
+          // up to 1000 x 6 vi.fn() per run of the predicate, all of them staying registered in the
+          // mock registry of Vitest for the lifetime of the test file and thus slowing down any
+          // subsequent call to vi.restoreAllMocks (it runs before every execution of a predicate)
```
dubzzz:
```suggestion
          // Reuse a single mocked arbitrary for faster throughput and prevent memory issues
```

PR #7228, opened 2026-08-18
✨ Add `timeout` plugin to stop long running predicates

dubzzz:
## Description

This PR ports the behavior of `TimeoutProperty` to the plugin API as a new `timeout` plugin, exposed publicly as `fc.timeout`. Users can now fail long-running executions of their predicates by declaring the plugin, either per run or globally:

```ts
await fc.assert(
  fc.asyncProperty(fc.integer(), async (x) => {
    /* ... */
  }),
  { plugins: [fc.timeout(1000)] },
);
```

Any execution of the predicate taking more than the requested delay gets marked as failed with the same error as the one produced by the `timeout` parameter: `Property timeout: exceeded limit of N milliseconds`. As predicates cannot be stopped, the underlying execution keeps running in the background but its outcome gets ignored.

- The plugin follows the pattern introduced by the `beforeEach` plugin (including the plain export name, aligned with the `beforeEach`/`afterEach` renaming from #7227), but contrary to it each `timeout(...)` produces its own standalone instance: no merging via the cross-plugin context.
- `TimeoutProperty` was only wired for asynchronous properties by `decorateProperty`, while a plugin cannot know upfront whether the property is asynchronous. To respect the sync-in-sync-out contract of `decorateRun`, the plugin starts its timer, calls the nested run and, when the output is synchronous, clears the pending timer and forwards the output synchronously. The plugin is therefore a safe no-op on synchronous properties, even if only meaningful for asynchronous ones.
- The `timeout` parameter and `TimeoutProperty` are left untouched: rewiring the parameter through the plugin would change semantics (a plugin wraps `runBeforeEach`/`runAfterEach` too, which `TimeoutProperty` explicitly excludes), so it is kept for a potential follow-up.
- Impact flagged as minor via a changeset. The PR stays focused on adding this single plugin (with its tests and documentation).
- Documentation: besides the dedicated page for the plugin, the general plugins page gains a "Combining plugins" section built on a realistic `timeout(10_000)` / `beforeEach(...)` / `timeout(5_000)` example, explaining that declaration order defines the nesting — the outer timeout covering the setup hook and the predicate, the inner one covering the predicate alone.
- Tests: the applicable specs of `TimeoutProperty.spec.ts` were ported against `decorateRun` (success/failure in time, timeout on slow and never-ending runs, timer cleanup on both outcomes), plus new cases covering the synchronous pass-through and the absence of instance merging. An end-to-end check through `fc.check` lives in a dedicated `test/e2e/TimeoutPlugin.spec.ts`, following the dedicated e2e file introduced for life-cycle plugins. They would have failed without this PR since the plugin did not exist.

Note: expect a trivial conflict with #7229/#7230/#7231 on the plugin import/export lines of `fast-check-default.ts` — whichever merges later needs a one-line rebase.

## Checklist

### DIFF
--- a/website/docs/core-blocks/plugins/index.md
+++ b/website/docs/core-blocks/plugins/index.md
@@ -58,13 +58,9 @@ await fc.assert(
 );
 ```
 
-For each execution of the predicate:
+The plugins setup above makes use of two instances of the `timeout` plugin. The first instance of it is responsible to make sure that the time taken by the `beforeEach` cleanup plus the one taken by the predicate will never exceed 10 seconds. The second one, only wraps the predicate and makes sure that it won't take more than 5 seconds.
 
-1. the 10-second `timeout` starts first: it wraps everything declared after it — the database reset, the 5-second `timeout` and the predicate,
-2. the `beforeEach` hook then resets the database,
-3. the 5-second `timeout` starts last, right before the predicate: it only wraps the predicate.
-
-As a consequence, a run fails as soon as a single call to the search endpoint takes more than 5 seconds, or when the database reset and the call together take more than 10 seconds. Order matters: moving `timeout(5_000)` before the `beforeEach` would make the reset count against the 5-second budget too, and the 10-second limit could then never fire.
+This short example clearly highlights why the declaration order of plugins is crucial. Two different orders may result in totally different expectations. Correctly ordering them will make you capable of finely configure your flows to fit your needs.
 
 ## The plugins
 
--- a/website/docs/core-blocks/plugins/index.md
+++ b/website/docs/core-blocks/plugins/index.md
@@ -6,7 +6,7 @@ description: Extend the default execution flow provided by runners to meet your 
 
 # Plugins
 
-Plugins provide a way to extend and refine the runtime and execution behavior of your properties. In the same way building custom arbitraries gives you the flexibility to tweak the generation flows, plugins gives you ways to customize how your properties run.
+Plugins provide a way to extend and refine the runtime and execution behavior of your properties. In the same way building custom arbitraries gives you the flexibility to tweak generation flows, plugins give you ways to customize how your properties run.
 
 Plugins are designed to support things such as:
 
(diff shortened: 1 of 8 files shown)


website/docs/core-blocks/plugins/index.md
```diff
@@ -34,6 +34,38 @@ fc.installGlobalPlugin(pluginA());
 
 Installed plugins run before the ones passed to the runner, so the snippet above followed by `fc.assert(myProp, { plugins: [pluginB()] })` is equivalent to `plugins: [pluginA(), pluginB()]`.
 
+## Combining plugins
+
+Plugins wrap every execution of the predicate and the declaration order defines the nesting: the first declared plugin is the outermost wrapper, each subsequent one sits a bit closer to the predicate.
+
+Let's illustrate it on a property checking a search endpoint, with the database brought back to a clean state before each run:
```
dubzzz:
```suggestion
Plugins have been designed to be combined together. When multiple plugins get defined for the same assertion, they execute themselves in declaration order. The first plugin runs first and wraps the execution of the second one until we reach the main operation. Plugins can wrap various stages of the flows including the execution of the predicate function.

Let's illustrate how plugins get combined on a property checking a search endpoint. In this example the database gets bring back to a clean state before each run:
```

website/docs/core-blocks/plugins/index.md
```diff
@@ -34,6 +34,38 @@ fc.installGlobalPlugin(pluginA());
...
+);
+```
+
+For each execution of the predicate:
+
+1. the 10-second `timeout` starts first: it wraps everything declared after it — the database reset, the 5-second `timeout` and the predicate,
+2. the `beforeEach` hook then resets the database,
+3. the 5-second `timeout` starts last, right before the predicate: it only wraps the predicate.
+
+As a consequence, a run fails as soon as a single call to the search endpoint takes more than 5 seconds, or when the database reset and the call together take more than 10 seconds. Order matters: moving `timeout(5_000)` before the `beforeEach` would make the reset count against the 5-second budget too, and the 10-second limit could then never fire.
```
dubzzz:
```suggestion
The plugins setup above makes use of two instances of the `timeout` plugin. The first instance of it is responsible to make sure that the time taken by the `beforeEach` cleanup plus the one taken by the predicate will never exceed 10 seconds. The second one, only wraps the predicate and makes sure that it won't take more than 5 seconds.

This short example clearly highlights why the declaration order of plugins is crucial. Two different orders may result in totally different expectations. Correctly ordering them will make you capable of finely configure your flows to fit your needs.
```

website/docs/core-blocks/plugins/timeout.md
```diff
@@ -0,0 +1,23 @@
...
+
+```ts
+{
+  plugins: [
+    timeout(1000), // fail any execution of the predicate taking more than 1 second
+  ];
+}
+```
+
+As predicates cannot be stopped, whenever a timeout occurs the underlying execution keeps running in the background but its outcome gets ignored. Note also that the plugin can only be useful for asynchronous properties: a synchronous predicate can never be observed timing out as it already came to an end when the runner gets its outcome back.
```
dubzzz:
```suggestion
:::warning[Predicates can't be stopped]
As predicates cannot be stopped, whenever a timeout occurs the underlying execution keeps running in the background but its outcome gets ignored.
:::

:::warning[Synchronous predicates can't be interrupted]
The plugin has no impact on purely synchronous predicates. It can't stop them so timeout won't really be capable of interrupting them.
:::
```

dubzzz:
@copilot look for typos in updated doc and jsdoc. Come up with a fix pr if anything needs to be fix for typos or if some parts really need rephrasing

PR #7229, opened 2026-08-18
✨ Add plugin to interrupt after time limit

dubzzz:
## Description

This PR ports the behavior of `SkipAfterProperty` to the plugin API as two plugins: `fc.skipAllAfterTimeLimit(timeLimitMs)` (port of the `skipAllAfterTimeLimit` parameter) and `fc.interruptAfterTimeLimit(timeLimitMs)` (port of the `interruptAfterTimeLimit` parameter). Users can now bound the total time spent assessing a property by declaring the plugins, either per run or globally:

```ts
await fc.assert(
  fc.asyncProperty(fc.integer(), async (x) => {
    /* ... */
  }),
  { plugins: [fc.interruptAfterTimeLimit(1000)] },
);
```

- Both plugins share one internal builder parameterized by the interrupt flag, mirroring how `SkipAfterProperty` took `interruptExecution`. Each call produces its own standalone instance: no merging via the cross-plugin context. Export names are plain (no suffix), aligned with the `beforeEach`/`afterEach` renaming from #7227, and match the existing parameter names.
- The clock starts at plugin instantiation, which happens when `check`/`assert` starts assessing the property — equivalent to the `SkipAfterProperty` constructor being called from `decorateProperty`.
- Two intentional behavior differences compared to the property-based decoration, both flagged rather than hidden: (1) once the limit is passed, the skip happens before `runBeforeEach`/`runAfterEach`, so hooks no longer execute for skipped runs — the old spec carried a comment saying "We may not want to run hooks in such context, but so far we do", and the plugin position naturally gives the arguably better behavior; (2) for the interrupting variant on asynchronous properties, the raced window now includes `runBeforeEach`/`runAfterEach` as plugins wrap the whole execution of the predicate.
- Sync-in-sync-out contract of `decorateRun` is respected: an expired limit returns the `PreconditionFailure` synchronously, and a synchronous nested run gets its pending interruption timer cleared and its output forwarded synchronously (matching the old code which never raced synchronous properties).
- Timers and `Date.now` are used directly, matching the direction taken for the `timeout` plugin in #7228. The `skipAllAfterTimeLimit`/`interruptAfterTimeLimit` parameters and `SkipAfterProperty` are left untouched.
- Impact flagged as minor via a changeset. The PR stays focused on porting this single property (two run modes, hence two plugins), with tests and a documentation page.
- Tests: the applicable specs of `SkipAfterProperty.spec.ts` were ported against `decorateRun` (forwarding within the limit, skip with the right interrupt flag past the limit, interruption of long-running executions, timer cleanup on success and failure), plus new cases for the synchronous pass-through, the absence of timers in the skip-only variant, and the absence of instance merging. End-to-end checks of both plugins through `fc.check` live in a dedicated `test/e2e/TimeLimitPlugins.spec.ts`, following the dedicated e2e file introduced for life-cycle plugins. They would have failed without this PR since the plugins did not exist.

Note: expect a trivial conflict with #7228/#7230/#7231 on the plugin import/export lines of `fast-check-default.ts` — whichever merges later needs a one-line rebase.

## Checklist


packages/fast-check/src/check/plugin/InterrruptAfterTimeLimitPlugin.ts
```diff
@@ -0,0 +1,65 @@
+import { PreconditionFailure } from '../precondition/PreconditionFailure.js';
+import type { IRawProperty } from '../property/IRawProperty.js';
+import type { Plugin, PluginInstance } from './Plugin.js';
+
+/** @internal */
+function interruptAfterDelay(timeMs: number) {
```
dubzzz, line 6: We should probably start it once for all runs and reuse then clean it in the afterall

packages/fast-check/test/unit/check/plugin/InterrruptAfterTimeLimitPlugin.spec.ts
dubzzz, line 1: We should rename the file defining the plugin

PR #7230, opened 2026-08-18
✨ Add plugins to drop runs on already covered cases

dubzzz:
## Description

This PR ports the behavior of `IgnoreEqualValuesProperty` to the plugin API as two plugins: `fc.ignoreEqualValues()` (port of the `ignoreEqualValues` parameter) and `fc.skipEqualValues()` (port of the `skipEqualValues` parameter). Users can now prevent predicates from being executed twice on the same value by declaring the plugins, either per run or globally:

```ts
fc.assert(
  fc.property(fc.integer(), (x) => {
    /* ... */
  }),
  { plugins: [fc.skipEqualValues()] },
);
```

As with the parameters, the `skip` variant marks replayed successes as skipped (giving guarantees on the number of distinct covered values, at the price of possibly failing with "too many skips") while the `ignore` variant replays the previous outcome as-is.

- Both plugins share one internal builder parameterized by the skip flag, mirroring how `IgnoreEqualValuesProperty` took `skipRuns`. Each call produces its own standalone instance with its own cache of covered cases: no merging via the cross-plugin context (and the cache lifetime — one `check`/`assert` run — matches the lifetime of the property decoration it replaces). Export names are plain (no suffix), aligned with the `beforeEach`/`afterEach` renaming from #7227, and match the existing parameter names.
- The old `fromCached` needed the property's `isAsync()` to know how to replay a cached outcome; a plugin has no property at hand, so asynchronicity is detected from the cached outcome itself (thenable or not). Synchronicity is preserved: a synchronous run stays synchronous, covered cases included.
- One behavior difference, flagged rather than hidden: on an already covered case the plugin short-circuits the whole execution, so `runBeforeEach`/`runAfterEach` no longer execute for replayed runs. The old spec carried a comment saying "We may not want to run hooks twice but once in such context, but so far we do" — the plugin position naturally gives the arguably better behavior.
- The `ignoreEqualValues`/`skipEqualValues` parameters and `IgnoreEqualValuesProperty` are left untouched.
- Impact flagged as minor via a changeset. The PR stays focused on porting this single property (two modes, hence two plugins), with tests and a documentation page.
- Tests: the specs of `IgnoreEqualValuesProperty.spec.ts` were ported against `decorateRun` (single execution per value, re-execution on new values, cached-outcome tables for success/failure/skip in sync and async), plus new cases for cache isolation between instances and synchronous output preservation on duplicates. End-to-end checks of both plugins through `fc.check` — including the deterministic "too many skips" failure of the skip variant on a constant arbitrary — live in a dedicated `test/e2e/EqualValuesPlugins.spec.ts`, following the dedicated e2e file introduced for life-cycle plugins. They would have failed without this PR since the plugins did not exist.

Note: expect a trivial conflict with #7228/#7229/#7231 on the plugin import/export lines of `fast-check-default.ts` — whichever merges later needs a one-line rebase.

## Checklist


dubzzz:
Specific handling needed for mutable values

Issue #7236, opened 2026-08-25
💡 Ideas of evolutions for our plugin API before it reaches a release


dubzzz:
## Context

The plugin API (#7216, #7221, #7222, #7224, #7227, #7232, #7235) is documented as "Since 4.10.0" but has not shipped yet — the package is still at 4.9.0. This issue gathers ideas of evolutions we could apply to it **while renames and reshapes are still free**, before the surface freezes with a release.

The long-term vision driving these ideas:

- plugins should eventually cover `decorateGenerate` and `decorateShrink` in addition to today's `decorateRun`,
- reporters (`Parameters.reporter` / `Parameters.asyncReporter`) should be droppable in favor of plugins enriching reporting,
- `assert` should become nothing more than `check` plus a plugin providing the predefined throwing reporter.

The proposals below were shaped by surveying plugin mechanisms from other projects — Rollup/Vite, Babel, Webpack, ESLint, Prettier, Jest/Vitest, Node loader hooks, and outside JS: pytest/pluggy, Hypothesis, JUnit 5, RSpec, FsCheck, ScalaCheck, proptest, tower. Each idea references its inspiration.

## Proposed surface

```ts
/**
 * A plugin is a factory instantiated once per property run.
 * Shape validated by Babel: `(api) => ({ name, pre, post, visitor, ... })`.
 */
export type Plugin<Ts> = (context: PluginContext) => PluginInstance<Ts>;

export type PluginContext = {
  /** fast-check version, for compatibility checks — Babel's `api.version` / `assertVersion` */
  version: string;
  /** Index of this plugin within the composed chain (today's 1st positional argument) */
  pluginIndex: number;
  /** Namespaced cross-plugin storage — JUnit 5's `ExtensionContext.Store` (today's `crossPluginContext`) */
  store: PluginStore;
};

export type PluginStore = {
  get<T>(key: symbol): T | undefined;
  set<T>(key: symbol, value: T): void;
  /** Makes the "find-or-register" dance of cooperating plugins a single call */
  getOrCreate<T>(key: symbol, create: () => T): T;
};

export type PluginInstance<Ts> = {
  /** Required — error attribution, diagnostics, dedup (Rollup / Webpack) */
  name: string;
  /** Ordering constraint relative to un-enforced plugins (Vite `enforce`, pytest `tryfirst`/`trylast`) */
  enforce?: 'pre' | 'post';

  // ── configuration phase ──
  /**
[...]

PR #7239, opened 2026-09-01
✨ Add extra plugin's method called `onAllRunsComplete`

dubzzz:
## Description

As suggested into https://github.com/dubzzz/fast-check/issues/7236, this new method will provide customers with ways to override the errors being reported by `assert` but also provide similar reporting capabilities to `check`.

As such we plan to advertise them as replacements for `reporter` and `asyncReporter`.

## Checklist

### DIFF
--- a/.changeset/shaggy-goats-know.md
+++ b/.changeset/shaggy-goats-know.md
@@ -1,0 +1,7 @@
+---
+"fast-check": minor
+"@fast-check/jest": patch
+"@fast-check/vitest": patch
+---
+
+✨ Add extra plugin's method called `onAllRunsComplete`
--- a/packages/fast-check/src/check/plugin/Plugin.ts
+++ b/packages/fast-check/src/check/plugin/Plugin.ts
@@ -24,11 +24,12 @@ export type PluginInstance<Ts> = {
    */
   decorateRun?: (nestedRun: IRawProperty<Ts, boolean>['run']) => IRawProperty<Ts, boolean>['run'];
   /**
-   * Called once at the end of the full property assessment.
-   * Gets called with the result of the execution.
+   * Called once at the end of the full property assessment, with the result of the execution.
    *
    * Throwing allows you to override the default error reporting provided by {@link assert}.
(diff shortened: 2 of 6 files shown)


dubzzz:
Ordering rationale, in short: **`onAllRunsComplete` runs in declaration order** — it is a notification list, so output follows `plugins: [...]` as written, and with the implicit default reporter appended last by `assert`, a user reporter's error stays the primary one. **`afterAll` runs reversed** — teardown must unwind nesting: a resource opened by an earlier plugin can be used by a later one, so the later one releases first.

Same choices elsewhere:

- Notifications in declaration order: Jest iterates reporters in registration order for `onRunComplete` ([ReporterDispatcher.ts](https://github.com/jestjs/jest/blob/main/packages/jest-core/src/ReporterDispatcher.ts)); ScalaCheck's `TestCallback.chain` calls self then the chained callback, i.e. declaration order ([Test.scala](https://github.com/typelevel/scalacheck/blob/main/core/shared/src/main/scala/org/scalacheck/Test.scala)).
- Teardown reversed: Go `t.Cleanup` — "last added, first called" ([testing.T.Cleanup](https://pkg.go.dev/testing#T.Cleanup)); pytest — "Finalizers are executed in a first-in-last-out order" ([fixtures how-to](https://docs.pytest.org/en/stable/how-to/fixtures.html)); JUnit 5 after-callbacks run in reverse registration order ([user guide](https://junit.org/junit5/docs/current/user-guide/#extensions-lifecycle-callbacks)); the language itself with `using` declarations — "`[Symbol.dispose]()` is invoked in the reverse order of their declaration" ([TC39 explicit resource management](https://github.com/tc39/proposal-explicit-resource-management)).

Opposite way, for the record: **Mocha** runs same-level hooks, `afterEach`/`after` included, "in the order they are defined" — FIFO teardown ([docs](https://github.com/mochajs/mocha/blob/v10.2.0/docs/index.md), live at [mochajs.org/#hooks](https://mochajs.org/#hooks)). No library was found notifying reporters in reverse.

---

PR #7242, opened 2026-09-02
⚡️ Avoid unneeded Promise overhead in life-cycle plugins

dubzzz:
The runner behind the beforeEach/afterEach plugins used to chain one
`.then` per hook as soon as the flow switched to asynchronous, even for
hooks being purely synchronous, and seeded the after part with an extra
`Promise.resolve`. It now runs hooks from root synchronous loops and only
chains a `.then` when a hook actually returns a `Promise`: the loop
resumes from within its resolution for the remaining hooks.

It also aligns the failure precedence of synchronously throwing after
hooks with the asynchronous flow: the first failure is preserved instead
of being overwritten by a later after hook throwing synchronously.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_015iGbbNKHix7TnTX3SuqP8U

### DIFF
--- a/.changeset/light-hooks-hurry.md
+++ b/.changeset/light-hooks-hurry.md
@@ -1,0 +1,5 @@
+---
+"fast-check": patch
+---
+
+⚡️ Avoid unneeded `Promise` overhead in life-cycle plugins
--- a/packages/fast-check/src/check/plugin/LifeCyclePlugins.ts
+++ b/packages/fast-check/src/check/plugin/LifeCyclePlugins.ts
@@ -1,4 +1,5 @@
-import type { IRawProperty } from '../property/IRawProperty.js';
+import type { IRawProperty, PropertyFailure } from '../property/IRawProperty.js';
+import type { PreconditionFailure } from '../precondition/PreconditionFailure.js';
 import type { Plugin, PluginInstance } from './Plugin.js';
 
 const LifeCyclePluginSymbol = Symbol.for('fast-check/plugin/life-cycle');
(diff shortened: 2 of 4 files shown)


dubzzz:
Added a benchmark for the life-cycle plugins in `packages/fast-check/test/bench/lifeCyclePlugins.bench.ts` (commit bc9ebea) and ran it against both implementations. Each entry runs a full `fc.assert` (100 runs) with the hooks of the setup; hooks increment a counter so their bodies cannot be dropped.

**Results** — median ops/s over 15 interleaved A/B rounds (`vitest bench`, `time: 1500ms` per entry per round, both variants built with rolldown and swapped into `lib/` between rounds):

| Setup | main (ops/s) | PR (ops/s) | Change |
|---|---:|---:|---:|
| async property without any hook (control) | 8,819 | 8,918 | +1.1% |
| sync property with sync-only hooks | 10,273 | 10,829 | +5.4% |
| async property with sync-only hooks | 5,726 | 6,470 | +13.0% |
| async property with an async beforeEach followed by sync-only hooks | 4,883 | 5,777 | +18.3% |
| async property with async-only hooks | 5,132 | 5,164 | +0.6% |

sync-only hooks = 2 `beforeEach` (1 returning a teardown) + 2 `afterEach`, all synchronous; async-only hooks = 1 async `beforeEach` returning an async teardown + 1 async `afterEach`.

Reading of the numbers:

- The control (no hook, untouched code path) lands at +1.1%, which is the noise floor of the machine: differences of that magnitude are not signal.
- The wins concentrate exactly where the change applies: setups where synchronous hooks used to be requeued behind intermediate `.then` (+13% / +18.3%), and the sync `afterEach` chain behind an async predicate (+5.4% end-to-end on a sync property despite the flow never leaving sync).
- Async-only hooks are flat (+0.6%), as expected: that flow is dominated by the hooks' own promises, which the change does not touch.
- These are end-to-end `fc.assert` numbers, so arbitrary generation and the runner dilute the per-hook effect; on the hook layer alone, the same mixed setup drops from ~14 to ~5 `Promise#then` calls per run.

Environment: Node v22.22.2, Intel Xeon @ 2.10GHz (cloud container). Per-setup interquartile spread was ±2-10% across rounds, hence the medians over interleaved rounds rather than a single `--compare` run.

---

